Issue #70, opened 2014-05-08
subtract_oscan assumes overscan is always the second axis


mwcraig:
e.g.:

``` python
oscan = np.median(overscan.data, axis=1)
```

Working on a fix...

PR #72, opened 2014-05-09
Allow overscan region to be along either axis

mwcraig:
As reported in #70, `subtract_overscan` assumed that the axis along which the overscan should be aggregated (with either median or mean) was the second of the two axes. This allows the overscan to be in either axis (though it will fail in the unlikely case that the overscan is square).

### DIFF
--- a/ccdproc/ccdproc.py
+++ b/ccdproc/ccdproc.py
@@ -84,7 +84,7 @@ def create_variance(ccd_data, gain=None, readnoise=None):
 
 
 @log_to_metadata
-def subtract_overscan(ccd, overscan=None, fits_section=None,
+def subtract_overscan(ccd, overscan=None, overscan_axis=1, fits_section=None,
                       median=False, model=None):
     """
     Subtract the overscan region from an image.  This will first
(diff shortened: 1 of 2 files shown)


evertrol:
Apologies for commenting on an old PR, but I was wondering whether it would be an option to *also* have the ability to guess the overscan axis, by explicitly using `overscan_axis=None` as argument.

The use case for this would be when the overscan section is obtained directly from a FITS header (TRIMSEC, BIASSEC or such) and passed on to `subtract_overscan` with the `fits_section` keyword. 
At that point, anyone creating a generic reduction pipeline, would have to programmatically inspect the FITS section string and deduce the axis from that. Alternatively, and easier on the programmer side, is to let `subtract_overscan` decide: since it creates the overscan section from the FITS section in the line
```python
overscan = ccd[slice_from_string(fits_section, fits_convention=True)] 
```
it could deduce the overscan axis from the shape of the overscan section.
It then becomes the programmers responsibility to go with ccdproc's educated guess, or deduce the axis themselves.

(Any decent FITS file should, of course, have the overscan direction as a header card, but that's quite often not the case, and the direction is implied.)

mwcraig:
@evertrol -- would you mind open a separate issue for this?

evertrol:
I have created an example implementation at PR #263

Issue #186, opened 2015-01-27
update to astropy v1.0


crawfordsm:
Things that will need to be updated:

- [x] CCDData to use new NDData
- [x] Being able to read/write CCDData objects

PR #187, opened 2015-02-03
Nddata1.0

crawfordsm:
Updated to CCData for changes to NDData appearing in astropy v1.0.  CCDData will not be compatible with earlier versions of astropy.   A number of properties or setters were needed to be added in order to make this work and a number of these should be considered being moved to NDataArray in astropy.   

In addition, there was two additional changes to the test suite:
1) CCDData.read will not work with a fits file with an empty data array.  In test_history_preserved_if_metadata_is_fits_header in test_ccddata, a line was added to create a data array to the fake image.  

2)  a problem needed to be fixed in the tests for scaling with the combiner class.   In this case, np.ma.median in numpy v.1.9.1  does not return the same result as np.median when the array does not have a mask.   The instance of np.ma.median was replaced with np.median as it is not important to the actual test of the scaling function.


crawfordsm:
Closing #186

crawfordsm:
Merging as it is building fine using the development build.   I have opened issue #188 to deal with the travis failures.

PR #198, opened 2015-03-04
Add proper handling of WCS

mwcraig:
This really ended up being a couple of changes:

+ Create a WCS object if the necessary information is present when creating a `CCDData` object from a FITS file.
+ If there is a WCS attribute then write that attribute's information into the HDU created with `to_hdu`.

I also added a test that slicing properly affects the WCS attribute.

There is still one potential issue here -- the header could have WCS information different than the WCS attribute if the `CCDData` object is created with a FITS header as the metadata and that header would get partially overwritten (but maybe not completely if the WCS in the header is a different projection/has distortions and the WCS attribute doesn''t).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,11 +9,14 @@ New Features
   it is not and there is no data in the primary extension, the first extension
   with data will be used.
 
+- Set wcs attribute when reading from a FITS file that contains WCS keywords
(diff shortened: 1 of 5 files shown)


mwcraig:
Managed to forget about the changelog entry and the documentation, additional commits shortly.

mwcraig:
@crawfordsm -- merge if this looks good to you.

PR #201, opened 2015-03-31
Updated astropy-helpers to release-1.0.1

crawfordsm:
This is an update of astropy-helpers to fix the most recent version.   I've initiated this due to the install of ccdproc when it is a dependency in another package to fail when building on readthedocs due to VersionConflicts in the astropy_helpers.   

I think this is related to https://github.com/astropy/astropy-helpers/pull/144 but I'm not 100% sure.   However, it is probably best to have the lastest version of helpers in ccdproc anyway.  

This will require a patch of ccdproc to release to ultimately fix this bug I think though.

Also a note about updating to the most recent version of astropy_helpers should be added to the release documents.

### DIFF
--- a/ah_bootstrap.py
+++ b/ah_bootstrap.py
@@ -56,6 +56,12 @@ else:
     _text_type = str
     PY3 = True
 
+
+# What follows are several import statements meant to deal with install-time
+# issues with either missing or misbehaving pacakges (including making sure
+# setuptools itself is installed):
+
+
 # Some pre-setuptools checks to ensure that either distribute or setuptools >=
 # 0.7 is used (over pre-distribute setuptools) if it is available on the path;
 # otherwise the latest setuptools will be downloaded and bootstrapped with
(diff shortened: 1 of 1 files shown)


crawfordsm:
Thoughts about this @mwcraig ?

mwcraig:
Looks good to merge!

crawfordsm:
Not directly related to this, but do you know if there is a way in the setup.py to point to an install_requires in the setup.py to a development version or do I need to go through a release for pypi to take advantage of this update?

crawfordsm:
Asking just to see if there is a way to test this prior to doing a full patch release?

mwcraig:
You could upload this branch to pypi if dev is in the version string, then pip install --pre (and maybe --no-deps too).

I would think you could pin to a dev version in setup.py but don't recall if I've tried it. 

If you have manually installed ccdproc dev version in the environment then setupnot should be happy with that and skip the install.

If you need it to work for others it it's probably easier to just do a patch release, can help with that in about 1 hour.

crawfordsm:
Ah actually what I needed was adding ccdproc to the rtd-requirements file and that worked, so releasing a patch isn't needed as something to be done immediately.

PR #208, opened 2015-04-22
tests added for core.py functions as requested in issue 117

### LINKED ISSUE #117: Add test to check input array does not change
When an input array is passed to a function, it should not be modified to the function.  A test should be added to the appropriate functions that the input array does not change.

walkerna22:
Closes #117


mwcraig:
@walkerna22 -- could you please rebase this? If you have the branch you are working on checked out you can do `git rebase upstream/master` and all should go well.

**Make a backup branch before you start** just in case things don't go well (i.e. branch from the place you are right now).

PR #227, opened 2015-07-14
Opt in to new container-based builds on travis

mwcraig:
See this for details: http://docs.travis-ci.com/user/migrating-from-legacy/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,17 @@
 language: python
 
+# Setting sudo to false opts in to Travis-CI container-based builds.
+sudo: false
+
+# The apt packages below are needed for sphinx builds, which can no longer
+# be installed with sudo apt-get.
+addons:
+    apt:
+        packages:
+            - graphviz
+            - texlive-latex-extra
+            - dvipng
+
 python:
     - 2.6
     - 2.7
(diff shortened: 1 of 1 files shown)


mwcraig:
@mdboom -- I'm seeing a failure against astropy-dev related to merging in wcslib-5.8 (I think). An example of the fail log is at https://travis-ci.org/mwcraig/ccdproc/jobs/70959072#L578

The test in question is [here](https://github.com/astropy/ccdproc/blob/master/ccdproc/tests/test_ccddata.py#L508), and it calls [this method](https://github.com/astropy/ccdproc/blob/master/ccdproc/ccddata.py#L220) of `CCDData` (link is to line where the failure happens).

Essentially, it looks like there is a keyword whose name is `''` when the `wcs` is converted to a header, and `io.fits` throws an error when trying to add that keyword to the header. 

I can work around this in our testing, please let me know if I should open an issue for it on astropy...

mdboom:
@embray (as the `astropy.io.fits` expert) may know better so should chime in.  A key of `''` is not in error -- it contains all of the "blank" cards that don't contain a keyword (but may contain a comment).  The relevant change is in wcslib which now produces a blank followed by a `COMMENT` about the version of wcslib that produced the header.  That's a change in behavior, yes, but I don't think it's a buggy one.

The problem is that `header['']` returns a list (since there may be multiple blank cards), and setting `header[''] = ['']` doesn't work.  I see the issue with duplicate blank cards with doing `header.update(wcs_header)`.  Maybe @embray has a better solution to offer for copying the cards over, but a `if key == '': continue` block would work here.

mwcraig:
Thanks @mdboom, that makes sense.

PR #229, opened 2015-07-16
Copy WCS header in a way that properly handles list-like keywords

### LINKED ISSUE #228: Check for presence of list-like keywords when converting WCS-produced header to FITS
As explained in #227, the new wcslib contains a blank keyword, which returns a list since there can be multiple values. The fix, I think, is to catch any of the possible list-like keywords (blank, `COMMENT`, `HISTORY`) and copy the list.

mwcraig:
Some keywords (blank, comment, history) can have multiple values. Using `iteritems` properly iterates over those cases.

Closes #228

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ matrix:
         # Try Astropy development version
         - python: 2.7
           env: ASTROPY_VERSION=development SETUP_CMD='test'
-        - python: 3.3
+        - python: 3.4
           env: ASTROPY_VERSION=development SETUP_CMD='test'
 
         # Try all python versions with the latest numpy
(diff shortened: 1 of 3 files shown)


mwcraig:
@crawfordsm -- I'm inclined to merge this even though the tests fail for astropy dev. The failure will be fixed once astropy/astropy#3968 is merged and astropy/astropy#3967 is fixed.

I could, instead, catch the exception that occurs and deal with it appropriately. Might be better so we don't need to increase the minimum version of astropy to 1.1.

mwcraig:
Merged because the tests now pass with latest dev version of astropy.

Issue #230, opened 2015-07-17
to_hdu should be using update, not unique, argument to extend if WCS is present.


mwcraig:
Assuming, of course, we view the information in the WCS to be authoritative, which we should. As currently implemented the WCS information in the header will be kept.

PR #231, opened 2015-07-18
Add test of and fix for #230 -- properly updating WCS in header

mwcraig:


mwcraig:
Merging, since this ought to have been what happened when I removed the loop anyway.

PR #237, opened 2015-08-19
Remove mention about in-place operation

evertrol:
The example given in the documentation does not provide in-place
operation, since this is not available (yet) in ccdproc.

### DIFF
--- a/docs/ccdproc/index.rst
+++ b/docs/ccdproc/index.rst
@@ -61,33 +61,33 @@ information to the image metadata.
 
 Logging can be simple -- add a string to the metadata:
 
-    >>> image_2_gained = ccdproc.gain_correct(image_2, 1.5 * u.photon/u.adu, add_keyword='gain_corrected')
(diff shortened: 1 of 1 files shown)


mwcraig:
Thanks, @evertrol -- @crawfordsm and I should get around to this by Tuesday afternoon at the latest. The failing tests are unrelated to your change, I think.

evertrol:
It appears these failing tests actually were caused by my PR. The removal of the `image_2` caused doc-test failures further down the line. Not enough testing on my side, sorry.

I note that the following happens on Python 3.4 (my go-to Python executable) when running the test suite:
> Running doctests in .rst files is not supported on Python 3.x

which means the failures will only show up when the tests are run with Python 2. I guess this is an upstream issue with either astropy testing, or even pytest. It strikes me as odd though.

I've now updated the PR, by replacing occurrences of `image_2` with `dark_sub` (and variants thereof), which appears to fix the failing doc-tests.

PR #244, opened 2015-10-20
Fix overscan subtraction on non-square image

### LINKED ISSUE #240: Overscan does not works with non-square images and axis != 1
I tried to do an `subtract_overscan` on a non-square image and got an error regarding the shape mismatch. To reproduce this I made this test:

	import numpy as np
	arr1 = ccdproc.CCDData(np.ones([100, 110]), unit=u.adu)
	no_scan = ccdproc.subtract_overscan(arr1, overscan=arr1[:, 90:100])
	no_scan = ccdproc.subtract_overscan(arr1, overscan=arr1[90:100], overscan_axis=0)

The first one works, the second raises:

	ValueError: operands could not be broadcast together with shapes (100,110) (110,1) 

The way I fixed it here is here: cfbf5e158bbd4239d09fd49605fc1c7bc5090b10.

mwcraig:
Closes #240 using the fix from @wschoenell

### DIFF
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -220,7 +220,10 @@ def subtract_overscan(ccd, overscan=None, overscan_axis=1, fits_section=None,
         else:
             oscan = np.reshape(oscan, (1, oscan.size))
     else:
-        oscan = np.reshape(oscan, oscan.shape + (1,))
+        if overscan_axis == 1:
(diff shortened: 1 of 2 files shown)


mwcraig:
Thanks for the fix, @wschoenell -- will get an update out in the next couple days.

Issue #224, opened 2015-07-12
Allow to chose a function for uncertainty calculation in Combiner


bsipocz:
Along the logic of the optional `median_func` and `scale_func`, there may be room for choosing the uncertainty function to be different than the default in `median_combine()` and `average_combine`.

MSeifert04:
Both combination functions contain an `uncertainty_func` in the 1.0.x branch: [`median_combine`](https://github.com/astropy/ccdproc/blob/master/ccdproc/combiner.py#L303) and [`average_combine`](https://github.com/astropy/ccdproc/blob/master/ccdproc/combiner.py#L379)

PR #258, opened 2015-12-07
Added default uncertainty function in average_combine of combiner.py

stottsco:
Added a default uncertainty function in average_combine of combiner.py so that an uncertainty function can be passed optionally. This closes part of issue #224. As for the other part would it be preferable to let the default uncertainty function in median_combine be median_absolute_deviation and then use a conditional to multiply by the factor of 1.4826 if the uncertainty function is the default using something like:  
 
def median_combine(self, median_func=ma.median, scale_to=None, uncertainty_func=median_absolute_deviation):
. . . 
if uncertainty_func is median_absolute_deviation:
    uncertainty_func *= 1.4826

or would it be better to define a function for median_absolute_deviation that does this multiplication so that uncertainty can simply be assigned the default or optionally passed function without having to worry about whether the factor of 1.4826 is applicable?

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -269,7 +269,7 @@ class Combiner(object):
                it overrides ``CCDData.scaling``. Defaults to None.
 
            uncertainty_func : function, optional
-               Function to calculate uncertainty. Defaults to ccdproc.sigma_func
+               Function to calculate uncertainty. Defaults to `ccdproc.sigma_func`
 
            Returns
            -------
@@ -332,7 +332,7 @@ class Combiner(object):
                it overrides ``CCDData.scaling``. Defaults to None.
 
            uncertainty_func: function, optional
-                Function to calculate uncertainty. Defaults to numpy.ma.std
+                Function to calculate uncertainty. Defaults to `numpy.ma.std`
 
            Returns
            -------
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -308,7 +308,7 @@ class Combiner(object):
         # return the combined image
         return combined_image
 
-    def average_combine(self, scale_func=ma.average, scale_to=None):
+    def average_combine(self, scale_func=ma.average, scale_to=None, uncertainty_func=ma.std):
         """ Average combine together a set of arrays.
 
            A `~ccdproc.CCDData` object is returned with the data property
@@ -352,7 +352,7 @@ class Combiner(object):
         mask = (mask == len(self.data_arr))
 
         # set up the deviation
-        uncertainty = ma.std(self.data_arr, axis=0)
+        uncertainty = uncertainty_func(self.data_arr, axis=0)
 
         # create the combined image with a dtype that matches the combiner
         combined_image = CCDData(np.asarray(data.data, dtype=self.dtype),
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -267,9 +267,9 @@ class Combiner(object):
            scale_to : float, optional
                Scaling factor used in the average combined image. If given,
                it overrides ``CCDData.scaling``. Defaults to None.
-               
+
            uncertainty_func : function, optional
-               Function to calculate uncertainty. Defaults to sigma_func
+               Function to calculate uncertainty. Defaults to ccdproc.sigma_func
 
            Returns
            -------
@@ -330,10 +330,10 @@ class Combiner(object):
            scale_to : float, optional
                Scaling factor used in the average combined image. If given,
                it overrides ``CCDData.scaling``. Defaults to None.
-               
+
            uncertainty_func: function, optional
-                Function to calculate uncertainty. Defaults to ma.std
-                
+                Function to calculate uncertainty. Defaults to numpy.ma.std
+
            Returns
            -------
            combined_image: `~ccdproc.CCDData`
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -6,7 +6,7 @@ from __future__ import (absolute_import, division, print_function,
 import numpy as np
 from numpy import ma
 from .ccddata import CCDData
-from .core import trim_image
+from .core import trim_image, sigma_func
 
 from astropy.stats import median_absolute_deviation
 from astropy.nddata import StdDevUncertainty
@@ -247,7 +247,7 @@ class Combiner(object):
             self.data_arr.mask[mask] = True
 
     # set up the combining algorithms
-    def median_combine(self, median_func=ma.median, scale_to=None):
+    def median_combine(self, median_func=ma.median, scale_to=None, uncertainty_func=sigma_func):
         """Median combine a set of arrays.
 
            A `~ccdproc.CCDData` object is returned
@@ -267,6 +267,9 @@ class Combiner(object):
            scale_to : float, optional
                Scaling factor used in the average combined image. If given,
                it overrides ``CCDData.scaling``. Defaults to None.
+               
+           uncertainty_func : function, optional
+               Function to calculate uncertainty. Defaults to sigma_func
 
            Returns
            -------
@@ -294,8 +297,7 @@ class Combiner(object):
         mask = (mask == len(self.data_arr))
 
         # set the uncertainty
-        uncertainty = 1.4826 * median_absolute_deviation(self.data_arr.data,
-                                                         axis=0)
+        uncertainty = uncertainty_func(self.data_arr.data, axis=0)
 
         # create the combined image with a dtype matching the combiner
         combined_image = CCDData(np.asarray(data.data, dtype=self.dtype),
@@ -328,7 +330,10 @@ class Combiner(object):
            scale_to : float, optional
                Scaling factor used in the average combined image. If given,
                it overrides ``CCDData.scaling``. Defaults to None.
-
+               
+           uncertainty_func: function, optional
+                Function to calculate uncertainty. Defaults to ma.std
+                
            Returns
            -------
            combined_image: `~ccdproc.CCDData`
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -571,7 +571,7 @@ def transform_image(ccd, transform_func, **kwargs):
     return nccd
 
 
-def sigma_func(arr):
+def sigma_func(arr, axis=0):
     """
     Robust method for calculating the deviation of an array. ``sigma_func``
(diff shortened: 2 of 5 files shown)


crawfordsm:
Over looks pretty good @stottsco.  Thanks for the contribution! 

Can you add in this parameter to the doc string?  Just a line describing what it does. 

One thing that may be helpful would be adding an example and I think this would solve your question above as well.  If someone wanted to use the median_absolute_deviation function as the uncertainty, they could do it as a lambda function.  So something like:

    ccd = c.average_combine(uncertainty_func=lambda x: 1.426*median_absolute_deviation(x))

Although it would be good to double check this works.

crawfordsm:
@mwcraig there is a failure in the test suite that is related to the image collection documentation with python 3.5.   I haven't looked at detail, but it is unrelated to this PR, but looks like some kind of ordering being checked and coming back in a different order.

stottsco:
For average_combine I added a line to the doc string as requested, and in median_combine I added an optional argument to the parameter list uncertainty_func=median_absolute_deviation, and changed the line

uncertainty = 1.4826*median_absolute_deviation(self.data_arr, axis=0)
to
uncertainty = uncertainty_func(self.data_arr, axis=0)
so that the 1.4 factor could be put in optionally by use of a lambda function as crawfordsm suggested. I wrote a test for it to see if it would work but I get a TypeError: <lambda>() got an unexpected keyword argument 'axis' when I ran the test. I pasted the error into a search engine and found some stuff on github and stackoverflow that wasn't particularly useful. The code for the test is pasted below also. I essentially recycled my test for average_combine_uncertainty and changed the last assignment statement to use a lambda function.

The code for the test is:
'''python

def test_median_combine_unc_lambda(ccd_data):
    c = Combiner(ccd_list)
    ccd = c.median_combine(uncertainty_func=lambda x: 1.4826*np.sum(x))
    assert np.all(ccd.uncertainty.array == 1.4826*np.sum(c.data_arr, 0))
'''

From the error it looks like lambda was expecting only one arg but was passed two so I tried lambda x,y: 1.4826*np.sum(x,y) hoping it would expect two but I received the same error.

mwcraig:
@stottsco -- now that you mention it, a `lambda` won't work here because the `uncertainty_func` needs to be able to accept two arguments and a `lambda` can only have one. There is a function from `astropy.stats` that *almost* does what we want: [`mad_std`](http://docs.astropy.org/en/stable/api/astropy.stats.mad_std.html).

I think the RIght Thing To Do is:

+ Open a PR on astropy to add an `axis` argument to `mad_std`.
+ Import `mad_std` from `astropy.stats` and use that as the default.

@stottsco, are you up for doing the astropy pull request?

mwcraig:
@stottsco -- would like to get this wrapped up this week. Turns out the function we need is already defined (almost) at https://github.com/astropy/ccdproc/blob/master/ccdproc/core.py#L574

So just need to add an `axis` argument to `sigma_func` with default value `None`, and pass that on and this can get closed!

mwcraig:
@stottsco -- the changes you made look good, I'd like to to get the git history simplified a bit. We'll tackle that at the Thursday morning coding meeting.

mwcraig:
@stottsco -- looks one `<<<<<<< HEAD` got left in during the rebase. Removing that line (just a regular commit to make that change, not another rebase) and this should be good.

ccdproc/core.py
```diff
@@ -580,13 +580,16 @@ def sigma_func(arr):
     ----------
     arr : `~ccdproc.CCDData` or `~numpy.ndarray`
         Array whose deviation is to be calculated.
+        
+    axis : None or int or tuple of ints, optional
+        Axis or axes along which a sum is performed. The default (axis = None) is perform a sum over all the dimensions of the input array. axis may be negative, in which case it counts from the last to the first axis.
```
MSeifert04: Could you split that line into multiple ones it's recommended that lines are not too much over 80 letters per row?  Also I find it a bit misleading: It's not a sum that is performed it's a "sigma function". The word "axis" should be capitalized when it's at the beginning of the sentence.

I would suggest:

```
Axis or axes along which the function is performed.
If ``None`` (the default) it is performed over all the dimensions of the input array.
The axis argument can also be negative, in which case it counts from the last to the first axis.
```

ccdproc/core.py
```diff
@@ -571,7 +571,7 @@ def transform_image(ccd, transform_func, **kwargs):
     return nccd
 
 
-def sigma_func(arr):
+def sigma_func(arr, axis=0):
```
MSeifert04: Default should be ``None`` not ``0``.

ccdproc/tests/test_combiner.py
```diff
@@ -332,3 +332,17 @@ def test_combine_limitedmem_scale_fitsimages():
 
     np.testing.assert_array_almost_equal(avgccd.data, ccd_by_combiner.data, decimal = 4)
 
+#test the optional uncertainty function in average_combine
+def test_average_combine_uncertainty(ccd_data):
+    ccd_list = [ccd_data, ccd_data, ccd_data]
+    c = Combiner(ccd_list)
+    ccd = c.average_combine(uncertainty_func=np.sum)
+    assert np.all(ccd.uncertainty.array == np.sum(c.data_arr, 0))
```
MSeifert04: Just a small thing: Numpy has some testing functions: ``np.testing.assert_array_equal`` and ``np.testing.assert_array_almost_equal`` for such comparisons. See line 333 for the usage.

That's not mandatory but I think it's a bit cleaner.

ccdproc/core.py
```diff
@@ -580,13 +580,16 @@ def sigma_func(arr):
...
+    axis : None or int or tuple of ints, optional
+        Axis or axes along which a sum is performed. The default (axis = None) is perform a sum over all the dimensions of the input array. axis may be negative, in which case it counts from the last to the first axis.
 
     Returns
     -------
     float
         standard deviation of array
     """
-    return 1.4826 * stats.median_absolute_deviation(arr)
+    return 1.4826 * stats.median_absolute_deviation(arr, 0)
```
MSeifert04: ``astropy`` uses some more digits for this factor: ``1.482602218505602`` (see http://docs.astropy.org/en/stable/_modules/astropy/stats/funcs.html#mad_std). It may not be necessary but in case ``ccdproc`` gets merged into astropy or ccdproc someday switches to the median the different precision might be problematic.

ccdproc/combiner.py
```diff
@@ -328,7 +330,10 @@ def average_combine(self, scale_func=ma.average, scale_to=None):
            scale_to : float, optional
                Scaling factor used in the average combined image. If given,
                it overrides ``CCDData.scaling``. Defaults to None.
-
+               
+           uncertainty_func: function, optional
+                Function to calculate uncertainty. Defaults to ma.std
```
MSeifert04: Maybe you could link to the function since ``std`` is in my opinion not unambiguous in it's meaning: \`numpy.ma.std\`

ccdproc/combiner.py
```diff
@@ -267,6 +267,9 @@ def median_combine(self, median_func=ma.median, scale_to=None):
            scale_to : float, optional
                Scaling factor used in the average combined image. If given,
                it overrides ``CCDData.scaling``. Defaults to None.
+               
+           uncertainty_func : function, optional
+               Function to calculate uncertainty. Defaults to sigma_func
```
MSeifert04: Same here: \`ccdproc.sigma_func\`

MSeifert04:
I hope I'm not too late. I just left some inline comments. Most of them are not very important. Only the thing about the default value for ``axis`` in ``sigma_func`` is a bit confusing.

stottsco:
Hi,
I just pushed to my repo I think I addressed all of the changes you requested. For the difference in precision between mad_std and sigma_func you were talking about I just pasted in the more precise scale value used in astropy.mad_std so that essentially ccdproc.sigma_func in core.py is the same exact function as astropy.mad_std in stats/funcs.py except that sigma_func takes an optional axis argument.

If anything else needs to be changed let me know.

ccdproc/tests/test_ccddata.py
```diff
@@ -574,6 +574,7 @@ def test_header(ccd_data):
 def test_wcs(ccd_data):
     ccd_data.wcs = 5
     assert ccd_data.wcs == 5
+<<<<<<< HEAD
```
MSeifert04: I guess that's a leftover from your rebase @mwcraig mentioned. You need to delete this line.

ccdproc/combiner.py
```diff
@@ -329,6 +331,9 @@ def average_combine(self, scale_func=ma.average, scale_to=None):
                Scaling factor used in the average combined image. If given,
                it overrides ``CCDData.scaling``. Defaults to None.
 
+           uncertainty_func: function, optional
+                Function to calculate uncertainty. Defaults to numpy.ma.std
```
MSeifert04: Could you wrap `numpy.ma.std` with single backticks? That will tell sphinx to convert it to a link to the numpy documentation:
\`numpy.ma.std\`

Same also for the sigma-func.

stottsco:
Okay, I edited the doc strings for those two functions. As for the <<<<<<<<HEAD thing I don't see this line in the file on my local machine nor if I look at the file in my repo on github. All I have is:

```python
def test_wcs(ccd_data):
    ccd_data.wcs = 5
    assert ccd_data.wcs == 5
```
and then its EOF so I'm not sure how to fix this?

MSeifert04:
@stottsco - I'm not too familiar with git myself I think/hope @mwcraig or @crawfordsm know what to do in this case.

mwcraig:
Looks like `test_wcs` shows up *twice* now, once at the very end of `test_ccddata.py` and again at much earlier in the file at line 574, and it is the version at line 574 that has the rebase leftover in it.

stottsco:
Oh gotcha, I didn't catch that. Thanks, I took that out hopefully should be good to go now.

mwcraig:
@stottsco -- just one last thing, I think. In the rebase frenzy I think you ended up with *duplicates* of the WCS property and setter, and duplicates of the test. Could you please delete the duplicates (not by another rebase ;), just delete and commit)?

mwcraig:
@MSeifert04 -- thanks for reviewing this!

stottsco:
Wow, good catch. Got that fixed too now, thanks.

mwcraig:
Thanks, @stottsco

PR #261, opened 2015-12-14
Banner

### LINKED ISSUE #214: add banner
Add banner to the documents and in other relevant places

crawfordsm:
This should add all of the files needed so that the top banner uses the ccdproc image and ccdproc name.  @mwcraig  can you check that I have not forgot any files but that these does build fine?

**edit:** closes #214 (github only pays attention to this in commit messages or the body of the PR).


crawfordsm:
closes #214

mwcraig:
Looks like it built, so I'm merging -- didn't see anything missing when I looked over the file changes.

mwcraig:
Weird, got the same fail as in #258 when I merged, even though it didn't fail in the PR.

In any event, have found the root cause of that problem, will open a PR to fix shortly.

Issue #262, opened 2015-12-14
Order of column names in ImageFileCollection is non-deterministic


mwcraig:
The reason is that the initial (ordered) list of keys is passed through a set (which is unordered) in https://github.com/astropy/ccdproc/blob/master/ccdproc/image_collection.py#L162-L168

Affects #258, among others...

mwcraig:
And: https://github.com/astropy/ccdproc/blob/master/ccdproc/image_collection.py#L359-L360

And then (though note this is an `OrderedDict`): https://github.com/astropy/ccdproc/blob/master/ccdproc/image_collection.py#L375-L383

evertrol:
Just an additional note, since I just stumbled upon this when deciphering the Travis failure for PR #263 : in the case where `_summayr_info` is assigned to an `astropy.table.Table` (https://github.com/astropy/ccdproc/blob/master/ccdproc/image_collection.py#L80-L84), the order of the `colnames` attribute is again indeterminate, since this is a list created from a the keys of a dict (https://github.com/astropy/astropy/blob/master/astropy/table/table.py#L1267).

This affects the doctest in docs/ccdproc/image_management.rst  at random.

PR #263, opened 2015-12-15
Add option to automatically determine the overscan axis

evertrol:
If the overscan axis is explicitly set to `None` in
`subtract_overscan`, the overscan direction is determined from the
shape of the overscan section, set the shortest dimension. (In case of
a square overscan, the default of 1 is picked.)

This can be convenient in cases where `fits_section` is used with an
(unknown) input string (for example, from an image's BIASSEC header
keyword): the user can let the function determine the appropriate
axis.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@ New Features
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- The overscan_axis argument in subtract_overscan can now be set to
+  None, to let subtract_overscan provide a best guess for the axis. [#263]
+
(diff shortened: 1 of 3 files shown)


evertrol:
This PR is incomplete, and largely meant as an issue/RFC, with an example implementation. 
Unit tests, docstring update and a CHANGES update are still missing.

evertrol:
I guess the failure of Travis here is directly related to #262 ?

crawfordsm:
Yes, the test failure is related to that, which shouldn't effect the rest of your work. If you want to go ahead and add the doc_string update, a test, and something to the change log that would be great.  Thanks for the contribution!

Issue #264, opened 2015-12-15
fits_section does not allow for '*' indexing


evertrol:
The `fits_section` argument used in `subtract_overscan` and `trim_image` does not allow for the use of an asterisk wildcard to indicate all rows or columns. Instead, it uses the Python convention for indicating this, using a lone double colon.

This is fairly easily resolved while still allowing for the `:` syntax, by replacing all `*` with `:` halfway the `slice_from_string` function. For example:

    ...
    no_space = no_space.strip('[]')                                                            
    if fits_convention:                                                                        
        no_space = no_space.replace('*', ':')                                                  
    string_slices = no_space.split(',')           
    ...

This should even work for the case of `[*:2, *]`, which correctly transforms to `[::2, :]`.  It will fail, however, for a case like `[-*, *]`, which is a typical FITS section style for flipping an axis. 

I've created this issue to see if there is interest in supporting this, and what varieties of `*` indexing it should support.

(Examples of FITS sectioning can be found at https://heasarc.gsfc.nasa.gov/docs/software/fitsio/c/c_user/node94.html .)

crawfordsm:
Thanks for the feedback @evertrol .   If you have a chance, please feel free to to put in a pull request with this update.  If not, we will see if it can be added soon.   I think the code you added looks like it would work for most cases with a small test added for it.   

Hmm, probably could open a second pull request to handle the special case of `[-*,*]`

evertrol:
Not a problem: I can implement a PR possibly tomorrow, or later this week.

For the special `[-*,*]` case, I wonder if the following cases are valid FITS sectioning syntax:
- `[-*:2,*]`
- `[*:-2,*]`

I suspect the former is, while the latter isn't. The link given above is not that explicit about the syntax, but from the example that uses `[*:2, 512:256:2]`, I don't think FITS uses, or even allows for, negative step sizes.

That last example also makes me wonder if this syntax is supported: start being larger than stop, but with a positive increment. Seen the handling of FITS sections in ccdproc, I guess this won't work either (haven't tried though).

Obviously, all these special cases are unlikely to occur, so it raises the question in how far ccdproc should support FITS sectioning syntax.

I'll give it a try with a PR, and see in how far I can reasonably accommodate these special cases.

mwcraig:
Closed by #265

PR #265, opened 2015-12-16
Extend handling of FITS section syntax for slicing

evertrol:
This handles FITS sections with
- wildcards: '[*]'
- negative wildcards: `[-*]`
- inverted range specification: `[20:10]`
- step size specification for each of the above cases:
  `[*:2]`, `[-*:2]` or `[20:10:2]`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ New Features
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- Add support for wildcard and reversed FITS style slicing. [#265]
+
 Bug Fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


evertrol:
This closes issue #264 

Three notes:
- Handling of the inverted sectioning turned out to be quite complicated, in particular for cases like `[10:1]`, where a simple manipulation could lead to `slice(9, -1, -1)`, and the special case of a stop value of `1` has to be handled differently.
  This has led to quite a few unit tests, but please double check if this looks correct.

- Is there a specific reason assertions use the form
```python
assert (sli[0].start is 3 and  
            sli[0].stop is 5 and
            sli[0].step is 2)
 ```
instead of simply
```python
assert sli[0] == slice(3, 5, 2) 
```
? 
If the latter is incorrect, I'll have to update the unit tests.

- I haven't updated CHANGEST.rst yet. I'll update the PR shortly to include it, but would this go under bug fixes or Changes and Additions? I assume this is for 0.3.2.

crawfordsm:
Test failure related to #262

mwcraig:
@evertrol -- please put the changes under 0.4...right now the changelog is a total mess. The most recent release is actually 0.3.3, and if we decide to do an 0.3.4 release I'll backport the changes.

I'd count this as a a change/addition. Will plan to review this tomorrow.

PR #267, opened 2015-12-16
Ensure column order in ImageFileCollection matches the order of keywords in constructor

mwcraig:
In addition, this clarifies the docstring a bit to explain how column order is determined.

The only case in which it is determined by the order of the keywords passed to the constructor is when the wildcard `*` is *not* used.

Otherwise the column order is the order in which the keywords appear in the first FITS file opened.


mwcraig:
Wow, think I made it worse...

mwcraig:
Merging to kill this bug...

PR #268, opened 2015-12-17
Remove use of print function for non-explicit output

### LINKED ISSUE #266: print function in combiner.combine() can be rather noisy
The explicit use of `print()` in `combiner.combine()` can't be turned off, and can cause noisy output where it's not wanted.

The offending lines are https://github.com/astropy/ccdproc/blob/master/ccdproc/combiner.py#L475-L476 and https://github.com/astropy/ccdproc/blob/master/ccdproc/combiner.py#L492

The latter could become `logging.info(...)` or `astropy.log.info(...)`.

For the former, I doubt the exception needs to be caught and printed at all, and the exception can just bubble up to the user. The usual error message tends to as clear or clearer (if the input path happens to be directory, ccdproc will still print a message saying the file can't be found, next to a message saying it's a directory). Also, in line https://github.com/astropy/ccdproc/blob/master/ccdproc/combiner.py#L517 , the exception isn't caught at all for the rest of the files in the input.

evertrol:
Change an informative print function to astropy.log.info(), and remove
a try-except clause with superfluous print functions.

Closes #266

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -10,6 +10,7 @@ from .core import trim_image
 
 from astropy.stats import median_absolute_deviation
 from astropy.nddata import StdDevUncertainty
+from astropy import log
 
 __all__ = ['Combiner','combine']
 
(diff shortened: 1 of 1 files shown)


evertrol:
I have (automated standard practice) also removed trailing whitespace. This may obfuscate the diff somewhat, but the general view is that trailing whitespace is bad, so I left the changes. Unless the whitespace change is wanted in a different commit, in which case I can backtrack and update the PR.

mwcraig:
@evertrol -- Thanks for all of the PRs! 

The whitespace changes are fine. Would you mind rebasing this, #263 and #265 just to make sure there are no issues after the fix I implemented for #262?

mwcraig:
@crawfordsm -- assuming tests pass it looks good.

PR #272, opened 2016-01-07
adding astroscappy to cosmicrays_lacosmic

### LINKED ISSUE #202: Integrate new lacosmicx into ccdproc
Once the new version of lacosmicx is accepted somewhere https://github.com/astropy/imageutils/pull/35 integrate it into ccdproc

crawfordsm:
This upates cosmicrays_lacosmic to use astroscrappy.   This closes #202.

Unfortunately, it does look like the algorithm has some problems near the edges of the images and so for right now, the assert tests for number of cosmic rays found has been commented out.  But this would appear to be an issue with astroscappy rather than the wrapper to it, and an issue has ben raised there.


crawfordsm:
The other question and this would require updating the pull request is whether astroscappy should be a requirement or optional requirement to ccdproc.   it does look like it might have some issues on windows and as such we may want to make it an optional requirement.

mwcraig:
@crawfordsm -- I do have `astroscrappy` building on Windows for the astropy conda channel. I'd be OK with making it a hard requirement (we already have require astropy, of course, which has compiled extensions, so compiling per se shouldn't be an issue).

crawfordsm:
closing this due to merging problems.

PR #273, opened 2016-01-08
Adding appveyor testing and opting in ci-helpers

bsipocz:
This is to address and close #259.

I've also added ccdproc as a project on appveyor itself for astropy.

PR #274, opened 2016-01-08
Add a 'sort' method to ImageFileCollection

hamogu:
This uses the underlying table structure to do that sort, it also updates
the separate self._files accordingly.
This method allows me to set a the order in which the files are processed,
e.g. I can iterate over all files and make plots sorted alphabetically by
'OBJECT' or by 'OBS-TIME' or 'FILTER'.

A related change is to sort the file list when reading files from a
directory. This sorting is alphabetical, if the user does not like it
they can always use the new 'sort' function later.
This change makes the order of files that are read in reproducible, which is useful
when the iterators are used to make plots of each file - I want the same order
when I re-run my script.
(Without this change the order of files seems to change randomly at times.)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,8 @@ Bug Fixes
 New Features
 ^^^^^^^^^^^^
 
+- add a ``sort`` method to ImageFileCollection [#274]
+
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


hamogu:
The appveyor tests fail, because the setup is not complete. I assume that that has to do with the fact the @bsipocz activated appveyor tests, but #273 is not merged yet.

mwcraig:
@hamogu -- once again, thank you! Could you please add a short mention of this to https://github.com/astropy/ccdproc/blob/master/docs/ccdproc/image_management.rst ?

hamogu:
@bsipocz : "1 failing check": Where did Travis go? It seems we are running only appveyor now?

hamogu:
OK, I pushed again and now I see Travis. That may have been a github fluke.

bsipocz:
@hamogu - Must have been a glitch. Sometimes there is an annoying delay in picking up the commits.

hamogu:
Docs added. Tests passed.

PR #275, opened 2016-01-08
Include an option to return the full path in ImageFileCollection.files_filtered

hamogu:
This introduces a new keyword "include_path".
Note that this keyword can never clash with any fits header keyword, because
those are limited to 8 characters.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@
 New Features
 ^^^^^^^^^^^^
 
+- Add a new keyword to ImageFileCollection.files_filtered to return the full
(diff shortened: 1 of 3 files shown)


hamogu:
The appveyor tests fail, because the setup is not complete. I assume that that has to do with the fact the @bsipocz activated appveyor tests, but #273 is not merged yet.

mwcraig:
@hamogu -- thanks, good idea!

@crawfordsm -- looks good to me, merge, if you agree.

crawfordsm:
Thanks @hamogu !

PR #277, opened 2016-01-13
wrap astroscrappy  for lacosmic

crawfordsm:
This is a second attempt at a pull request to merge in astroscrappy support as the previous one ended up with a mess of different merging problems.


crawfordsm:
@mwcraig I've finally gotten it to past all the tests.  You might want to check the appveyor file as I'm not sure that I did install astroscrappy in the best possible way, but it is a way that does work.   

Let me know if there is anything else to change, otherwise, this should be good to merge.

PR #280, opened 2016-01-15
Modifies CCDData arithmetic with number/Quantity to preserve WCS

### LINKED ISSUE #278: CCDData arithmetic with number/Quantity removes wcs attributes
The current implementation of CCDData._ccddata_arithmetic() returns a new CCDData object with valid data, unit, uncertainty and meta attributes derived/copied from the original but simply discards any WCS  present, e.g.:

    In [12]: test = ccdproc.CCDData.read(path, unit = 'adu')
    
    In [13]: test.wcs
    Out[13]: 
    WCS Keywords
    
    Number of WCS axes: 2
    CTYPE : 'RA---TAN'  'DEC--TAN'  
    CRVAL : 13.614671744880001  -37.589520634069999  
    CRPIX : 2608.0  2134.5  
    CD1_1 CD1_2  : -0.00079067726619539997  0.0  
    CD2_1 CD2_2  : 0.0  0.00079067726619539997  
    NAXIS    : 5215 4268
    
    In [14]: test2 = test.multiply(1.0)
    
    In [15]: test2.wcs
    
    In [16]:

I would argue that for simple arithmetic operations with numbers/Quantities any WCS would be just as valid afterwards as before and so should be copied over into the result of CCDData._ccddata_arithmetic() in the same way meta attributes currently are.  This would simplify code that, for example, uses CCDData.multiply() to apply a flux calibration to an image while preserving all its metadata.

AnthonyHorton:
Fixes #278

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,7 @@ Other Changes and Additions
   primary hdu, the resultant header object is a combination of the 
   header information in the primary hdu and the first hdu with data. [#271]
 - Changed cosmicray_lacosmic to use astroscrappy for cleaning. [#272]
+- CCDData arithmetic with number/Quantity now preserves any existing WCS. [#278]
 
 Bug Fixes
 ^^^^^^^^^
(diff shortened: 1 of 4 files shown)


crawfordsm:
This looks like it might need a rebase as I just merged another pull request with edits to the change log.

AnthonyHorton:
OK, will do.

crawfordsm:
Thanks @AnthonyHorton !

Issue #281, opened 2016-01-15
trim_image() does not correctly update WCS


AnthonyHorton:
The trim_image() function returns a trimmed copy of the input CCDData object but does not correctly update the WCS.  The CRPIX values appear to be modified accordingly but the NAXIS values are unchanged and consequently the trimmed image ends up with an invalid WCS, e.g.:

```
>>> test = ccdproc.CCDData.read(path, unit='adu')
>>> test.shape
(4268, 5215)
>>> test.wcs
WCS Keywords

Number of WCS axes: 2
CTYPE : 'RA---TAN'  'DEC--TAN'  
CRVAL : 13.614671744880001  -37.589520634069999  
CRPIX : 2608.0  2134.5  
CD1_1 CD1_2  : -0.00079067726619539997  0.0  
CD2_1 CD2_2  : 0.0  0.00079067726619539997  
NAXIS    : 5215 4268
>>> result = ccdproc.trim_image(test[100:500,250:850])
>>> result.shape
(400, 600)
>>> result.wcs
WCS Keywords

Number of WCS axes: 2
CTYPE : 'RA---TAN'  'DEC--TAN'  
CRVAL : 13.614671744880001  -37.589520634069999  
CRPIX : 2358.0  2034.5  
CD1_1 CD1_2  : -0.00079067726619539997  0.0  
CD2_1 CD2_2  : 0.0  0.00079067726619539997  
NAXIS    : 5215 4268
```

The documentation in https://github.com/astropy/ccdproc/blob/master/docs/ccdproc/ccddata.rst includes the statement "Either way, the wcs attribute is kept up to date if the ~ccdproc.ccddata.CCDData image is trimmed." however until this is fixed that's not quite true.

A look at the code suggests this is actually an upstream bug in astropy.wcs.

crawfordsm:
Thanks @AnthonyHorton for the report.   The NAXIS should definitely be changed and the wcs updated (or at the very least a warning raised when trim is used with an object with wcs).   I know there is a large pull request underway with `astropy.nddata`, but if you have more details on what needs to change or can submit a PR (either here or at astropy.nddata), please go ahead.

MSeifert04:
[...]

PR #282, opened 2016-01-15
trim_image calls CCDData slicing even if fits_section is given.

MSeifert04:
From a discussion in https://github.com/astropy/ccdproc/issues/281#issuecomment-171936763.

I think it would be best to rely on ``CCDData.__getdata__`` instead of trying to the same inside the ``trim_image`` function itself.

### DIFF
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -299,11 +299,7 @@ def trim_image(ccd, fits_section=None):
     trimmed = ccd.copy()
     if fits_section:
         python_slice = slice_from_string(fits_section, fits_convention=True)
-        trimmed.data = trimmed.data[python_slice]
-        if trimmed.mask is not None:
-            trimmed.mask = trimmed.mask[python_slice]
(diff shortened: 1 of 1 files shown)


crawfordsm:
Thanks for the contribution @MSeifert04 !   I'm just thinking about whether this needs any tests as it was addressing a specific issue, but since there is already a WCS test in the unit tests and the NAXIS issue is an astropy.wcs issues, I don't think so.   Also since no functionality is changed, it doesn't need a changelog, so I'll test it out and then merge it.

MSeifert04:
@crawfordsm Sorry I was a bit sparse with the details: Since it wasn't found out as a bug I did not add it to the changelog. With test cases I wasn't sure since the only difference was with the ``wcs`` not being sliced and I don't know how to test different ``WCS`` for equality (since that is part of the original issue) - at least for someone like me, who has little experience with the ``WCS``-class.

crawfordsm:
Thanks @MSeifert04 -- merging

Issue #260, opened 2015-12-14
Add photutils as dependency


mwcraig:
Will be needed for using `astrometry.net`  to get WCS and for matching seeing between frames.

mwcraig:
Recommend that we close this. The astrometry.net pull request to astroquery takes care of importing photutils if needed.

crawfordsm:
I think, in the end, astrometry, will not be part of ccdproc.   So I agree, and closing this issue.

PR #284, opened 2016-01-15
Adding photoutils as dependency

souravsingh:


souravsingh:
This PR is related to issue #260 
Do we need to update to docs for this?

crawfordsm:
Thanks for the contribution @souravsingh !   I need to take care of a few other pull requests before this one.  At this time, ccdproc does not require this, but it will in the future so thanks for the PR.   We will also have to update the testing files as well when it does require this so that it is included, so there might be some further comments on this.

MSeifert04:
@souravsingh - Isn't it called``photutils``? You do mean https://photutils.readthedocs.org/en/latest/?

souravsingh:
My Bad! I am sorry for the mistake

MSeifert04:
Thank you @souravsingh but I'm going to close this for now. 

If we really need it as dependency it will be included in the relevant PR. We don't need to add dependencies "just in case".

Issue #283, opened 2016-01-15
Arithmetic between CCDData objects with WCS fails even with identical WCS


AnthonyHorton:
The CCDData arithmetic functions cannot be used with CCDData objects that have WCS even if the WCS are the same. Example:

    In [10]: ccd = ccdproc.CCDData.read(path, unit='adu')
    WARNING: FITSFixedWarning: 'datfix' made the change 'Changed '' to '2014-10-17''. [astropy.wcs.wcs]
    
    In [11]: ccd2 = ccd.copy()
    
    In [12]: result = ccd.subtract(ccd2)
    ---------------------------------------------------------------------------
    ValueError                                Traceback (most recent call last)
    ...
    ValueError: WCS properties do not match
    
The cause of the problem is upstream; `CCDData._ccddata_arithmetic` inherits from `astropy.nddata.NDArithmeticMixin._arithmetic` which checks for equivalence of the WCS of the arguments (`if self.wcs != operand.wcs: raise ValueError("WCS properties do not match"`) however this check always fails because `astropy.wcs.WCS` does not implement `__eq__()` or `__ne__` so this check is effectively a check for identity, not equivalence.  Demo:

    In [13]: w = ccd.wcs
    
    In [14]: w2 = copy.deepcopy(w)

    In [15]: w == w
    Out[15]: True
    
    In [16]: w == w2
    Out[16]: False

I'll file an issue with `astropy.wcs.WCS` and try to fix it there but felt I should also report it here.

dokeeffe:
When 204 gets fixed it will probably fix this too 
https://github.com/astropy/ccdproc/issues/204

AnthonyHorton:
@dokeeffe Huh, somehow didn't find #204 when I search for existing issues describing this problem. This is a closely related but slightly different issue, though, i.e. implementing `__eq__()` and `__ne__` in `astropy.wcs.WCS` would fix this but not #204 whereas removing the exception from `astropy.nddata.NDArithmetic._arithmetic` WCS equivalence check would fix both (in a somewhat unsatisfactory way).

MSeifert04:
@AnthonyHorton - I'm currently working on the ``NDArithmeticMixin`` to make the ``check_wcs`` optional. But I still think there should be an ``__eq__`` function in ``astropy.wcs.WCS`` since even if the check is optional it would be nice to have some way of comparing two ``WCS`` instances.

AnthonyHorton:
[...]

Issue #204, opened 2015-04-12
arithmatic on data with wcs


crawfordsm:
Current arithmatic between two `CCDData` objects is  not possible unless they have the exact same wcs.   This is due to a line in 
    if self.wcs != operand.wcs:

I'm not sure what that comparison entails, but it would seem that at times you would want to be able to subtract fits files that have different wcs values.    For example, when calibrating between calibration frames and science frames.

hamogu:
+1 fir a way to do arithmetic with non-matching WCSs. (On a pixel-by-pixel basis.)

dokeeffe:
+1 fir this. Flats and Darks will never have the same WCS as image frames.

mwcraig:
I'll have to look into this again...we are clearly able to do this in ``ccdproc.subtract_dark``, for example. Turns out the comparison is essentially meaningless anyway because `wcs.WCS` does not implement `__eq__`...

dokeeffe:
That would be great thanks! I'm basically trying to do `ccdproc.subtract_dark` using two `ccdproc.CCDData` instances which are loaded from fits files (master-dark and light files).

crawfordsm:
I'll be opening a small pull request at `NDArithmeticMixin`to try to remove this exception being raised for this as that is where the problem is.  However, if this isn't fixed and released there by the time we are ready with 0.4, we will implement meant a work around in ccdproc

crawfordsm:
Closed by #285

PR #285, opened 2016-01-19
Ignore wcs

crawfordsm:
For arithmetic between two CCDData objects, this will ignore any difference between WCS unless compare_wcs is set to True

### DIFF
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -112,11 +112,11 @@ class CCDData(NDDataArray):
     @data.setter
     def data(self, value):
         self._data = value
-        
+
     @property
     def wcs(self):
         return self._wcs
-    
+
     @wcs.setter
     def wcs(self, value):
         self._wcs = value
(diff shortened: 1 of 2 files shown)


crawfordsm:
Address #204 and #283 for `ccdproc`

MSeifert04:
I think the naming is quite good since in https://github.com/astropy/astropy/pull/4272 I'll introduce the same parameter. But with a slightly different mechanism. Please also note: https://github.com/astropy/astropy/pull/4503. I think your change will probably be no problem with either of these but better safe than sorry.

crawfordsm:
Thanks @MSeifert04, I did try to follow your naming as that would make future updates hopefully easier.  Unfortunately, I had to do something so that `CCDData` would work with astropy v1.0, otherwise I would have just updated it to use your PRs.   Hopefully this can be depreciated in the not too distant future.

mwcraig:
Looks good, merging. @AnthonyHorton this may be of interest to you.

PR #287, opened 2016-01-25
Astroscrappy cleanup

bsipocz:
This is to speed up travis a bit, and to mention the astroscrappy dependency in the docs.

Note that it's failing as no conda packages are available for older python/numpy versions. Waiting for the resolution of https://github.com/astropy/conda-builder-affiliated/issues/91 and then I'll either push another commit to install it from pip in those cases or there will be conda packages available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,8 @@ env:
         # to repeat them for all configurations.
         - NUMPY_VERSION=1.10
         - ASTROPY_VERSION=stable
-        - CONDA_DEPENDENCIES='scipy'
-        - PIP_DEPENDENCIES='pytest-capturelog astroscrappy'
+        - CONDA_DEPENDENCIES='scipy astroscrappy'
+        - PIP_DEPENDENCIES='pytest-capturelog'
 
     matrix:
         - SETUP_CMD='egg_info'
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,7 @@ addons:
             - dvipng
 
 python:
-    - 2.6
     - 2.7
-    - 3.3
     - 3.4
     - 3.5
 
@@ -56,12 +54,8 @@ matrix:
 
         # Try all python versions with the latest numpy available on that
         # python version.
-        - python: 2.6
-          env: NUMPY_VERSION=1.9 SETUP_CMD='test'
         - python: 2.7
           env: SETUP_CMD='test'
-        - python: 3.3
-          env: NUMPY_VERSION=1.9 SETUP_CMD='test'
         - python: 3.4
           env: SETUP_CMD='test'
         - python: 3.5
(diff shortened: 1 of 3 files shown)


mwcraig:
thanks @bsipocz, hope to have progress on that related issue soon. That, or we'll drop support for older pythons (2.6 and 3.3) and maybe for older numpys.

mwcraig:
@bsipocz -- could you please drop python 2.6 and 3.3 and numpy 1.7 and 1.8 from the test matrix? If you don't have time just let me know -- I can do it then restart the jobs that failed.

bsipocz:
@mwcraig - Done.

bsipocz:
@mwcraig - It looks like that recent changes in ``NDDataArray`` brakes the tests in ccdproc. Should I open an issue on it, or it's something that will be fixed soon?

mwcraig:
*sigh*

I though I had run the ccdproc test suite against the big NDData PR before merging it but apparently I missed this. Will take a look this evening, but am inclined to adjust the test.

MSeifert04:
@bsipocz - I had a look and some of these will be quite easy to solve (since they weren't intentional :-( ). The failure regarding ``compare_wcs`` on the other hand will be quite hard to resolve because there are several conflicting PRs: https://github.com/astropy/astropy/pull/4272, https://github.com/astropy/astropy/pull/4503 and https://github.com/astropy/ccdproc/pull/285.

bsipocz:
@mwcraig @MSeifert04 - What about fixing them to work with the dev version and xfail for older versions?

And this also brings up the question whether you think it would do any good to include a build against astropy dev in appveyor, too?

MSeifert04:
@bsipocz - I've made two PRs for the unintentional Failures on astropy, since they weren't intentional.

The remaining issue is a bit more complicated but xfailing it is probably not so good since it will need a real fix (not just a fix for the test). I'll think about it. Should I open an issue for it?

Is #285 included in a released ccdproc or only in dev?

bsipocz:
@MSeifert04 - Since it's milestoned to 0.4, which is unreleased, I think it's not yet released.

crawfordsm:
#285 is not currently included in a release (we are hoping to do a release at the end of the month).  I'll put in another PR which we can merge before this one to update the compare_wcs behavior to the one currently in NDData.

MSeifert04:
The one remaining failing test would be because of the setter of ``NDUncertainty``(https://github.com/astropy/astropy/pull/4558). It's a bit modified but your tests don't depend on the former shape check of the setter so if that PR is merged everything should pass again. I'm very sorry I missed these.

mwcraig:
Will merge once astropy/astropy#4558 passes.

bsipocz:
No worries @MSeifert04. You did a super big refactoring, and there are always things missed out in those.

crawfordsm:
Definitely agree with @bsipocz, nothing to apologize about and catching things like this are part of the process.   Thanks for getting the fixes in so quickly @MSeifert04

PR #290, opened 2016-02-02
Divide contained a Bug for compare_wcs=True

MSeifert04:
Division always set the other wcs to the one ``self`` had. This PR should fix this. The other arithmetics methods hadn't had this bug.


MSeifert04:
I forgot to include the modified test that had demonstrated the bug.

crawfordsm:
Sorry also only saw this one after my PR, but that should take care of this problem as well and include the testing fix.

MSeifert04:
Closing this PR in favor of #293.

PR #292, opened 2016-02-03
compare_wcs will not work with astropy-dev

MSeifert04:
For astropy v1.2 the refactoring of ``NDArithmeticMixin`` will produce problems with arithmetic operations of ``CCDData`` instances *when* one wants to compare the different ``WCS``. (see #287)

This solution is a not optimal and I'm a bit embarrassed to make this PR but I couldn't think of a better way to:
- Check which version of astropy is used. The ``hasattr(self, '_arithmetics_wcs')`` was the only way I found to distinguish between the NDData that compares the wcs and the NDData that needs a comparison function that compares different WCS'es
- Make the tests work without xfailing them and having a reminder for post-astropy-1.2 changes in there.

### DIFF
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -284,6 +284,11 @@ class CCDData(NDDataArray):
             nother = other.copy()
             if not compare_wcs:
                 nother.wcs = self.wcs
+            elif hasattr(self, '_arithmetic_wcs'):
+                # We want to compare the WCS but for astropy v1.2+ compare_wcs
+                # must be a callable which compares the wcs.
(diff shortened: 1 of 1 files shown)


crawfordsm:
Thanks for the PR,  As I mentioned in the other PR, I'd rather have `compare_wcs` behave the same way in astropy as it does in ccdproc and it was my earlier error to assume `compare_wcs` was going to be a only a boolean.

crawfordsm:
sorry I should have just commented on this PR instead of creating a new one myself to do something similar, but for some reason I am only seeing this now.

MSeifert04:
Closing this PR in favor of #293.

PR #293, opened 2016-02-03
New wcs arith

crawfordsm:
Update how compare_wcs is handled in ccddata arithmetic to match the current handling in the development version of astropy.

### DIFF
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -282,15 +282,25 @@ class CCDData(NDDataArray):
     def multiply(self, other, compare_wcs='first_found'):
         if isinstance(other, CCDData):
             if compare_wcs is None or compare_wcs == 'first_found':
-                nother = other.copy()
-                nother.wcs = self.wcs
-                return super(CCDData, self).multiply(nother)
+                tmp_wcs_1, tmp_wcs_2 = self.wcs, other.wcs
+                self.wcs, other.wcs = None, None
+
+                # Determine the WCS of the result
+                if compare_wcs is None:
+                    result_wcs = None
+                else:
+                    result_wcs = tmp_wcs_1 if tmp_wcs_1 else tmp_wcs_2 
+   
+                result = super(CCDData, self).multiply(other)
+                result.wcs = result_wcs
+                self.wcs, other.wcs = tmp_wcs_1, tmp_wcs_2
+                return result
             else:
-                try:
-                    return super(CCDData, self).multiply(
-                        other, compare_wcs=compare_wcs)
-                except:
-                    raise ValueError('WCS can not be compared')
+                if hasattr(self, '_arithmetics_wcs'):
+                    return super(CCDData, self).multiply(other,
+                                                    compare_wcs=compare_wcs)
+                else:
+                    raise ImportError("wcs_compare functionality requires astropy 1.2 or greater")
 
         return self._ccddata_arithmetic(other, np.multiply,
                                         scale_uncertainty=True)
@@ -298,15 +308,25 @@ class CCDData(NDDataArray):
     def divide(self, other, compare_wcs='first_found'):
         if isinstance(other, CCDData):
             if compare_wcs is None or compare_wcs == 'first_found':
-                nother = other.copy()
-                nother.wcs = self.wcs
-                return super(CCDData, self).divide(nother)
+                tmp_wcs_1, tmp_wcs_2 = self.wcs, other.wcs
+                self.wcs, other.wcs = None, None
+
+                # Determine the WCS of the result
+                if compare_wcs is None:
+                    result_wcs = None
+                else:
+                    result_wcs = tmp_wcs_1 if tmp_wcs_1 else tmp_wcs_2 
+   
+                result = super(CCDData, self).divide(other)
+                result.wcs = result_wcs
+                self.wcs, other.wcs = tmp_wcs_1, tmp_wcs_2
+                return result
             else:
-                try:
+                if hasattr(self, '_arithmetics_wcs'):
                     return super(CCDData, self).divide(other,
-                                                       compare_wcs=compare_wcs)
-                except:
-                    raise ValueError('WCS can not be compared')
+                                                    compare_wcs=compare_wcs)
+                else:
+                    raise ImportError("wcs_compare functionality requires astropy 1.2 or greater")
 
         return self._ccddata_arithmetic(other, np.divide,
                                         scale_uncertainty=True)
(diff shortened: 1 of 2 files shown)


crawfordsm:
I removed the asserts on the operations on the meta data as that has changed between 1.1 and the dev astropy versions in test_arithmetic_overload_ccddata_operand.

ccdproc/ccddata.py
```diff
@@ -279,43 +279,66 @@ def _ccddata_arithmetic(self, other, operation, scale_uncertainty=False):
...
-            nother.wcs = self.wcs
-            if not compare_wcs:
+            if compare_wcs is None or compare_wcs == 'first_found':
+                nother = other.copy()
                 nother.wcs = self.wcs
-            return super(CCDData, self).divide(nother)
+                return super(CCDData, self).divide(nother)
+            else:
+                try:
+                    return super(CCDData, self).divide(other,
```
MSeifert04: For astropy 1.1.2 and 1.0.9 the upstream-fix (https://github.com/astropy/astropy/pull/4503) allows comparing WCS objects but will raise a ``TypeError`` in case one tries to give it a ``compare_wcs`` (which only comes for astropy 1.2). That's the reason why I chose a different approach in #292.

MSeifert04:
The failure concerning ``meta`` is actually an oversight of my astropy-PR. I've provided a PR that will fix this: https://github.com/astropy/astropy/pull/4557.

ccdproc/ccddata.py
crawfordsm: was there an earlier version of` `compare_wcs` in ndarithmetic that was a boolean?  Or did I just make that up?   Since the ccdproc `compare_wcs` is new, I'd prefer to change it to have the same behavior as the astropy one

crawfordsm:
Also didn't see  https://github.com/astropy/astropy/pull/4557 so thought this bug was different behaviors but I'll revert the last commit of this PR

MSeifert04:
@crawfordsm - You would prefer merging this over #292 and #290? If yes, there might be some (but not too severe) problems.

crawfordsm:
My main thought right now is that we want the test suite to pass with the different supported version of astropy and I'd like `compare_wcs` to behave in the same way as it does in astropy, so changing it from boolean to having the multiple parameters.   

I'm not too bothered how it is done, and if you think it would be easier/better to use your PRs, then I'm happy to close this one.  Otherwise, if there are just some minor problems that are easily updated with this one, then I'm happy to continue with this one.  Feel free to either do a PR against my branch or add comments and I should be able to update them fairly quickly.   

Thanks for the help @MSeifert04 !

MSeifert04:
@crawfordsm @bsipocz @mwcraig - I've been thinking for some while now about a solution. But it boils down to a few questions:
- Do you want to make the ``compare_wcs`` of ``CCDData`` to look like the one used in the astropy 1.2 ``NDArithmeticMixin``, even though the results are different?
- Do you want to be able to do *any* kind of ``WCS`` comparison until then?
- Do you want to keep the overloaded ``add``, etc. and ``ccddata_arithmetics`` methods if the nddata refactor is released?

Depending on you answers I might be able to know how to help.

bsipocz:
@MSeifert04 - I'm totally neutral on this. (This may probably change when I pick up the project threads where I wanted to use ccdproc, but atm I really don't mind either solution).

crawfordsm:
Since ccdproc hasn't been released, my feeling is that:

* `compare_wcs` should look and behave as the same way as `compare_wcs` in astropy 1.2
* For other versions of astropy, `compare_wcs` should raise an error if the user actually tries to compare the wcs as that isn't very well supported at this time or the time scale at which astropy might release a patch.  
* Once we no longer support <1.1 and if we no longer have a need for the overloaded methods, it would be good to get rid of them.  However, that isn't likely anytime soon.

MSeifert04:
@crawfordsm - I don't know how to make a commit or PR against your branch. But a solution would be:

```
    def add(self, other, compare_wcs='first_found'):
        if isinstance(other, CCDData):
            if compare_wcs is not None and compare_wcs != 'first_found':
                # Using astropy 1.2 can be inferred if there is an wcs arithmetic method:
                if hasattr(self, '_arithmetics_wcs'):
                    return super(CCDData, self).add(other, compare_wcs=compare_wcs)
                else:
                    # Not using astropy 1.2
                    # Option 1: Just try it with broken comparison
                    return super(CCDData, self).add(other)
                    # Option 2: Do not compare ... ever ...
                    raise ValueError('WCS can not be compared')
                    # Option 3: Call the compare_wcs function and if it's true continue,
                    # otherwise raise an Exception.
                    if compare_wcs(self.wcs, other.wcs):
                        compare_wcs = 'first_found'  # So that it is evaluated correctly later
                    else:
                        raise ValueError('WCS can not be compared')
                    
            # We don't want any wcs comparison, so we set both wcs 
            # to None for the calculation
            # but store the current values so they can be assigned later.
            tmp_wcs_1, tmp_wcs_2 = self.wcs, other.wcs
            self.wcs, other.wcs = None, None
            
            # Determine the WCS of the result
            if compare_wcs is None:
                result_wcs = None
            else:
                result_wcs = tmp_wcs_1 if tmp_wcs_1 else tmp_wcs_2 
                    
            # Do the calculation without WCS
            result = super(CCDData, self).add(nother)
            # Set the wcs of the result (needs copy?)
            result.wcs = result_wcs
            # Reassign the WCS of the two operands
            self.wcs, other.wcs = tmp_wcs_1, tmp_wcs_2
            return result

        # Not a CCDData ...
        return self._ccddata_arithmetic(other, np.add, scale_uncertainty=False)
```
That's very lengthy and might be better off in a seperate method but I'll leave it to you if something like this would be appropriate.

Another very different possibiliy would be just to copy all of the ``NDArithmeticMixin`` of astropy-dev and insert it into ``CCDData``. :-)

bsipocz:
@MSeifert04 - You can add his fork as a remote, and checkout his branch. Make the changes, push to your github, and when opening a PR you can select his fork/branch from the dropdowns (rather than going with the default astropy/ccdproc and master as the base of the PR).

crawfordsm:
@MSeifert04 thanks for the suggestion.  I think I see the differences and I'll tweak my PR to account for those.

crawfordsm:
Okay I did two things differently:  if `_arithmetics_wcs` is missing, it raises an `ImportError` which I think gives a little more explicit information about what is wrong and why the function failed and removed the testing for the raising of errors as that seems unduly complex.   

If it passes its tests ignoring anything else that might fail and there are no other thoughts, I'll go ahead and merge this and then open an issue to say that at someday, these should be updated once we no longer support the older versions of astropy

MSeifert04:
I think a ``NotImplementedError`` would be more appropriate than an import error but everything else looks ok :1

crawfordsm:
I was going back and forth with the `ImportError` and the `NotImplementedError` as well as neither seems to fit exactly what we want to imply as described in https://docs.python.org/2/library/exceptions.html.   So if you had strong feelings, I would be happy to change.

MSeifert04:
I've read through the same page again and reached the same conclusion so I've deleted that part of the comment before you answered. Sorry.

I also think ``ImportError``, ``NotImplementedError`` and ``ValueError`` aren't exactly suitable so I think ``ImportError`` is ok.

Issue #294, opened 2016-02-04
test failure with numpy 1.11.0 beta 2


olebole:
With the new beta release of numpy (which just entered Debian unstable), I get the test failures due to using `numpy.float64` as index:
```
_____________________________ test__blkavg_larger ______________________________

ccd_data = CCDData([[ -1.08563060e+00,   9.97345447e-01,   2.82978498e-01,
           -1....91506713e-01,   2.45379407e-01,
           -1.68426432e-01,   2.44026938e-01]])

    @pytest.mark.data_size(20)
    def test__blkavg_larger(ccd_data):
        a = ccd_data.data
>       b = _blkavg(a, (10, 10))

ccdproc/tests/test_ccdproc.py:535: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
ccdproc/core.py:838: in _blkavg
    return eval(''.join(evList))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   TypeError: 'numpy.float64' object cannot be interpreted as an index
```
This happens in `test__blkavg_larger`, `test_cosmicray_lacosmic`, `test_cosmicray_lacosmic_does_not_change_input`, `test_cosmicray_lacosmic_gbox`, `test_cosmicray_lacosmic_rbox`, and `test_cosmicray_lacosmic_ccddata`
Using non-integers was already deprecated in numpy-1.10; however since this is a regression, I opened an issue numpy/numpy#7162 (sunpy has the same problem).

crawfordsm:
closed by #295

PR #295, opened 2016-02-04
Fix non-integer indexes for future numpy versions

olebole:
This is a fix for #294.


crawfordsm:
Thanks @olebole !   We have some other test issues to sort out and updating the test suite but it is good to know about this issue before hand although I think I ran into somewhere else as well.   

Do you need this fix right away or can we sort out the other issues first?

olebole:
Dont hurry; I added a patch to the Debian version which will do it for the moment.

PR #297, opened 2016-02-09
Allow combination of images using WCS

### LINKED ISSUE #129: Combining data with WCS
A method that will combine data using only wcs information and handles error propagation correctly.

EDIT: See #133 so a separate issue for the error propagation.

mwcraig:
Closes #129 

Needs:
- [x] Documentation, including warning that flux conservation is not guaranteed (or at least that the user really ought to carefully check whether flux is conserved well enough for their purposes).
- [x] Proper handling of pixels in the result for which there was not a corresponding pixel in the input image.

### DIFF
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -23,7 +23,7 @@ __all__ = ['background_deviation_box', 'background_deviation_filter',
            'cosmicray_median', 'cosmicray_lacosmic',
            'create_deviation', 'flat_correct', 'gain_correct', 'rebin',
            'sigma_func', 'subtract_bias', 'subtract_dark', 'subtract_overscan',
-           'transform_image', 'trim_image', 'Keyword']
+           'transform_image', 'trim_image', 'wcs_project', 'Keyword']
 
 # The dictionary below is used to translate actual function names to names
 # that are FITS compliant, i.e. 8 characters or less.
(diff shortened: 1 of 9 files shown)


mwcraig:
Some questions:

+ How should we handle masks? [ignore?]
+ Propagating headers?
+ Method on` Combiner` or function?

mwcraig:
## Masking

Options:
+ Reproject the mask
+ Mask anything that was at one time masked
+ Do nothing

mwcraig:
## Better idea for adding reprojection

+ Add a reproject method on to `CCDData` instead of combiner, then can invoke that form `Combiner` or the `combine` function

crawfordsm:
:+1:

bsipocz:
So what happens with the mask in the normal case when combining images? In the first approximation I think we should do the same here.
(But on the long run it would be nice if the mask would behave as confidence maps rather than binary masks, thus we can do meaningful arithmetic with it).

mwcraig:
Right now a pixel in the combined image is masked only if that pixel was masked in all of the images that went into the combination.

I think the question really isn't so much how to combine the masks but how to reproject the mask.

crawfordsm:
Aslo, this is likely to be changed to be just a function focused on reprojection similar to the other functions in ccdproc.core, so to split the reprojection issues and the combining issues

Although @bsipocz you raise a good point about weighting and that should probably be opened as a separate issue for something to integrate in the future.

mwcraig:
@crawfordsm -- assuming this passes, it is ready for a look. One question remains, whose answer I think I know.

If a pixel in the output image requires information from a pixel in the input image that does not exist then those values are set to `nan`, and (I think) the value of the `footprint` returned by the reprojection is set to 0.

I'm thinking mask construction really should be 

(mask from interpolating the data's mask) `and` (mask from `nan`) values. 

Agreed?

crawfordsm:
Yes agreed to how the masks should behave

mwcraig:
@crawfordsm -- I think all the tests will pass now, so this is ready for review.

ccdproc/core.py
```diff
@@ -571,6 +572,88 @@ def transform_image(ccd, transform_func, **kwargs):
...
+    Given a CCDData image with WCS, project it onto a target WCS and
+    return the reprojected data as a new CCDData image.
+
+    Any mask, flags, weight, or uncertainty are ignored in doing the
+    combination.
+
+    Parameters
+    ----------
+    ccd : `~ccdproc.CCDData`
+        Data to be flatfield corrected
```
MSeifert04: ``projected`` instead of "flatfield-corrected"

ccdproc/core.py
```diff
@@ -571,6 +572,88 @@ def transform_image(ccd, transform_func, **kwargs):
...
+    target_wcs: `astropy.wcs.WCS` object
+
```
MSeifert04: Why the empty line before the parameter description? I think this empty line should be above ``target_wcs: ...`` also before ``:`` there should be a whitespace.

ccdproc/core.py
```diff
@@ -571,6 +572,88 @@ def transform_image(ccd, transform_func, **kwargs):
...
+        WCS onto which all images should be projected.
+
+    target_shape: two element list-like, optional
```
MSeifert04: Same here: Whitespace before the ``:``.

ccdproc/core.py
```diff
@@ -571,6 +572,88 @@ def transform_image(ccd, transform_func, **kwargs):
...
+        + 'nearest-neighbor'
+        + 'bilinear'
+        + 'biquadratic'
+        + 'bicubic'
+
+    {log}
+
+    Returns
+    -------
+    ccd :  `~ccdproc.CCDData`
```
MSeifert04: Here is a whitespace too much after ``:`` :-)

ccdproc/core.py
```diff
@@ -571,6 +572,88 @@ def transform_image(ccd, transform_func, **kwargs):
     return nccd
 
 
+@log_to_metadata
+def wcs_project(ccd, target_wcs, target_shape=None, order='bilinear'):
+    """
+    Given a CCDData image with WCS, project it onto a target WCS and
+    return the reprojected data as a new CCDData image.
+
+    Any mask, flags, weight, or uncertainty are ignored in doing the
```
MSeifert04: Is this a copy/paste error? I think at least the mask is not ignored during "reprojection" (instead of ``combination``)

MSeifert04:
I'll try to install this branch on my computer at the weekend. I've added some comments about the documentation but for the actual code I'll need to debug it - I haven't had much experience with ``reproject`` and I'm not sure I can follow the tests. @crawfordsm is probably better suited for reviewing the code in detail.

crawfordsm:
add a `_short_names` for reproject

mwcraig:
Comments have all been addressed, I think.

Issue #219, opened 2015-07-11
Make docstring for `CCDData.read` more helpful


mwcraig:
Right now it seems to be using the docstring from the base metaclass.

PR #302, opened 2016-02-11
writing multi extension fits files with mask, uncertainty

### LINKED ISSUE #196: CCDData written as MEF FITs
Since a CCDData object can include multiple planes (data, uncertainty, masks), it would be good if these could all be written out at the same time in order to save the data as a multi-extension FITS file

MSeifert04:
Fixes #196

### DIFF
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -502,19 +502,19 @@ def fits_ccddata_reader(filename, hdu=0, unit=None, hdu_uncertainty='UNCERT',
         unit for the image in the FITS header (the keyword ``BUNIT`` is used
         as the unit, if present), this argument is used for the unit.
 
-    hdu_uncertainty : str or int, optional
+    hdu_uncertainty : str or None, optional
         FITS extension from which the uncertainty should be initialized. If the
         extension does not exist the uncertainty of the CCDData is ``None``.
         Default is ``'UNCERT'``.
 
-    hdu_mask : str or int, optional
+    hdu_mask : str or None, optional
         FITS extension from which the mask should be initialized. If the
         extension does not exist the mask of the CCDData is ``None``.
         Default is ``'MASK'``.
 
-    hdu_flags : str or int, optional
+    hdu_flags : str or None, optional
         Currently not implemented.
-        Default is None.
+        Default is ``None``.
 
     kwd :
         Any additional keyword parameters are passed through to the FITS reader
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -206,10 +206,14 @@ class CCDData(NDDataArray):
            Raises
            -------
            ValueError
-              Multi-Exenstion FITS files are not supported
+              - If ``self.mask`` is set but not a `~numpy.ndarray`.
+              - If ``self.uncertainty`` is set but not a
+                `~astropy.nddata.StdDevUncertainty`.
+              - If ``self.uncertainty`` is set but has another unit then
+                ``self.data``.
 
            NotImplementedError
-              Saving flags is not supported
+              Saving flags is not supported.
 
            Returns
            -------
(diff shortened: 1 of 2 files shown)


ccdproc/ccddata.py
```diff
@@ -466,46 +532,61 @@ def fits_ccddata_reader(filename, hdu=0, unit=None, **kwd):
...
-    # Try constructing a WCS object. This may generate a warning, but never
-    # an error.
-    wcs = WCS(hdr)
-    # Test for success by checking to see if the wcs ctype has a non-empty
-    # value.
-    wcs = wcs if wcs.wcs.ctype[0] else None
-    ccd_data = CCDData(hdus[hdu].data, meta=hdr, unit=use_unit,
-                       wcs=wcs)
-    hdus.close()
+    with fits.open(filename, **kwd) as hdus:
```
MSeifert04: I hope you don't mind that I switched this one to the context manager. I had some problems with Errors and that file handles on Windows are still open because it did not reach the ``hdus.close()``.

ccdproc/ccddata.py
```diff
@@ -189,14 +189,28 @@ def uncertainty(self, value):
...
+              If it is a string append this attribute to the HDUList as
+              `~astropy.io.fits.ImageHDU` with the string as extension name.
+              Flags are not supported at this time. If ``None`` this attribute
+              is not appended.
+              Default is ``'MASK'`` for mask, ``'UNCERT'`` for uncertainty and
+              ``None`` for flags.
+
            Raises
            -------
            ValueError
```
crawfordsm: Actually this could be removed as it isn't relevant here, but the ValueErrors rased for bad mask or uncertainty frames should be mentioned here

ccdproc/ccddata.py
```diff
@@ -466,46 +532,61 @@ def fits_ccddata_reader(filename, hdu=0, unit=None, **kwd):
...
+        else:
+            uncertainty = None
+
+        if hdu_mask in hdus:
+            # Mask is saved as uint but we want it to be boolean.
+            mask = hdus[hdu_mask].data.astype(np.bool_)
+        else:
+            mask = None
+
+        if hdu_flags in hdus:
```
crawfordsm: Probably be better to make this a warning rather than an error as the user most likely still wants the file to be opened.

ccdproc/ccddata.py
crawfordsm: :+1:

ccdproc/ccddata.py
MSeifert04: If the user explicitly overrides the default value (```None``) then an Error is in my opinion more appropriate. With the default value the ``if`` will evaluate to ``False``.

crawfordsm: ah good point.   Since the default is None, then that should be fine

crawfordsm:
Overall looks good.   I think I just had the one first comment and then it would look good to go (that is does now support MEF files and writing it out).  Thanks for the contribution @MSeifert04 !

MSeifert04:
No problem. I've pushed a change concerning the ``ValueError``.

There were some things I haven't checked:
- Coverage makes strange things on my computer and just won't run. So I don't know how the coverage is. I know that I haven't checked the ``Errors`` because I wasn't sure if they are senseable. But I think I got almost everything else (but I can't check)
- ``fits_ccddata_reader`` isn't included in the API-documentation. I actually don't know how to include it since it's just the function that is registered with the ``CCDData.read`` method (which is what should be included in the documentation).
- same for ``CCDData.write`` and ``fits_ccddata_writer``.

And ``hdu_mask``, etc. can only be ``strings``. Numerical values are not possible. That's a drawback but I hope not too much. I've added a small commit concerning the (not visible) documentation. :-)

crawfordsm:
Thanks @MSeifert04.  I am merging this but can you add an issue on the missing documentation or add a comment to #219

PR #305, opened 2016-02-15
include read and write method documentation to CCDData [skip ci]

### LINKED ISSUE #304: CCDData.read and write API documentation is missing
See also https://github.com/astropy/ccdproc/pull/302#issuecomment-183057798.

MSeifert04:
fixes #304 

I'm not sure if that's the best way but it at least makes the documentation of ``CCDData.read()`` and ``CCDData.write`` accessible.

### DIFF
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -76,7 +76,7 @@ class CCDData(NDDataArray):
         ``Classmethod`` to create an CCDData instance based on a ``FITS`` file.
         This method uses :func:`fits_ccddata_reader` with the provided
         parameters.
-    writeto(*args, **kwargs)
+    write(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


ccdproc/ccddata.py
```diff
@@ -70,6 +70,18 @@ class CCDData(NDDataArray):
         If the ``uncertainty`` or ``mask`` inputs cannot be broadcast (e.g.,
         match shape) onto ``data``.
 
+    Methods
+    -------
+    read(*args, **kwargs)
+        ``Classmethod`` to create an CCDData instance based on a ``FITS`` file.
+        This method uses :func:`fits_ccddata_reader` with the provided
+        parameters.
+    writeto(*args, **kwargs)
```
crawfordsm: this should be write

PR #306, opened 2016-02-15
fixes add/subtract arithmetics with quantities of different unit.

### LINKED ISSUE #291: CCDData Arithmetics (add/subtract) with Quantity
I think something is wrong when using CCDData arithmetics with Quantities (at least addition and subtraction):

```
import ccdproc
import astropy.units as u
ccdproc.CCDData(10, unit='meter').add(1*u.cm).data
#Out: 11
ccdproc.CCDData(10, unit='meter').add(1*u.cm).unit
#Out: Meter
```

This should be 10.01 meters.

MSeifert04:
Fixes #291.

I've included a test and changelog entry.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,10 @@ Other Changes and Additions
 Bug Fixes
 ^^^^^^^^^
 
+- Adding/Subtracting a CCDData instance with a Quantity with a different unit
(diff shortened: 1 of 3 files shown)


crawfordsm:
Thanks @MSeifert04

PR #311, opened 2016-02-18
Scale uncertainty with the number of combined pixel during combination.

### LINKED ISSUE #309: Combiner and resulting uncertainty
I've used the ``ccdproc`` combiner today and compared it to what I had for results using my custom combination (before I found out about ccdproc) and noticed the uncertainty of the ``ccdproc``-combine result is much higher than my custom one. After some investigation I found a possible explanation:

I've always used ``1/sqrt(n-1) * np.std(...)`` instead of ``np.std(...)``. I always thought the standard deviation would be the uncertainty of a single value and the uncertainty of the combined values should be divided by the square-root of the number of combined points (corrected for finite population).

If that's not the case this issue can be closed but I wasn't sure and searching the internet didn't help either.

MSeifert04:
Fixes #309 

Please review this carefully.

Some (remaining) open issues:
- Since you did not pass ``ddof`` to ``np.ma.std`` I didn't correct for finite populations either and just divided by ``np.sqrt(valid_pixel)``. This might not be exactly correct but I'm not too sure about statistics. The downside with such an finite population correction is also that the number of valid pixel could be 0 and then subtracting 1 and taking the square root ... not good.
- both combiners ignore scaling and weighting for the uncertainty computation.
- median combine ignores the masked values (probably because ``median_absolute_deviation`` cannot handle these. Therefore I corrected just with the whole number of pixel because these are used for uncertainty computation. It's inconsistent but I don't have time right now to validate what needs to be done to make that take masked arrays.

If you have any suggestions regarding these issues I would be very happy to include these and if not I can raise another issue for these.

### DIFF
--- a/ccdproc/tests/test_combiner.py
+++ b/ccdproc/tests/test_combiner.py
@@ -258,7 +258,7 @@ def test_combiner_scaling_fails(ccd_data):
 
 
 #test data combined with mask is created correctly
-def test_combiner_mask_media(ccd_data):
+def test_combiner_mask_median(ccd_data):
     data = np.zeros((10, 10))
     data[5, 5] = 1
     mask = (data == 0)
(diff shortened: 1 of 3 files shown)


crawfordsm:
@MSeifert04 This pull request should probably hold off until #258 is merged, which is very close and will likely have a little bit of overlap.

MSeifert04:
@crawfordsm : No problem. Maybe the other PR solves some of the issues mentioned above. :-) Haven't had time to look at it.

crawfordsm:
@MSeifert04 #258 has been merged, can you rebase if necessary and check to see what changes might need to be made?  Thanks for taking care of this.

MSeifert04:
@crawfordsm - Thanks for the notification. I rebased and adapted the tests from #258 to meet the scaling condition. I hope I did not mess up anything - I'm still feeling rather awkward when rebasing.

mwcraig:
@MSeifert04 -- awkward is  the normal way to fell when rebasing :grin:

ccdproc/tests/test_combiner.py
```diff
@@ -332,16 +332,52 @@ def test_combine_limitedmem_scale_fitsimages():
...
     ccd_list = [ccd_data, ccd_data, ccd_data]
     c = Combiner(ccd_list)
     ccd = c.median_combine(uncertainty_func=np.sum)
-    np.testing.assert_array_equal(ccd.uncertainty.array, np.sum(c.data_arr, 0))
+    uncert_ref = np.sum(c.data_arr, 0) / np.sqrt(3)
+    np.testing.assert_array_equal(ccd.uncertainty.array, uncert_ref)
+
+
+# test resulting uncertainty is corrected for the number of images
+def test_combiner_uncertainty_average(ccd_data):
```
mwcraig: Could drop the `ccd_data` argument since you don't use it. Same is true for the other new test below.

crawfordsm:
Ready to be merged after small comments from @mwcraig addressed

MSeifert04:
I've adressed the comment. Thanks for reviewing :+1:

crawfordsm:
Sorry @MSeifert04 , I merged #297 and that may have caused some small conflicts.  Can I ask you to rebase and then I'll merge?  Sorry for the extra delay and thank you once again!

MSeifert04:
of course

PR #313, opened 2016-02-26
Create AUTHORS.rst

souravsingh:
Added a file for Contributors of the Project.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -1,0 +1,36 @@
+*******************
+Authors and Credits
+*******************
+
+ccdproc Project Contributors
+============================
+
+* Matt Craig(@mwcraig)
+* Steve Crawford(@crawfordsm)
+* Thomas Robitaille(@astrofrog)
+* Brigitta Sipocz(@bsipocz)
+* Joe Philip Ninan(@indiajoe)
+* Michael Droettboom(@mdboom)
+* Erik Bray(@embray)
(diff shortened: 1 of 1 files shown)


crawfordsm:
Thanks @souravsingh.   I think we will need to do a clean up of the git log as the commits related to astropy-helpers should probably be removed as they are not direct contributions to ccdproc.

bsipocz:
@crawfordsm - I don't think it's necessary to hack the log and rewrite history.

The commits are coming from the ``package-template`` and not from ``astropy-helpers``. Maybe the easiest would be to cross match the commit hashes against the template. The ones that don't match are ccdproc only.

bsipocz:
Also, all commits before 9185b8bb2da61d85db038b19ad41a3e13d51907b seems to be coming from the template.

crawfordsm:
@souravsingh can you implement @bsipocz suggestion and see what the list returns?  All the wording should be updated to reflect this is for ccdproc and not astropy

souravsingh:
@crawfordsm I will get on to it.

mwcraig:
Merging as-is, will open a separate PR in a few minutes to clean up the list.

mwcraig:
Thanks, @souravsingh!

souravsingh:
Thanks for merging the PR @mwcraig

MSeifert04:
@souravsingh - I think you missed the ``@`` when crediting hamogu.

PR #324, opened 2016-03-13
Add axis keyword to call to median_absolute_deviation

mwcraig:
When the option of specifying an `uncertainty_function` was added I think this was overlooked.

I'd be inclined to include it in 0.4 and so have milestoned it there. No changelog entry because the real change (adding the `uncertainty_function`) is included already in #323


crawfordsm:
Merged, but also added a new issue in https://github.com/astropy/astropy/issues/4688 to add this to astropy.stats and then we can depreciate sigma_func

PR #325, opened 2016-03-14
Add slash so full path is correctly returned

EP-Guy:
Without the slash, the filename is appended directly to the path string.

E.g. for the path returned by self._location: `C:\\Users\\EP_Guy\\images`, and an image file `M42_001.fits`, `filtered_files` returns `C:\\Users\\EP_Guy\\imagesM42_001.fits`

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -257,7 +257,7 @@ class ImageFileCollection(object):
         filtered_files = self.summary_info['file'].compressed()
         self.summary_info['file'].mask = current_file_mask
         if include_path:
-            filtered_files = [self._location + f for f in filtered_files]
+            filtered_files = [self._location + '/' + f for f in filtered_files]
         return filtered_files
 
     def refresh(self):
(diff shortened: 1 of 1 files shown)


crawfordsm:
Thanks for the pull request.  There is likely a better way to fix this by creating a location setter and having check that the location is a complete path and/or raising an error when an uncompleted path is passed.  This also requires a test.   Would you be able to make those changes today?

mwcraig:
@EP-Guy -- great catch!

I think the correct fix here is

```python
filtered_files = [os.path.join(self._location, f) for f in filtered_files]
```

`os.path.join` will ensure that the correct separator is used on mac, linux and windows.

mwcraig:
@EP-Guy -- if you don't have time today please let us know; we can add your commit so you get credit then add tests/etc. Once again, thanks for catching, reporting and proposing a fix!

EP-Guy:
@mwcraig -- Thanks! I've changed it to your suggested fix.

mwcraig:
@EP-Guy -- looks like we need an `import os` at the top of `image_collection.py`

EP-Guy:
@mwcraig -- my console had `os` imported already, so I didn't catch that.  

Anyways, `image_collection.py` already makes use of `path.join` elsewhere and imports `path` from `os` directly. Therefore, I simply removed the `os.` from the call to `join`.

mwcraig:
Thanks @EP-Guy, merging!

@crawfordsm -- separate PR coming for test, contributors update, etc

PR #335, opened 2016-03-17
adding .fts support to image_collection.py

jmccormac01:
I tried running image_collection on my data but the 3rd fits extension is not recognised. .fts is also used as well as .fits and .fit. Seems to be a simple fix. I could also just rename my files but this should make the extension grabbing complete.

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -527,7 +527,7 @@ class ImageFileCollection(object):
         ----------
         extension : list of str, optional
             List of filename extensions that are FITS files. Default is
-            ``['fit', 'fits']``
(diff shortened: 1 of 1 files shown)


crawfordsm:
Thanks for the contribution @jmccormac01 I have not used '.fts' extensions before, but if it is helpful to you, then it is easy to include.   The failure isn't related to your PR so I have restarted and hopefully it will pass once it runs successfully.

jmccormac01:
It is used quite infrequently in general but several of the instruments I use seem to like .fts :) Thanks for including it.

mwcraig:
Thanks @jmccormac01!

Issue #336, opened 2016-03-18
meta attribute is inconsistent


evertrol:
The fact that the `meta` attribute can be `dict`, an `OrderedDict` or a `astropy.io.fits.Header` instance has irked me on occasions. In particular because the three types are incompatible with each other.

This means at times that one has to either do a manual conversion, or test the type of the meta attribute, before performing some operations. Case sensitivity may be one issue to run into (issue #156 indicates `meta` should be case sensitive); _comment_ or _history_ keywords may be another thing.
(That same issue also indicates a willingness to move away from FITSness of `CCDData`, though this may not always be doable.)

My suggestion would be use a standard `dict` only. This means a loss of information converting from an `OrderedDict`, but I doubt that that will be really an issue.
For a `Header` instance, there would be a need to decide on 
- history and comment keywords. I would probably create a separate issue to make these list of strings in special `meta` keys, if there is an overall approval of this issue. Note: `astropy.table.Table` has `meta["comments"]` (note the plural key), so it'd be good to use that approach.
- should the keywords be lower-cased, or stay upper-case (I don't like using CAPITALS in my code for various reasons, so I prefer lower-case).

For going from a `dict` to a `Header` when writing to a FITS file or HDU, I suggest issuing a warning in case duplicate keywords are found (and thus lost). This would be the case for `dict` keys that are the same except for their case. Currently, this is silently ignored and it is random what is written to the actual `Header`.
_history_ and _comment_ keywords can be written with the above list approach (arguably, handling "history" and "comments" keys in `meta` differently is a FITSness; this is probably something I can live with, in particular because upon writing to FITS, any value can be written as a (1 element) list of strings).

Perhaps this `meta` inconsistency is a non-issue for the majority of people (in which case this issue can be quickly closed), but otherwise I'd be happy to create a PR for this. There may be the downside of some loss of backward compatibility, for example, if people have code that uses something like `ccd.meta.add_history("blah blah")`, but my feel is that that is minimal.

crawfordsm:
I've typically mainly used the meta attribute basically as a FITS header and so I am hesitant to move too far away from that.   From my experience, this is what the average user will expect (as they will be opening up data from a FITS file I would guess)  but that is totally due to my own bias.  

I'd agree that having an object that can operate in multiple different ways is bad for the downstream user.   However, what the best way that object should be I'm not sure about and would seem like a major change to the API so at the very least, something for a version 2.0 release.  

[...]

PR #339, opened 2016-04-06
Make meta and header attributes separate objects

evertrol:
This PR is to provide an example for the discussion started in #336 

This PR ensures the `meta` and `header` attributes of a `CCDData` object are always a `dict` and a `astropy.io.fits.Header`, respectively. Under the hood, they are kept in sync through their properties; one is updated if the other has previously been accessed. (Updating `_meta` or `_header` directly may still keep them in sync, but this is definitely not guaranteed.)
- It's a rather dirty hack. Not sure if this is really the way to go, seen the amount of extra code needed.
  There's also some extra code necessary in `CCDData.__init__.py`; see the relevant comment.
- I had to adjust some unit tests.  `Header`s keep their key order, and can be compared with a simple `==` in assertions, but `dict`s don't keep their key order and can mess up the `header` attribute key order. 
  Alternatively, 
  - an `OrderedDict` could be used (less generic)
  - the various `result.meta` assignments in `subtract_bias` and the like could be changed to update `result.header` instead.
- On the order of keywords and `OrderedDict`.
  I see there is even the possibility to use a `CaseInsensitiveOrderedDict` (from `utils.collections`, which is otherwise nowhere used). If a user needs one of these classes (to maintain key order and/or be case-insensitive in `meta`), it should be possible to use this as value to the `meta` attribute or parameter: the property checks for `isinstance(value, dict)`.
  I haven't tested this, but it could be used as a user option/preference.
- I've borrowed the short_names alternative from `_insert_in_metadata_fits_safe`.
- Comments and history are stored as lists in `meta`.
- There are probably a few lines/sections that can disappear, such as `if isinstance(self.header, fits.Header)` in `to_hdu`, and the `_insert_in_metadata_fits_safe` function
- doc strings and extra unit tests are still missing

### DIFF
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -98,9 +98,9 @@ class CCDData(NDDataArray):
     This is useful, for example, when plotting a 2D image using
     matplotlib.
 
-        >>> from ccdproc import CCDData   
+        >>> from ccdproc import CCDData
         >>> from matplotlib import pyplot as plt   # doctest: +SKIP
-        >>> x = CCDData([[1,2,3], [4,5,6]], unit='adu') 
+        >>> x = CCDData([[1,2,3], [4,5,6]], unit='adu')
         >>> plt.imshow(x)   # doctest: +SKIP
 
     """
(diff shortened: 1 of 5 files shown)


crawfordsm:
Okay I haven't had a chance to review it yet, but leaving a note here about astropy/astropy#4509

MSeifert04:
I think `meta` and `header` are quite ambiguous terms while `dict` but `astropy.io.fits.Header` are **not**. So having only a `dict` or the very special `fits.Header` is too restrictive in my opinion. 

For example `HDF5` uses a `AttributeManager`-`dict`-like class. I don't know how `h5py` reacts if I just give it a `dict` during writing or if I convert this `AttributeManager` to a `dict`. In general reading and just writing it to another file should never change the internals of the file (besides write-date and maybe checksum if that's depending on the time). Especially losing the ordering (if present) or the comments (fits headers) would be a no-go in my opinion.

The real problem only starts when writing to a file or using special cases. But that's just a matter of writing an appropriate `fits.io.registry.register_writer`?

mwcraig:
Closing, CCDData has now moved to astropy.nddata

PR #344, opened 2016-04-19
Fixes #343 by defaulting to OrderedDict in CCDData.meta.setter

### LINKED ISSUE #343: Regression test failure: CCDData.meta.setter
There is a test failure because of the newly merged https://github.com/astropy/astropy/pull/4509

It's pretty easy to fix:

- Delete `meta.setter` and rely on `NDData`
- Change `meta.setter` to use an `OrderedDict` (instead of `{}`) if `None`

Actually I find it rather strange that (before the newest astropy-master) during init `meta` defaults to `OrderedDict` but if manually set on the attribute it defaults to `dict`. Was there any specific reason?

MSeifert04:
See #343 

`CCDData.meta.setter` will initialize an empty `collections.OrderedDict` in case the new value is `None`. 

This is because `NDData` implements `meta`-checks as part of the setter itself (astropy >= 1.2) instead of during `__init__` (astropy < 1.2).

### DIFF
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -5,6 +5,7 @@ from __future__ import (absolute_import, division, print_function,
 
 import copy
 import numbers
+from collections import OrderedDict
 
 import numpy as np
 
@@ -168,7 +169,7 @@ class CCDData(NDDataArray):
     @meta.setter
     def meta(self, value):
         if value is None:
-            self._meta = {}
+            self._meta = OrderedDict()
         else:
             if hasattr(value, 'keys'):
                 self._meta = value


MSeifert04:
@crawfordsm I just found a note regarding this PR and that I didn't do a changelog because there was no appropriate section in the Changes.rst file.

Will this PR need a Changelog and if so should I add the section myself? and which version would come next?

crawfordsm:
Okay, I've updated the change log to include a 1.0.2 release which would be the next bug-fix release and the 1.1 release which would be the next major release.    This should probably go under 1.0.2 as it probably should be back dated to correspond to the astropy 1.2 release.

PR #354, opened 2016-05-30
Use axis when calculating median absolute deviation

javierblasco:
The axis should be passed to stats.median_absolute_deviation in order to behave as explained in the docstring.


mwcraig:
Oops, thought that had been fixed, thanks @javierblasco!

PR #358, opened 2016-06-04
Add iraf_minmax_clipping method to give IRAF-like minmax clipping.

### LINKED ISSUE #356: Question about minmax rejection in ccdproc.combine
I'm working on removing IRAF dependency in some code and I want to replace use of the pyraf/IRAF imcombine task.  I need to maintain compatibility with existing code, so I want to implement the rejection options in the same way as IRAF imcombine does.  

For the minmax rejection option, it looks to me as if IRAF and ccdproc treat this very differently.  In ccdproc, it masks pixels with a value above or below a specified threshold.  In IRAF imcombine, it looks to me as if they are removing nlow and nhigh pixels from each set of N values for each pixel in the combination (where N is the number of images being combined), so the number of pixels being averaged would be N-nlow-nhigh (see [IRAF imcombine help page](http://stsdas.stsci.edu/cgi-bin/gethelp.cgi?imcombine)).  Is this the correct interpretation?

If so, what would be the best way to proceed?  I'm happy to work on adding the appropriate functionality to ccdproc.

joshwalawender:
This is a quick attempt at adding the rejection method described in issue #356 

This rejection method needs a better name.  There could be confusion due to people coming from the IRAF "minmax" terminology.  I've put in a cumbersome name of iraf_minmax_clipping for now.

Note that this method changes the combiner object's self.data_arr in place by doing a numpy sort.  Is this a problem?  Will it interact badly with other parts of ccdproc?

Edit: closes #356

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -214,9 +214,9 @@ class Combiner(object):
             combination.
             Default is ``None``.
         """
-        if nlow == None:
+        if nlow is None:
             nlow = 0
-        if nhigh == None:
+        if nhigh is None:
             nhigh = 0
         nimages = self.data_arr.mask.shape[0]
         nx = self.data_arr.mask.shape[1]
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -214,19 +214,21 @@ class Combiner(object):
             combination.
             Default is ``None``.
         """
-        if nlow == None: nlow=0
+        if nlow == None: nlow = 0
         if nhigh == None: nhigh = 0
         nimages = self.data_arr.mask.shape[0]
-#         if nlow+nhigh >= nimages:
-#             raise ValueError("Can not reject more pixels"\
-#                              " than there are images to combine")
+        nx = self.data_arr.mask.shape[1]
+        ny = self.data_arr.mask.shape[2]
 
-        newmask = np.zeros(self.data_arr.mask.shape, dtype=bool)
-        ind = np.argsort(self.data_arr.data, axis=0)
+        self.argsorted = np.argsort(self.data_arr.data, axis=0)
 
-        newmask[(ind < nlow)] = True
-        newmask[(ind >= nimages-nhigh)] = True
-        self.data_arr.mask[newmask] = True
+        for x in range(nx):
+            for y in range(ny):
+                for i in range(nimages-nhigh,nimages):
+                    self.data_arr.mask[self.argsorted[i,x,y],x,y] = True
+                for j in range(0,nlow):
+                    self.data_arr.mask[self.argsorted[j,x,y],x,y] = True
+
 
     # set up min/max clipping algorithms
     def minmax_clipping(self, min_clip=None, max_clip=None):
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -181,15 +181,15 @@ class Combiner(object):
     def iraf_minmax_clipping(self, nlow=None, nhigh=None):
         """Mask pixels using an IRAF-like minmax clipping algorithm.  The
         algorithm will mask the lowest nlow values and the highest nhigh values
-        before combining the vlues to make up a single pixel in the reuslting
+        before combining the values to make up a single pixel in the resulting
         image.  For example, the image will be a combination of
         Nimages-low-nhigh pixel values instead of the combination of Nimages.
 
         Note that this differs slightly from the nominal IRAF behavior when
-        other masks are in use.  For example, if the nhigh>=1 and the highest
+        other masks are in use.  For example, if the nhigh>=1 and any
         pixel is already masked for some other reason, then this algorithm will
         count the masking of that pixel toward the count of nhigh masked pixels.
-        IRAF behaves slightly diifferently in this case (see IRAF help for that
+        IRAF behaves slightly differently in this case (see IRAF help for that
         behavior): http://stsdas.stsci.edu/cgi-bin/gethelp.cgi?imcombine
         
         Here is a copy of the relevant IRAF help text:
@@ -214,22 +214,18 @@ class Combiner(object):
             combination.
             Default is ``None``.
         """
-        self.data_arr.sort(axis=0)
+        if nlow == None: nlow=0
+        if nhigh == None: nhigh = 0
         nimages = self.data_arr.mask.shape[0]
-        newmask = np.ndarray(self.data_arr.mask.shape, dtype=bool)
+#         if nlow+nhigh >= nimages:
+#             raise ValueError("Can not reject more pixels"\
+#                              " than there are images to combine")
 
-        if nlow != None:
-            if nlow > nimages:
-                raise ValueError("Can not reject more pixels"\
-                                 " than there are images to combine")
-            for i in np.arange(0,nlow,1):
-                newmask[i,:,:] = True
-        if nhigh != None:
-            if nhigh > nimages:
-                raise ValueError("Can not reject more pixels"\
-                                 " than there are images to combine")
-            for i in np.arange(nimages-nhigh,nimages,1):
-                newmask[i,:,:] = True
+        newmask = np.zeros(self.data_arr.mask.shape, dtype=bool)
+        ind = np.argsort(self.data_arr.data, axis=0)
+
+        newmask[(ind < nlow)] = True
+        newmask[(ind >= nimages-nhigh)] = True
         self.data_arr.mask[newmask] = True
 
     # set up min/max clipping algorithms
@@ -619,7 +615,7 @@ def combine(img_list, output_file=None, method='average', weights=None,
 
     if iraf_minmax_clip:
         to_call_in_combiner['iraf_minmax_clipping'] = {'nlow': nlow,
-                                                  'nhigh': nhigh}
+                                                       'nhigh': nhigh}
 
     if minmax_clip:
         to_call_in_combiner['minmax_clipping'] = {'min_clip': minmax_clip_min,
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -200,6 +200,14 @@ class Combiner(object):
         self.data_arr.sort(axis=0)
         nimages = self.data_arr.mask.shape[0]
         newmask = np.ndarray(self.data_arr.mask.shape, dtype=bool)
+
+        if nlow > nimages:
+            raise ValueError(
+                "Can not reject more pixels than there are images to combine")
+        if nhigh > nimages:
+            raise ValueError(
+                "Can not reject more pixels than there are images to combine")
+
         if nlow != None:
             for i in np.arange(0,nlow,1):
                 newmask[i,:,:] = True
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -178,7 +178,7 @@ class Combiner(object):
             self._scaling = self.scaling[:, np.newaxis, np.newaxis]
 
     # set up IRAF-like minmax clipping
-    def iraf_minmax_clipping(self, nlow=0, nhigh=0):
+    def clip_extrema(self, nlow=0, nhigh=0):
         """Mask pixels using an IRAF-like minmax clipping algorithm.  The
         algorithm will mask the lowest nlow values and the highest nhigh values
         before combining the values to make up a single pixel in the resulting
@@ -445,7 +445,7 @@ class Combiner(object):
 
 def combine(img_list, output_file=None, method='average', weights=None,
             scale=None, mem_limit=16e9,
-            iraf_minmax_clip=False, nlow=1, nhigh=1,
+            clip_extrema=False, nlow=1, nhigh=1,
             minmax_clip=False, minmax_clip_min=None, minmax_clip_max=None,
             sigma_clip=False,
             sigma_clip_low_thresh=3, sigma_clip_high_thresh=3,
@@ -490,7 +490,7 @@ def combine(img_list, output_file=None, method='average', weights=None,
         Maximum memory which should be used while combining (in bytes).
         Default is ``16e9``.
 
-    iraf_minmax_clip : bool, optional
+    clip_extrema : bool, optional
         Set to True if you want to mask pixels using an IRAF-like minmax
         clipping algorithm.  The algorithm will mask the lowest nlow values and
         the highest nhigh values before combining the vlues to make up a single
@@ -498,8 +498,8 @@ def combine(img_list, output_file=None, method='average', weights=None,
         combination of Nimages-low-nhigh pixel values instead of the combination
         of Nimages.
 
-        Parameters below are valid only when iraf_minmax_clip is set to True,
-        see :meth:`Combiner.iraf_minmax_clipping` for the parameter description:
+        Parameters below are valid only when clip_extrema is set to True,
+        see :meth:`Combiner.clip_extrema` for the parameter description:
 
         - ``nlow`` : int or None, optional
         - ``nhigh`` : int or None, optional
@@ -614,8 +614,8 @@ def combine(img_list, output_file=None, method='average', weights=None,
         else:
             to_set_in_combiner['scaling'] = scale
 
-    if iraf_minmax_clip:
-        to_call_in_combiner['iraf_minmax_clipping'] = {'nlow': nlow,
+    if clip_extrema:
+        to_call_in_combiner['clip_extrema'] = {'nlow': nlow,
                                                        'nhigh': nhigh}
 
     if minmax_clip:
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -223,15 +223,16 @@ class Combiner(object):
         ny = self.data_arr.mask.shape[2]
 
         argsorted = np.argsort(self.data_arr.data, axis=0)
+        mg = np.mgrid[0:nx,0:ny]
         for i in range(nlow):
             where = (argsorted[i,:,:].ravel(),\
-                     np.mgrid[0:nx,0:ny][0].ravel(),\
-                     np.mgrid[0:nx,0:ny][1].ravel())
+                     mg[0].ravel(),\
+                     mg[1].ravel())
             self.data_arr.mask[where] = True
         for j in range(nimages-nhigh,nimages):
             where = (argsorted[j,:,:].ravel(),\
-                     np.mgrid[0:nx,0:ny][0].ravel(),\
-                     np.mgrid[0:nx,0:ny][1].ravel())
+                     mg[0].ravel(),\
+                     mg[1].ravel())
             self.data_arr.mask[where] = True
 
 
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -201,17 +201,16 @@ class Combiner(object):
         nimages = self.data_arr.mask.shape[0]
         newmask = np.ndarray(self.data_arr.mask.shape, dtype=bool)
 
-        if nlow > nimages:
-            raise ValueError(
-                "Can not reject more pixels than there are images to combine")
-        if nhigh > nimages:
-            raise ValueError(
-                "Can not reject more pixels than there are images to combine")
-
         if nlow != None:
+            if nlow > nimages:
+                raise ValueError("Can not reject more pixels"\
+                                 " than there are images to combine")
             for i in np.arange(0,nlow,1):
                 newmask[i,:,:] = True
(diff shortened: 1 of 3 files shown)


ccdproc/combiner.py
```diff
@@ -177,6 +177,61 @@ def scaling(self, value):
...
+        self.data_arr.sort(axis=0)
+        nimages = self.data_arr.mask.shape[0]
+        newmask = np.ndarray(self.data_arr.mask.shape, dtype=bool)
+
+        if nlow != None:
+            if nlow > nimages:
+                raise ValueError("Can not reject more pixels"\
+                                 " than there are images to combine")
+            for i in np.arange(0,nlow,1):
+                newmask[i,:,:] = True
```
MSeifert04: Instead of the loop you could also use ``newmask[:nlow, :, :]`` here. Same for the ``nhigh`` (``newmask[nimages-nhigh:, :, :]``) case.

ccdproc/combiner.py
```diff
@@ -177,6 +177,61 @@ def scaling(self, value):
...
+            Default is ``None``.
+
+        nhigh : int or None, optional
+            If not None, the number of high values to reject from the
+            combination.
+            Default is ``None``.
+        """
+        self.data_arr.sort(axis=0)
+        nimages = self.data_arr.mask.shape[0]
+        newmask = np.ndarray(self.data_arr.mask.shape, dtype=bool)
```
MSeifert04: I've never seen ``np.ndarray`` as constructor before. Is this safe to use? Will this initialize a new array with all zeros?

I would have used ``np.zeros(self.data_arr.mask.shape, dtype=bool)`` just to be explicit about the zeros.

ccdproc/combiner.py
```diff
@@ -177,6 +177,61 @@ def scaling(self, value):
             # reshape so that broadcasting occurs properly
             self._scaling = self.scaling[:, np.newaxis, np.newaxis]
 
+    # set up IRAF-like minmax clipping
+    def iraf_minmax_clipping(self, nlow=None, nhigh=None):
+        """Mask pixels using an IRAF-like minmax clipping algorithm.  The
+        algorithm will mask the lowest nlow values and the highest nhigh values
+        before combining the vlues to make up a single pixel in the reuslting
```
MSeifert04: "values [...] resulting"

ccdproc/combiner.py
```diff
@@ -177,6 +177,61 @@ def scaling(self, value):
...
+        image.  For example, the image will be a combination of
+        Nimages-low-nhigh pixel values instead of the combination of Nimages.
+
+        Note that this differs slightly from the nominal IRAF behavior when
+        other masks are in use.  For example, if the nhigh>=1 and the highest
+        pixel is already masked for some other reason, then this algorithm will
+        count the masking of that pixel toward the count of nhigh masked pixels.
```
MSeifert04: Actually ``np.ma.MaskedArray`` will "sort" masked pixel as if they contain huge values (not sure if fill-value is used or inf but they always ended up as highest pixel.

Maybe reformulate the sentence that already masked pixel will count towards the nhigh limit?

ccdproc/combiner.py
```diff
@@ -177,6 +177,61 @@ def scaling(self, value):
...
+        IRAF behaves slightly diifferently in this case (see IRAF help for that
```
MSeifert04: "differently"

ccdproc/combiner.py
```diff
@@ -177,6 +177,61 @@ def scaling(self, value):
...
+            If not None, the number of high values to reject from the
+            combination.
+            Default is ``None``.
+        """
+        self.data_arr.sort(axis=0)
+        nimages = self.data_arr.mask.shape[0]
+        newmask = np.ndarray(self.data_arr.mask.shape, dtype=bool)
+
+        if nlow != None:
+            if nlow > nimages:
```
MSeifert04: Maybe better to check nlow and nhigh together? Currently one could discard 10 nlow pixel and 10 nhigh pixel with 10 stacked images.

Is the ValueError meant as Warning or is it necessary because it would fail otherwise?

ccdproc/combiner.py
```diff
@@ -546,6 +617,10 @@ def combine(img_list, output_file=None, method='average', weights=None,
         else:
             to_set_in_combiner['scaling'] = scale
 
+    if iraf_minmax_clip:
+        to_call_in_combiner['iraf_minmax_clipping'] = {'nlow': nlow,
+                                                  'nhigh': nhigh}
```
MSeifert04: one small nitpicky comment here: The ``'nhigh' ...`` could be aligned with the ``'nlow': ...``.

ccdproc/combiner.py
```diff
@@ -177,6 +177,61 @@ def scaling(self, value):
...
+        -----------
+        nlow : int or None, optional
+            If not None, the number of low values to reject from the
+            combination.
+            Default is ``None``.
+
+        nhigh : int or None, optional
+            If not None, the number of high values to reject from the
+            combination.
+            Default is ``None``.
```
MSeifert04: I don't know if it's consistently used here but since you raise custom exceptions inside the method you could add another section here:

```
Raises
------
ValueError
    If ``nlow`` is bigger than the number of images ........
```

MSeifert04:
@joshwalawender I left some comments but I'm actually not sure if the in-place sorting is a good idea. It could make trouble when applying weights during combination - maybe also scaling.

It would be great having this method - but it needs some unittests. Especially together with weighting/scaling. It's hard to know if it really works without having some tests that cover simple and/or complicated cases.

ccdproc/combiner.py
joshwalawender: > Is the ValueError meant as Warning or is it necessary because it would fail otherwise?

What is the best behavior here?  I guess there's really no problem if we mask everything, then the combine result comes out as zero, so perhaps just a warning?  Is even a warning needed?

MSeifert04: It rather depends on preferences. I only raise explicit Exceptions if (a) it would result in unexpected and wrong behaviour later or (b) it would fail with a cryptic exception message later.

I would emit a Warning ``if nlow + nhigh > nimages``. Just to give users a hint why the result is completly masked and because the condition is cheap to check (one addition and one comparison with integers won't be a bottleneck if the method also contains a ``sort()`` call).

But that's just my opinion. In the end it's up to you (or the package maintainers) what should be raised and what not.

crawfordsm:
Thanks for the submission @joshwalawender and comments @MSeifert04 .  

I would agree that the sorting would be a problem and the other thing that would be needed is adding some tests.   One solution to the sorting is to create a deep copy of the data array although that could potentially run into memory problems, but may offer a solution in the mean time.

joshwalawender:
@MSeifert04 Thanks for the comments.  I've addressed all the points except for the sorting and testing (I'll push updates soon).

@MSeifert04 @crawfordsm I think I have a solution which does not sort in place (using np.argsort), but am still checking things out.  I'm going to try to put some tests in as well, but I am new to unit testing and pytest, so it'll take me a little while to work out how best to set up the tests.  I'll look at the existing tests to start with, but any guidance would be appreciated.

MSeifert04:
@joshwalawender Yes, ``np.ma.argsort`` would be a solution. 

I haven't tested it but I think something like this could work:
```
if nlow is None and nhigh is None:
    return None
argsorted = np.ma.argsort(self.data_arr, axis=0)
if nlow is None:
    newmask = (argsorted >= nimages-nhigh)
elif nhigh is None:
    newmask = (argsorted < nlow)
else:
    newmask = (argsorted < nlow) | (argsorted >= nimages-nhigh)
self.data_arr.mask[newmask] = True
```

joshwalawender:
@MSeifert04 Yes, that's basically the method I came up with too, but now I'm either getting weird behavior from argsort or else I don't understand what argsort is supposed to do.  Here's an example where I've stripped things down to the bare numpy rather than CCDData: https://gist.github.com/joshwalawender/8b170508bb765ecf316cf0584ed32299#file-test_argsort-py

If I run that file, argsorted is:

```
[[[2 2 2]
  [2 2 2]
  [2 2 2]]

 [[1 1 1]
  [1 1 1]
  [1 1 1]]

 [[3 3 3]
  [3 3 3]
  [3 3 3]]

 [[0 0 0]
  [0 0 0]
  [0 0 0]]]
```

I would expect it to be:

```
[[[3 3 3]
  [3 3 3]
  [3 3 3]]

 [[1 1 1]
  [1 1 1]
  [1 1 1]]

 [[0 0 0]
  [0 0 0]
  [0 0 0]]

 [[2 2 2]
  [2 2 2]
  [2 2 2]]]
```

Any idea what I'm missing?

MSeifert04:
@joshwalawender Oh, I'm sorry. It seems that argsort returns an index-array. That means simplified:

```
>>> a = np.array([90, 20, 10, 40])
>>> a.argsort()
array([2, 1, 3, 0], dtype=int64)
```
Explanation: The smallest element has index 2, then index 1, 3 and then the highest one has index 0.

That makes it quite tricky and probably just making a temporary copy like @crawfordsm suggested, would be easier (even if it's memory expensive).

joshwalawender:
@MSeifert04 Wow, I should have picked up on that.  Thanks.

joshwalawender:
I've got code working which uses argsort, though it is a bit inelegant (nested loops).  I've got one test going with a simple test case to make sure it works.  As this method only modifies the mask and not the data, it shouldn't affect other rejection methods.

ccdproc/combiner.py
```diff
@@ -177,6 +177,59 @@ def scaling(self, value):
...
+        """
+        if nlow == None: nlow = 0
```
MSeifert04: Again a nitpicky comment but one generally uses ``is`` when comparing to ``None`` and puts the "body" in the next line.
```
if nlow is None:
    nlow = 0
```

joshwalawender: Fixed.

MSeifert04:
@joshwalawender Could you provide some timings, say for a ``10x100x100 data_arr`` with nlow and nhigh 1, 2, 3? If you use IPython timings can be done with ``%timeit``:

```
import ccdproc

def test():
    combiner = ccdproc.Combiner([ccdproc.CCDData(np.full((100, 100), dtype=float, fill_value=np.random.randint(0, 100)), unit='adu') for _ in range(10)])
    combiner.iraf_minmax_clipping(nlow=1, nhigh=2)
%timeit test()
```

If these timings are relativly fast then also some timings with ``20x2048x2048`` would be nice (but with nested loops do the steps from 100x100 - 2048x2048 gradually before it takes half an hour). Just to have some reference point how long it would take with "real" data.

joshwalawender:
With a modified method (no more looping over pixels), I get `%timeit` results of around 2 seconds for combining 10 2048x2048 arrays.

```
In [2]: %timeit test(nx=1024, ny=1024, nlow=2, nhigh=2)
1 loop, best of 3: 403 ms per loop

In [3]: %timeit test(nx=2048, ny=2048, nlow=2, nhigh=2)
1 loop, best of 3: 2 s per loop

In [4]: %timeit test(nx=2048, ny=2048, nlow=1, nhigh=1)
1 loop, best of 3: 1.53 s per loop
```

MSeifert04:
Maybe one could save some more time by calculating the ``mgrid`` once and just reuse it instead of creating them multiple times. 

But the timings are very good :+1:

joshwalawender:
Yep, it does:

```
In [2]: %timeit test(nx=1024, ny=1024, nlow=2, nhigh=2)
1 loop, best of 3: 272 ms per loop

In [3]: %timeit test(nx=2048, ny=2048, nlow=2, nhigh=2)
1 loop, best of 3: 1.34 s per loop

In [4]: %timeit test(nx=2048, ny=2048, nlow=1, nhigh=1)
1 loop, best of 3: 1.23 s per loop
```

ccdproc/tests/test_combiner.py
```diff
@@ -393,3 +393,23 @@ def test_writeable_after_combine(ccd_data, tmpdir, comb_func):
     ccd2 = getattr(combined, comb_func)()
     # This should not fail because the resulting uncertainty has a mask
     ccd2.write(tmp_file.strpath)
+
+def test_iraf_minmax_masking():
+    ccdlist = [CCDData(np.ones((3, 5))*90., unit="adu"),\
```
MSeifert04: If there is some open bracket (``()``, ``{}`` or ``[]``) there is no need to use ``\`` at the end of the line.

ccdproc/tests/test_combiner.py
```diff
@@ -393,3 +393,23 @@ def test_writeable_after_combine(ccd_data, tmpdir, comb_func):
...
+    ccdlist[1].data[1,2] = 100.1
+    ccdlist[1].data[2,0] = 100.1
+    c = Combiner(ccdlist)
+    c.iraf_minmax_clipping(nlow=1, nhigh=1)
+    result = c.average_combine()
+    expected = [[ 30.,  22.5, 30. , 30., 30.],\
+                [ 30. , 30. , 47.5, 30., 30.],\
+                [ 47.5, 30. , 30. , 30., 30.]]
+    expected = np.array(expected)
+    assert (result == expected).all()
```
MSeifert04: There is some ``numpy`` testing utility for this ``np.testing.assert_array_equal`` (or ``np.testing.assert_all_close`` if there might be floating point inequality). Then you don't need to wrap ``expected`` as ``numpy.array``:

```
expected = [[30, ...], ...]
np.testing.assert_array_equal(result, expected)
```

ccdproc/combiner.py
```diff
@@ -177,6 +177,65 @@ def scaling(self, value):
...
+        if nhigh is None:
+            nhigh = 0
+        nimages = self.data_arr.mask.shape[0]
+        nx = self.data_arr.mask.shape[1]
+        ny = self.data_arr.mask.shape[2]
+
+        argsorted = np.argsort(self.data_arr.data, axis=0)
+        mg = np.mgrid[0:nx,0:ny]
+        for i in range(nlow):
+            where = (argsorted[i,:,:].ravel(),\
```
MSeifert04: Same here for the trailing ``\``. Should not be necessary.

ccdproc/combiner.py
```diff
@@ -177,6 +177,65 @@ def scaling(self, value):
...
+        nhigh : int or None, optional
+            If not None, the number of high values to reject from the
+            combination.
+            Default is ``None``.
+        """
+        if nlow is None:
+            nlow = 0
+        if nhigh is None:
+            nhigh = 0
+        nimages = self.data_arr.mask.shape[0]
```
MSeifert04: in theory it should be possible to use ``nimages, nx, ny = self.dara_arr.mask.shape`` (this will unpack all the shape values).

ccdproc/combiner.py
```diff
@@ -177,6 +177,65 @@ def scaling(self, value):
...
+        ny = self.data_arr.mask.shape[2]
+
+        argsorted = np.argsort(self.data_arr.data, axis=0)
+        mg = np.mgrid[0:nx,0:ny]
+        for i in range(nlow):
+            where = (argsorted[i,:,:].ravel(),\
+                     mg[0].ravel(),\
+                     mg[1].ravel())
+            self.data_arr.mask[where] = True
+        for j in range(nimages-nhigh,nimages):
```
MSeifert04: The inner part of the loop is exactly identical to the "nlow" one. Is it possible to combine both? Maybe something like: ``for i in range(-nhigh, nlow)``. Negative indices are interpreted as indexed from "the end". 
I've tested it on python3 and it could work (could you test it?)
```
>>> for i in range(-2, 2):
...     print(i)
-2
-1
0
1
```

MSeifert04:
I didn't expect it would yield such a speedup. :+1: 

I've included yet another set of comments (feel free to ignore them, they are mostly cosmetical). 

Some suggestion for tests:
- compare your existing test (manual rejection + combine) with the result of ``ccd_process``.
- one test including ccds with masks (maybe sequential rejections).

mwcraig:
FYI, test fails are due to a regression introduced in https://github.com/astropy/astropy/pull/5053 , which will be fixed in https://github.com/astropy/astropy/pull/5091.

Once that is merged I'll restart the failed tests here.

mwcraig:
Failing jobs have been restarted...

joshwalawender:
> one test including ccds with masks (maybe sequential rejections).

I've now got a test in place checking for proper results when there is another rejection method also being used (a simple manual rejection in this test).

> compare your existing test (manual rejection + combine) with the result of an ccd_process call

@MSeifert04 I'm not sure I understand what you mean by the above test.  Can you elaborate?

ccdproc/combiner.py
```diff
@@ -177,6 +177,58 @@ def scaling(self, value):
             # reshape so that broadcasting occurs properly
             self._scaling = self.scaling[:, np.newaxis, np.newaxis]
 
+    # set up IRAF-like minmax clipping
+    def iraf_minmax_clipping(self, nlow=None, nhigh=None):
+        """Mask pixels using an IRAF-like minmax clipping algorithm.  The
+        algorithm will mask the lowest nlow values and the highest nhigh values
+        before combining the values to make up a single pixel in the resulting
+        image.  For example, the image will be a combination of
+        Nimages-low-nhigh pixel values instead of the combination of Nimages.
```
mwcraig: low → nlow

ccdproc/combiner.py
```diff
@@ -177,6 +177,58 @@ def scaling(self, value):
             # reshape so that broadcasting occurs properly
             self._scaling = self.scaling[:, np.newaxis, np.newaxis]
 
+    # set up IRAF-like minmax clipping
+    def iraf_minmax_clipping(self, nlow=None, nhigh=None):
```
mwcraig: would default values of 1 for these make more sense? That would also eliminate the `if`s below that check for `None`.

mwcraig:
@joshwalawender -- thanks for your work on this! Just a couple minor comments. If you prefer, since this has been ongoing for a while, I can merge this as is then do a separate PR to make these minor adjustments.

ccdproc/combiner.py
joshwalawender: I debated this.  My feeling is that I'd rather default to 0 or None (no rejection) because (thinking as a user) if I call this and only specify nhigh=1 with no call to nlow, I would expect one value to be rejected, not two.

I prefer to leave the None check in because I could see a user specifying that to try to turn off low or high rejection.  I think it makes more sense for the defaults should be 0, not None though, so I'll change that.

joshwalawender:
We've still got the (to me) rather cumbersome name of "iraf_minmax_clipping" and "iraf_minmax_clip".  Any suggestions for better naming or should we go with this?

MSeifert04:
What about `clip_extrema`?

@joshwalawender sorry, I missed your question before: I though about a test where you call [`combine`](https://github.com/astropy/ccdproc/blob/master/ccdproc/combiner.py#L394) to do the clipping rather than the method directly.

joshwalawender:
@MSeifert04 Got it.  We're just testing another path through the same code.  Makes sense.  I've added it.

joshwalawender:
I think `clip_extrema` is an improvement on `iraf_minmax_clip`.  Other ideas?

mwcraig:
`clip_extrema` sounds good to me too, though it brings to mind this: https://youtu.be/dhRUe-gz690?t=2m5s

mwcraig:
@joshwalawender -- one last question, which should have occurred to me earlier. Do you have time to add this to the docs at https://github.com/astropy/ccdproc/blob/master/docs/ccdproc/image_combination.rst ? Last change, I think -- and I can take care of it if you don't have time.

joshwalawender:
Renamed to clip_extrema and added documentation.

docs/ccdproc/image_combination.rst
```diff
@@ -74,6 +74,19 @@ deviations below the median with
 
     A
 
+Extrema clipping
+++++++++++++++++
+
+For each pixel position in the input arrays, the algorithm will mask the
+highest `nhigh` and lowest `nlow` pixel values.  The resulting image will be a
```
mwcraig: @joshwalawender -- the single backticks around nhigh, nlow, Nimages-nlow-nhigh, Nimages should be changed to double backticks. In Sphinx, singe backticks are references, double backticks are code font.

Looks like this was the only error in the travis build!

mwcraig:
@crawfordsm -- this is ready to merge, I think, assuming you are OK with the name `clip_extrema`.

PR #363, opened 2016-06-23
Ensure uncertainty is scaled properly in flat_correct

### LINKED ISSUE #345: flat_correct doesn't normalize flat field for uncertainty array
When using flat_correct on an image with an uncertainty array the flat field is normalized before the data is divided through it, but not for the uncertainty array.

mwcraig:
This should fix #345 by doing a slight rearrangement of the calculations in `flat_correct`


mwcraig:
@crawfordsm -- this and #345 are milestoned to 1.0.2. Should we skip right to 1.1 instead? Changelog for this in a moment...

PR #365, opened 2016-06-27
Remove pytest-caplog plugin

mwcraig:
This tests the log only using the standard library.


crawfordsm:
Will this have any conflicts with #367 ?  Doesn't look like it initially but I'll let you merge these when you are ready.

bsipocz:
@crawfordsm - I don't think it will, but even if it is then I can rebase #367.

mwcraig:
Merging...

Issue #361, opened 2016-06-18
ccdproc fails with error "operand units do not match"


mrkite92:
[Archive.zip](https://github.com/astropy/ccdproc/files/321651/Archive.zip)

Hi ccdproc gurus,

I work on Mac with latest version of OS and Python 2.7 ressources linked to ccdproc. I wrote a short script to deflat a few images, and I got:

Traceback (most recent call last):
  File "deflat.py", line 34, in <module>
    data_exposure = img_expos)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/ccdproc/log_meta.py", line 68, in wrapper
    result = func(*args, **kwd)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/ccdproc/core.py", line 208, in ccd_process
    nccd = nccd.subtract(master_bias)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/ccdproc/ccddata.py", line 444, in subtract
    result = super(CCDData, self).subtract(other)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/astropy/nddata/mixins/ndarithmetic.py", line 215, in subtract
    operand, propagate_uncertainties, "subtraction", np.subtract)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/astropy/nddata/mixins/ndarithmetic.py", line 87, in _arithmetic
    raise ValueError("operand units do not match")
ValueError: operand units do not match

After digging into the code, I strongly suspect the value of the gain (2.35 electrons/adu) not be to properly applied. In line 349 of ccddata.py, the unit to be used (result_unit) is 'electron'. Then, in function subtract, line 433, other and self have different units. I attach here my script (dir_wrk to be modified) and relevant images.

mwcraig:
@mrkite92 -- thanks for the report, I'll take a look on Monday.

mwcraig:
Could you please let me know what version of astropy and ccdproc you are using?

mrkite92:
Hi Matt,

My ccdproc is 1.0.1, and astropy is 1.1.2

Cheers,
JB

mwcraig:
[...]

PR #366, opened 2016-06-28
Address units expected for calibration images in ccd_process

mwcraig:
This pull request address #361 in two ways:

1. The documentation for `ccd_process` now clearly states what the expected units of the calibration images are.
2. I've implement, for `subtract_bias`, a more explicit check for a units mismatch and raise a `UnitsError` instead of a `ValueError` in that case.

The error message looks like this: `UnitsError: Unit adu of the uncalibrated image do not match unit m of the calibration image`

@mrkite92 -- would this have been clearer in your case?

@crawfordsm -- if you like this approach, let me know and I'll do it for subtract dark also.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ Other Changes and Additions
 
 - Auto-identify files with extension ``fts`` as FITS files. [#355, #364]
 
+- Raise more explicit exception if unit of uncalibrated image and master do
+  not match in ``subtract_bias`` or ``subtract_dark``. [#361, #366]
+
 Bug Fixes
 ^^^^^^^^^
 
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -13,6 +13,7 @@ from astropy.modeling import fitting
 from astropy import stats
 from astropy.nddata import StdDevUncertainty
 from astropy.wcs.utils import proj_plane_pixel_area
+import astropy  # To get the version.
 
 from scipy import ndimage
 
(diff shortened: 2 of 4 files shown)


mrkite92:
Hi Matt,

> This pull request address #361 <https://github.com/astropy/ccdproc/issues/361> in two ways:
> 
> The documentation for ccd_process now clearly states what the expected units of the calibration images are.
> I've implement, for subtract_bias, a more explicit check for a units mismatch and raise a UnitsError instead of a ValueError in that case.
> The error message looks like this: UnitsError: Unit adu of the uncalibrated image do not match unit m of the calibration image
> 
> @mrkite92 <https://github.com/mrkite92> -- would this have been clearer in your case?
> 
Fine to me, thanks for the fix!

Cheers,
JB

crawfordsm:
@mwcraig Seems reasonable -- I've actually run into this problem myself, so it seems like something good to fix.

mwcraig:
@crawfordsm -- I think this is ready to go now (assuming tests pass). I raise a more helpful error in `subtract_bias` and `subtract_dark` if the units don't match. 

Oh, this does need a changelog entry. I'll add that once tests pass.

crawfordsm:
Can you add a test with the 'operand units' just in case that changes at some point in the future?

mwcraig:
@crawfordsm -- done. The check for the specific wording applies only to astropy 1.0.x, I believe. In other versions it should be the case that the class of the exception is `UnitConversionError`. I test for both cases in the bit I added.

PR #375, opened 2016-07-19
Combine 3d

### LINKED ISSUE #340: Combiner class dosen't work with >2D data.
It seems to be because of this: https://github.com/astropy/ccdproc/blob/master/ccdproc%2Fcombiner.py#L97

crawfordsm:
Updated the combine class so that it can use multi-dimensions

closes #340


CHANGES.rst
```diff
@@ -13,6 +13,9 @@ Other Changes and Additions
 - Raise more explicit exception if unit of uncalibrated image and master do
   not match in ``subtract_bias`` or ``subtract_dark``. [#361, #366]
 
+- Updated the ``Combiner`` class so that it could process images with >2 
+  dimensions.
```
mwcraig: Could you please add the PR number here?

MSeifert04:
Unfortunatly I suspect there are two things that might not work with stacked 3d ccds:

- scaling.setter (broadcasts to 3D for correct broadcasting)
- clip.extrema (assumes the stacked images to be 3d)

Could you add these to the test, just to verify if it fails/works?

crawfordsm:
Okay I've updated scaling but I could quickly figure out a way to update clip_extrema so for right now I've added an exception if clip_extrema is used for >2d arrays.  However, if you have some idea how to change that, let me know.

MSeifert04:
Could you try:

```
argsorted = np.argsort(self.data_arr.data, axis=0)
# Create a grid for each dimension except the first one
mg = np.mgrid[[slice(ndim) for i, ndim in enumerate(self.data_arr.shape) if i > 0]]
for i in range(-1*nhigh, nlow):
    # create a tuple with the indices
    where = tuple([argsorted[i,:,:].ravel()] + [i.ravel() for i in mg])
    self.data_arr.mask[where] = True
```

crawfordsm:
Ah thanks @MSeifert04 that seemed to have done the trick

mwcraig:
Thanks, merging. ping @Cadair since he opened the issue originally. New release will be out within a couple days.

Cadair:
Sweet thanks!

Issue #372, opened 2016-07-08
Doctest fails on cddata.rst


olebole:
With the current astropy, I get the following exception when executing the doctest:
```
______________________ [doctest] docs/ccdproc/ccddata.rst ______________________
183 appropriately. Note that the metadata of the result is *not* set if the operation 
184 is between two `~ccdproc.ccddata.CCDData` objects.
185 
186     >>> result = ccd.multiply(0.2 * u.adu)
187     >>> uncertainty_ratio = result.uncertainty.array[0, 0]/ccd.uncertainty.array[0, 0]
188     >>> round(uncertainty_ratio, 5)   # doctest: +FLOAT_CMP
189     0.2
190     >>> result.unit
191     Unit("adu electron")
192     >>> result.header
Expected:
    OrderedDict()
Got:
    {}

/tmp/ccdproc-test-v5YrZs/docs/ccdproc/ccddata.rst:192: DocTestFailure
```

mwcraig:
Since this type has changed a few times I'm going to show the length of the header instead of showing its `__repr__`.

PR #377, opened 2016-07-22
Fix a couple doctest issues

mwcraig:


MSeifert04:
hm, I remember that this test failed when I investigated the changes for astropy 1.2.

Do you know which combination(s) of astropy/ccdproc made this fail again?

mwcraig:
See #372 for an example. It has changed a couple times but this test should work no matter the type.

MSeifert04:
Sorry, I didn't see that issue.

I have two suggestions:
- include a "Fixes/closes issue" somewhere, so the issue is closed when this PR is merged.
- instead of displaying the default header you could explicitly set the header before. The `CCDData._ccddata_arithmetic` explicitly keeps the header so the example could show that that Header/Meta is copied in this case.

mwcraig:
The issue number is buried down in the commit message (which is really not so useful :))

I'm trying to demonstrate that, by default, arithmetic produces no meta. It might be helpful to add an example, though, for which the header is copied. Any idea if that will work with LTS? I've lost track of which changes in 1.2 were backported here for compatibility with LTS....

MSeifert04:
Arithmetics with anything except other `CCDData` will keep it's header because it doesn't go through the `NDData` methods but is redirected to `CCDData._ccddata_arithmetic`.

So if you want an example where the header is discarded you need to do some `CCDData + CCDData` operation. If you want to demonstrate that the header is kept you should do `CCDData + number/array`. I think I made sure that the defaults in `NDData` remained the same (Header is lost, except you use `handle_meta='first_found'`). 

I just digged through the issue and it's probably convolved/complicated:
The problem with the default header/meta is that before astropy v.1.2.x `NDData.__init__` created an empty OrderedDict when no meta was given. But for 1.2 the init now calls instead the `meta.setter` which implements this logic but it is also overloaded in `CCDData`. Here the complicated bit starts: The 1.0.x setter in `CCDData` produces an empty dict while the master-branch creates an empty OrderedDict in case no meta is given.

Long story short: Given that `len()` works it makes the test more robust. But if the purpose of the test is to show that the header/meta is lost then the example should be changed to ccd+ccd (insert the operator of your choice) or you keep it as ccd+number but explicitly set the header before.

mwcraig:
> Sorry again, I hadn't looked through the commit message 😞

Sorry, should have been clearer -- I think you are correct that it is better to put the reference to the issue in the pull request body. I had forgotten about it in the commit message until you asked.

I'm currently thinking about just removing this piece of the example entirely and opening an issue to rewrite the section of the docs for 1.2.

What do you think?

MSeifert04:
> I'm currently thinking about just removing this piece of the example entirely and opening an issue to rewrite the section of the docs for 1.2.

Maybe just remove the `meta` part. Given that it wasn't set before doing the `multiply` someone reading the docs doesn't immediatly see if it was copied or discarded so it doesn't add any obvious information.

I wouldn't remove the complete example, I think it's the only example not using some wrapper (gain/bias/flat) function so it would leave a gap. 
Besides, ccdproc 1.1 already includes a fix for this doctest error so we would remove the example where it would work correctly (1.1.x docs) but still keep it in the 1.0.x docs where it fails if someone uses astropy >= 1.2 that would be ... weird.

mwcraig:
I was just going to remove the last bit, where I currently have `len(results.header)`.

I thought I did see the error in master still...though I don't remember which version of astropy what was with. Might have been 1.1.x...

MSeifert04:
> I was just going to remove the last bit, where I currently have len(results.header).

:+1: 

> I thought I did see the error in master still...though I don't remember which version of astropy what was with. Might have been 1.1.x...

The LTS Travis test with Astropy: 1.0.10 doesn't report any Errors. I'll do a quick check with astropy 1.1 locally.

mwcraig:
@MSeifert04 @crawfordsm -- I've dropped the numpy 1.9 test for a sort of stupid reason: the reproject conda package for numpy 1.9 is broken and it was easier to drop the test than to build reproject for an old numpy on linux64.

Any objections?

MSeifert04:
@mwcraig I was just going to test if it would work if one would install reproject from pip instead of the openastronomy channel and it seems to work (https://travis-ci.org/MSeifert04/ccdproc/builds/147614974 and unfortunatly also #379).

So if you don't really want to remove numpy 1.9 from the test matrix that would be an easy alternative. 

But that's a never ending story. There will always be a package not build against an old numpy/python combination so I wouldn't mind removing numpy 1.9 altogether.

crawfordsm:
That's fine with me and we can revisit if necessary.

PR #382, opened 2016-07-31
Use local scope variables in the helper function in FileCollection._dict_from_fits_header

MSeifert04:
The helper function has an argument ``n_prev`` but used ``n_previous``. This doesn't cause any bugs because the "not local" variable ``n_previous`` holds the same value. But I've renamed the function argument to ``n_previous`` (and added ``missing_marker``) so Python doesn't need to search in nonlocal scopes.

The third small change here was that the helper function now immediatly converts the key to lower-case so ``key.lower()`` is only called once even if the key wasn't present in the ``tbl_dict``.

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -317,12 +317,13 @@ class ImageFileCollection(object):
         """
         from collections import OrderedDict
 
-        def _add_val_to_dict(key, value, tbl_dict, n_prev):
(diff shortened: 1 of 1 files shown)


MSeifert04:
FYI: The test failure is unrelated (see also https://github.com/astropy/ccdproc/pull/377#issuecomment-235135528)

mwcraig:
Thanks, merging, since the only test that failed was the one that was dropped in #377

PR #393, opened 2016-09-07
replace the trim_image call in combine with direct slicing to avoid logging.

### LINKED ISSUE #390: Combine fits.fz fails with hdu argument
Ccdproc.combine will automatically combine the first hdu with data unless the hdu argument is passed. Trying to combine a list of fpacked fits files (i.e. file.fits.fz), the ccdproc.combine works fine with the default hdu, but if you specify an hdu it fails.

E.g:
files = ['file1.fits.fz','file2.fits.fz']
ccdproc.combine(files)     *works
ccdproc.combine(files,hdu=1)    *fails

The problem seems to be with astropy.io.fits?

Error:

ValueError                                Traceback (most recent call last)
<ipython-input-60-aed23739b557> in <module>()
----> 1 bias_master = ccdproc.combine(bias_list,hdu=1,unit=u.adu)

/Users/Kerry/anaconda2/lib/python2.7/site-packages/ccdproc-1.2.dev916-py2.7.egg/ccdproc/combiner.pyc in combine(img_list, output_file, method, weights, scale, mem_limit, clip_extrema, nlow, nhigh, minmax_clip, minmax_clip_min, minmax_clip_max, sigma_clip, sigma_clip_low_thresh, sigma_clip_high_thresh, sigma_clip_func, sigma_clip_dev_func, **ccdkwargs)
    642 
    643                 # Trim image
--> 644                 ccd_list.append(trim_image(imgccd[x:xend, y:yend]))
    645 
    646             # Create Combiner for tile

/Users/Kerry/anaconda2/lib/python2.7/site-packages/ccdproc-1.2.dev916-py2.7.egg/ccdproc/log_meta.pyc in wrapper(*args, **kwd)
     90 
     91         for k, v in six.iteritems(meta_dict):
---> 92             result._insert_in_metadata_fits_safe(k, v)
     93         return result
     94 

/Users/Kerry/anaconda2/lib/python2.7/site-packages/ccdproc-1.2.dev916-py2.7.egg/ccdproc/ccddata.pyc in _insert_in_metadata_fits_safe(self, key, value)
    551             # Shorten, sort of...
    552             short_name = _short_names[key]
--> 553             self.meta[key] = (short_name, "Shortened name for ccdproc command")
    554             self.meta[short_name] = value
    555         else:

/Users/Kerry/anaconda2/lib/python2.7/site-packages/astropy/io/fits/hdu/compressed.pyc in __setitem__(self, key, value)
    138             return
    139 
--> 140         super(CompImageHeader, self).__setitem__(key, value)
    141 
    142 

/Users/Kerry/anaconda2/lib/python2.7/site-packages/astropy/io/fits/header.pyc in __setitem__(self, key, value)
    193             # If we get an IndexError that should be raised; we don't allow
    194             # assignment to non-existing indices
--> 195             self._update((key, value, comment))
    196 
    197     def __delitem__(self, key):

/Users/Kerry/anaconda2/lib/python2.7/site-packages/astropy/io/fits/hdu/compressed.pyc in _update(self, card)
    247             return
    248 
--> 249         super(CompImageHeader, self)._update(card)
    250 
    251         if keyword in Card._commentary_keywords:

/Users/Kerry/anaconda2/lib/python2.7/site-packages/astropy/io/fits/header.pyc in _update(self, card)
   1676         else:
   1677             # A new keyword! self.append() will handle updating _modified
-> 1678             self.append(card)
   1679 
   1680     def _cardindex(self, key):

/Users/Kerry/anaconda2/lib/python2.7/site-packages/astropy/io/fits/hdu/compressed.pyc in append(self, card, useblanks, bottom, end)
    191 
    192         super(CompImageHeader, self).append(card=card, useblanks=useblanks,
--> 193                                             bottom=bottom, end=end)
    194 
    195         remapped_keyword = self._remap_keyword(card.keyword)

/Users/Kerry/anaconda2/lib/python2.7/site-packages/astropy/io/fits/header.pyc in append(self, card, useblanks, bottom, end)
   1272                 # this will just be 80 (Card.length) but it may be longer for
   1273                 # CONTINUE cards
-> 1274                 self._useblanks(len(str(card)) // Card.length)
   1275 
   1276         self._modified = True

/Users/Kerry/anaconda2/lib/python2.7/site-packages/astropy/io/fits/header.pyc in _useblanks(self, count)
   1835         for _ in range(count):
   1836             if self._cards[-1].is_blank:
-> 1837                 del self[-1]
   1838             else:
   1839                 break

/Users/Kerry/anaconda2/lib/python2.7/site-packages/astropy/io/fits/hdu/compressed.pyc in __delitem__(self, key)
    154 
    155         if isinstance(key, int):
--> 156             keyword, index = self._keyword_from_index(key)
    157         elif isinstance(key, tuple):
    158             keyword, index = key

/Users/Kerry/anaconda2/lib/python2.7/site-packages/astropy/io/fits/header.pyc in _keyword_from_index(self, idx)
   1744 
   1745         keyword = self._cards[idx].keyword
-> 1746         repeat = self._keyword_indices[keyword].index(idx)
   1747         return keyword, repeat
   1748 

ValueError: 68 is not in list

MSeifert04:
Closes #390 

The reason why the exception in #390 occured was because for some reason (haven't investigated it) it wasn't possible to create a HIERARCH card for the compressed image. For normal calls to `ccdproc.core` functions one could disable logging by setting the `add_keywords` parameter but that's currently not possible for `ccdproc.combine`. 

However it's just not necessary to add the logging message for the trimming to Header because the trimming is not really done, it's just there to avoid using too much memory.

So instead of trimming the ccddata object it's just sliced in combine. This fixes the error locally but I haven't been able to create a test case for this.

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -657,7 +657,7 @@ def combine(img_list, output_file=None,
                     imgccd = CCDData.read(image, **ccdkwargs)
 
                 # Trim image
-                ccd_list.append(trim_image(imgccd[x:xend, y:yend]))
+                ccd_list.append(imgccd[x:xend, y:yend])
 
             # Create Combiner for tile
             tile_combiner = Combiner(ccd_list, dtype=dtype)


crawfordsm:
@MSeifert04 I think I'd prefer to keep trim_image but use `add_keyword=False` to it.  Just in case there is some behavior or change in terms of how the slicing is done.   No keywords need to be added and so this can be skipped.

MSeifert04:
Yes, there will be one (but I thought that might be advisable) change in behaviour: The array isn't copied twice.

The `trim_image` function [copies](https://github.com/astropy/ccdproc/blob/master/ccdproc/core.py#L523) the slice **but** it's copied once more directly afterwards in [`Combiner`](https://github.com/astropy/ccdproc/blob/master/ccdproc/combiner.py#L103) otherwise this is in this case equivalent to using slicing directly.

MSeifert04:
@crawfordsm If no `fits_section` is given the `trim_image` function only copies and returns the `CCDData` object. And providing a `fits_section` makes no sense at this point in `combine` because it's only sliced to avoid overflowing the RAM.

So I'm pretty certain this won't change the behaviour in any way.

crawfordsm:
Actually that's a good point and probably optimizes it a bit more.  Okay merging.

PR #394, opened 2016-09-08
removed test if transform_func is a callable in transform_image

### LINKED ISSUE #342: hasattr(..., '__call__') in transform_image
`hasattr` is generally considered not very good style [[1]](http://bugs.python.org/issue7006), [[2]](https://hynek.me/articles/hasattr/) and [[3]](https://news.ycombinator.com/item?id=10893226) (and probably more) and it might have some unwanted-side-effects. It's used in [transform_image](https://github.com/astropy/ccdproc/blob/master/ccdproc/core.py#L750) to find out if the input is a callable.

There are some ways to get the same result in a more appropriate manner:
- `callable(...)` (only Python 2.x and python 3.x > 3.2)
- `isinstance(..., collections.Callable)`

or even just don't check and have the `TypeError` raised as soon as it's tryed to be called but cannot be called.

MSeifert04:
Fixes #342 by removing the check if the ``transform_func`` is a callable. However if the first call to `transform_func` fails with a `TypeError` containing `is not callable` then a more informative exception is raised.

Actually I'm pretty sure I've seen a context manager for exactly this type of `try ... except ...` but I can't find it at the moment. Maybe it was in some external package.

### DIFF
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -712,7 +712,7 @@ def flat_correct(ccd, flat, min_value=None):
     Parameters
     ----------
     ccd : `~ccdproc.CCDData`
-        Data to be flatfield corrected.
+        Data to be transformed.
 
     flat : `~ccdproc.CCDData`
         Flatfield to apply to the data.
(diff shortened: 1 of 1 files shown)


MSeifert04:
I also had a closer look at the docstring of that function and removed some extra brackets in the function body.

MSeifert04:
Can't find it anymore so I think this is ready to be reviewed.

PR #402, opened 2016-10-13
Add thin wrapper for block_reduce and block_replicate functions

### LINKED ISSUE #388: Rebin Behavior
The `rebin` function isn't doing what I expect from a binning process on a CCD.  Here's an example:

```
In [31]: arr1 = CCDData([[1,2,3,4], [11,12,13,14], [21,22,23,24], [31,32,33,34]], unit=u.adu)

In [32]: arr1
Out[32]:
CCDData([[ 1,  2,  3,  4],
         [11, 12, 13, 14],
         [21, 22, 23, 24],
         [31, 32, 33, 34]])

In [33]: rebin(arr1, (2,2))
Out[33]:
CCDData([[ 1,  3],
         [21, 23]])
```

It looks like it is taking only the upper left element of the original pixel grid and using it as the value for the output pixel.  I would expect this to do an average or some other combination on the pixel values to get the value for the output pixel.  i.e. I would expect the output of the above example to be:

```
CCDData([[ 6.5,  8.5],
         [26.5, 28.5]])
```

Am I misunderstanding the role of this function or is this the expected behavior?

MSeifert04:
This puts these `astropy.nddata.utils` functions into the namespace of `ccdproc` and deprecates `rebin`.

I'm not sure about the deprecation but `block_replicate` does almost the same and the name is more descriptive.

Closes #388 (maybe)

I'm not sure if it should have the `log_to_metadata` decorator?


crawfordsm:
One question:  should there be either a warning in the documentation or a warning raised if the original source has a mask or uncertainty set?

MSeifert04:
The `block_replicate` and `block_reduce` functions are wrapped by `support_nddata`. This decorator will print a Warning about all set but unprocessed attribute:

```
>>> from astropy.nddata import block_reduce
>>> block_reduce(CCDData(np.random.random((20, 20)), 
                         unit='adu', 
                         mask=np.random.random((20, 20))>0.5),
                 (2, 2))
WARNING: The following attributes were set on the data object, but will be ignored by the function: mask, unit [astropy.nddata.decorators]
```

crawfordsm:
Thanks @MSeifert04 !

Issue #374, opened 2016-07-19
Allow ImageFileCollections to take a list


crawfordsm:
Instead of giving a directory as the entry to ImageCollections, it should also be able to take a list of images so the user can specify exactly what files they want included.

mirca:
Hi @crawfordsm,

should the files be assumed to be in ``location`` or should it be any list of relative (or absolute) paths?

I was thinking in add a parameter which would be the list of files specified by the user wrt to ``location``, is that the intention?

crawfordsm:
@mirca maybe both?   If location is supplied then an absolute path not needed, but if location is None then it would need an absolute path

crawfordsm:
One part of this has been implement in #403, but different locations still needs to be implemented.

mwcraig:
What if, instead of switching based on image location, we do the Unix-y thing of using an absolute path if the filename starts with `/` but otherwise we use pre-pend location?

mwcraig:
Am leaning towards either a list of paths (allowing both relative and absolute) or dropping...

PR #403, opened 2016-10-14
Allow ImageFileCollections to take a list of file names

mirca:
Implements #374.

I added an optional parameter named ``filenames``  which should contain the names of the FITS files in the directory which should be added to the collection. If ``filenames`` is ``None`` and ``location`` is not, then all FITS files in ``location`` are added to the collection.

Appreciate any feedback on this.

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -34,7 +34,12 @@ class ImageFileCollection(object):
     Parameters
     ----------
     location : str or None, optional
-        path to directory containing FITS files.
+        Path to directory containing FITS files.
+        Default is ``None``.
+
+    filenames: str, list of str, or None, optional
+        List of the names of FITS files which will be added to the collection.
+        The filenames are assumed to be in ``location``.
         Default is ``None``.
 
     keywords : list of str, '*' or None, optional
@@ -57,11 +62,15 @@ class ImageFileCollection(object):
         Raised if keywords are set to a combination of '*' and any other
         value.
     """
-    def __init__(self, location=None, keywords=None, info_file=None):
+    def __init__(self, location=None, filenames=None, keywords=None,
+                 info_file=None):
         self._location = location
         self._files = []
         if location:
-            self._files = self._fits_files_in_directory()
+            if filenames:
+                self._files = filenames
+            else:
+                self._files = self._fits_files_in_directory()
         if self._files == []:
             warnings.warn("no FITS files in the collection.",
                           AstropyUserWarning)
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -309,7 +309,7 @@ class ImageFileCollection(object):
         """
         files = []
         if self._filenames:
-            if isinstance(self._filenames, str):
+            if isinstance(self._filenames, six.string_types):
                 files.append(self._filenames)
             else:
                 files = self._filenames
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -69,7 +69,7 @@ class ImageFileCollection(object):
         self._files = []
         if location:
             if self._filenames:
-                if type(self._filenames) is str:
+                if isinstance(self._filenames, str):
                     self._files.append(self._filenames)
                 else:
                     self._files = self._filenames
@@ -285,7 +285,7 @@ class ImageFileCollection(object):
         keywords = '*' if self._all_keywords else self.keywords
         # Re-load list of files
         if self._filenames:
-            if type(self._filenames) is str:
+            if isinstance(self._filenames, str):
                 self._files.append(self._filenames)
             else:
                 self._files = self._filenames
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -37,11 +37,6 @@ class ImageFileCollection(object):
         Path to directory containing FITS files.
         Default is ``None``.
 
-    filenames: str, list of str, or None, optional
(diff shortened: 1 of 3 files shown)


MSeifert04:
@mirca Thank you for the PR. I have been working on the same issue but that somewhat stalled. So I am very happy you found the time to implement this 👍 

Some problems I have encountered which are not adressed in this PR:

- `.refresh` simply loads all files from the `location` again. (intended or broken by design?)
- it is not possible to load files from different `location`s (that's not necessary, that could be a further enhancement possibility)

I'll add some additional comments inline.

ccdproc/image_collection.py
```diff
@@ -57,11 +62,15 @@ class ImageFileCollection(object):
         Raised if keywords are set to a combination of '*' and any other
         value.
     """
-    def __init__(self, location=None, keywords=None, info_file=None):
+    def __init__(self, location=None, filenames=None, keywords=None,
```
MSeifert04: New arguments should be added at the end of the list of arguments to ensure backwards-compatibility.

ccdproc/image_collection.py
```diff
@@ -34,7 +34,12 @@ class ImageFileCollection(object):
     Parameters
     ----------
     location : str or None, optional
-        path to directory containing FITS files.
+        Path to directory containing FITS files.
+        Default is ``None``.
+
+    filenames: str, list of str, or None, optional
```
MSeifert04: I haven't tested it but is it possible to provide a single `str`?

ccdproc/tests/test_image_collection.py
```diff
@@ -85,6 +85,13 @@ def test_summary_is_summary_info(self, triage_setup):
             location=triage_setup.test_dir, keywords=['imagetyp', 'filter'])
         assert img_collection.summary is img_collection.summary_info
 
+    def test_filenames_are_set_properly(self, triage_setup):
+        fn = ['filter_no_object_bias.fit', 'filter_object_light_foo.fit']
+        img_collection = image_collection.ImageFileCollection(
+            location=triage_setup.test_dir, filenames=fn, keywords=['filter'])
+        assert img_collection.files is fn
```
MSeifert04: `is` tests are prone to errors, wouldn't `==` suffice?

ccdproc/tests/test_image_collection.py
```diff
@@ -85,6 +85,13 @@ def test_summary_is_summary_info(self, triage_setup):
...
+        assert img_collection.summary is img_collection.summary_info
```
MSeifert04: This test is not needed because they are always equal. `summary_info` is just the deprecated property.

ccdproc/image_collection.py
mirca: Thanks! Now it is possible, but ``ImageFileCollection.files`` is kept as list.

ccdproc/image_collection.py
mirca: Great! Addressed.

ccdproc/tests/test_image_collection.py
mirca: Done :)

ccdproc/tests/test_image_collection.py
mirca: Addressed :)

mirca:
@MSeifert04 Thank you very much for your review :) For now, ``location`` only points to one directory, as you mentioned, but I can work to handle that case (better in a different PR?).

ccdproc/image_collection.py
```diff
@@ -51,17 +51,30 @@ class ImageFileCollection(object):
...
-    def __init__(self, location=None, keywords=None, info_file=None):
+    def __init__(self, location=None, keywords=None, info_file=None,
+                 filenames=None):
         self._location = location
+        self._filenames = filenames
         self._files = []
         if location:
-            self._files = self._fits_files_in_directory()
+            if self._filenames:
+                if type(self._filenames) is str:
```
MSeifert04: I think there is some convention that one compares types using `isinstance`: `isinstance(self._filenames, six.string_types)` (not sure if it's really called that 😄 )

ccdproc/image_collection.py
```diff
@@ -271,7 +284,13 @@ def refresh(self):
         """
         keywords = '*' if self._all_keywords else self.keywords
         # Re-load list of files
-        self._files = self._fits_files_in_directory()
+        if self._filenames:
```
MSeifert04: This seems exactly like the block in `__init__`. To ease maintenance could you create a helper-method or check if this block could be put into `_fits_files_in_directory`?

MSeifert04:
@mirca I think having the option to specify files in the location is already a great addition 👍

Let's see what @crawfordsm says, since he proposed the enhancement in #374.

ccdproc/image_collection.py
MSeifert04: This _should_ also fix the Python 2.7 errors (I think it got unicode instead of str there).

ccdproc/image_collection.py
mirca: @MSeifert04 I added a helper method called ``_get_files()``, do you like this name? xD

mirca:
@mwcraig thanks for your review  :)

crawfordsm:
This looks really great and it is a good addition-- thanks @mirca !   

I'm merging this pull request as I think it is useful on its own, but it would be good to still add the ability to add files from different locations so for the time being I'll leave the issue open.

Issue #400, opened 2016-10-13
image file collection CCDData


crawfordsm:
Image file collection currently can return an hdu, data, or header, but should also have a function to return CCDData objects

PR #405, opened 2016-10-17
CCDData generator for ImageCollection

mirca:
Hello everyone,

Is this the right approach to implement #400?

I used ``'adu'`` as the default unit to instantiate an CCDData object in case ``'BUNIT'`` is not present in the ``header``. I'm not sure if that is intended. Should the ccddata unit be given as parameter somehow (for instance, through ``kwd``)?

Thanks!

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -15,6 +15,8 @@ from astropy.extern import six
 import warnings
 from astropy.utils.exceptions import AstropyUserWarning
 
+from .ccddata import fits_ccddata_reader
+
 logger = logging.getLogger(__name__)
 
 __all__ = ['ImageFileCollection']
@@ -680,10 +682,16 @@ class ImageFileCollection(object):
 
             file_name = path.basename(full_path)
 
+            if 'BUNIT' in hdulist[0].header:
+                ccddata_unit = hdulist[0].header['BUNIT']
+            else:
+                ccddata_unit = 'adu'
+
             return_options = {'header': hdulist[0].header,
                               'hdu': hdulist[0],
-                              'data': hdulist[0].data}
-
+                              'data': hdulist[0].data,
+                              'ccddata': fits_ccddata_reader(full_path,
+                                                             unit=ccddata_unit)}
             try:
                 yield (return_options[return_type]  # pragma: no branch
                        if (not return_fname) else
@@ -748,3 +756,9 @@ class ImageFileCollection(object):
     data.__doc__ = _generator.__doc__.format(name='image',
                                              default_scaling='False',
                                              return_type='numpy.ndarray')
+
+    def ccddata(self, **kwd):
+        return self._generator('ccddata', **kwd)
+    ccddata.__doc__ = _generator.__doc__.format(name='CCDData',
+                                                default_scaling='True',
+                                                return_type='ccdproc.CCDData')
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -651,10 +651,10 @@ class ImageFileCollection(object):
             Default is ``False``.
 
         ccd_kwargs : dict, optional
-            Dict with parameters for `~ccdproc.ImageFileCollection.ccddata`.
+            Dict with parameters for `~ccdproc.fits_ccddata_reader`.
             For instance, the key ``'unit'`` can be used to specify the unit
-            of the data. If ``'unit'`` is not given and if it cannot be
-            taken in the header, then ``'adu'`` is used as the default unit.
+            of the data. If ``'unit'`` is not given then ``'adu'`` is used as
+            the default unit.
             See `~ccdproc.fits_ccddata_reader` for a complete list of
             parameters that can be passed through ``ccd_kwargs``.
 
@@ -692,16 +692,13 @@ class ImageFileCollection(object):
             file_name = path.basename(full_path)
 
             if 'unit' not in ccd_kwargs:
-                if 'BUNIT' in hdulist[0].header:
-                    ccd_kwargs['unit'] = hdulist[0].header['BUNIT']
-                else:
-                    ccd_kwargs['unit'] = 'adu'
+                ccd_kwargs['unit'] = 'adu'
 
             return_options = {'header': hdulist[0].header,
                               'hdu': hdulist[0],
                               'data': hdulist[0].data,
-                              'ccddata': fits_ccddata_reader(full_path,
-                                                             **ccd_kwargs)}
+                              'ccd': fits_ccddata_reader(full_path,
+                                                         **ccd_kwargs)}
             try:
                 yield (return_options[return_type]  # pragma: no branch
                        if (not return_fname) else
@@ -767,8 +764,8 @@ class ImageFileCollection(object):
                                              default_scaling='False',
                                              return_type='numpy.ndarray')
 
-    def ccddata(self, ccd_kwargs={}, **kwd):
-        return self._generator('ccddata', ccd_kwargs=ccd_kwargs, **kwd)
-    ccddata.__doc__ = _generator.__doc__.format(name='CCDData',
-                                                default_scaling='True',
-                                                return_type='ccdproc.CCDData')
+    def ccds(self, ccd_kwargs={}, **kwd):
+        return self._generator('ccd', ccd_kwargs=ccd_kwargs, **kwd)
+    ccds.__doc__ = _generator.__doc__.format(name='CCDData',
+                                             default_scaling='True',
+                                             return_type='ccdproc.CCDData')
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -603,6 +603,7 @@ class ImageFileCollection(object):
                    overwrite=False,
                    do_not_scale_image_data=True,
                    return_fname=False,
+                   ccd_kwargs={},
                    **kwd):
         """
         Generator that yields each {name} in the collection.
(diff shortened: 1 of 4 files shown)


crawfordsm:
I'll let @mwcraig comment on the implementation format but it looks like a good start.   However, I think the user should have the option of specifying the default unit through a kwd. Thanks for taking this on!

MSeifert04:
I agree with the option to pass arguments to the reader. Maybe just add another argument like `kwargs_open` (the name is horrible) that expects a dictionary and forwards it to the reader?

MSeifert04:
@mirca I meant an option/parameter to pass any amount of parameters to the reader, not just the unit. That also removes duplicate code (unit processing is also done in the reader).

mirca:
@MSeifert04 Got it! :) I don't have any strong feelings about the name, but maybe it could be ``ccd_kwargs``?

ccdproc/image_collection.py
```diff
@@ -652,10 +663,17 @@ def _generator(self, return_type,
 
             file_name = path.basename(full_path)
 
+            if 'unit' not in ccd_kwargs:
```
MSeifert04: This shouldn't be needed, there is already some [code inside the reader](https://github.com/astropy/ccdproc/blob/master/ccdproc/ccddata.py#L643) to evaluate this. It's easier to maintain if the behaviour does not differ and is only implemented in one place.

ccdproc/tests/test_image_collection.py
```diff
@@ -217,6 +217,17 @@ def test_generator_data(self, triage_setup):
         for img in collection.data():
             assert isinstance(img, np.ndarray)
 
+    def test_generator_ccddata(self, triage_setup):
+        from ccdproc import CCDData
+        collection = image_collection.ImageFileCollection(location=triage_setup.test_dir,
+                                             keywords=['imagetyp'])
```
MSeifert04: the indentation seems wrong. If lines get too long you can also break before the first argument:

```
        collection = image_collection.ImageFileCollection(
            location=triage_setup.test_dir, keywords=['imagetyp'])
```

ccdproc/image_collection.py
```diff
@@ -720,3 +738,9 @@ def data(self, do_not_scale_image_data=False, **kwd):
     data.__doc__ = _generator.__doc__.format(name='image',
                                              default_scaling='False',
                                              return_type='numpy.ndarray')
+
+    def ccddata(self, ccd_kwargs={}, **kwd):
```
MSeifert04: I know it is already inconsistent but there is `headers` and `hdus` and I think the plural does make sense here, maybe `ccds` as name?

ccdproc/tests/test_image_collection.py
```diff
@@ -217,6 +217,17 @@ def test_generator_data(self, triage_setup):
         for img in collection.data():
             assert isinstance(img, np.ndarray)
 
+    def test_generator_ccddata(self, triage_setup):
+        from ccdproc import CCDData
```
MSeifert04: I think you can import this at the top of the file. It's likely someone does need in at some point for another test.

ccdproc/tests/test_image_collection.py
```diff
@@ -217,6 +217,17 @@ def test_generator_data(self, triage_setup):
...
+        collection = image_collection.ImageFileCollection(location=triage_setup.test_dir,
+                                             keywords=['imagetyp'])
+        for img in collection.ccddata():
+            assert isinstance(img, CCDData)
+
+        ccd_kwargs = {'unit': 'adu'}
+        for img in collection.ccddata(ccd_kwargs=ccd_kwargs):
```
MSeifert04: That's a really nitpicky comment but generally `CCDData` objects are labelled `ccd`, not `img`. 

Feel free to ignore this :smile:

ccdproc/image_collection.py
```diff
@@ -619,6 +622,14 @@ def _generator(self, return_type,
             not the full path to the file.
             Default is ``False``.
 
+        ccd_kwargs : dict, optional
+            Dict with parameters for `~ccdproc.ImageFileCollection.ccddata`.
```
MSeifert04: I think you meant `~ccdproc.fits_ccddata_reader` here, right?

MSeifert04:
Optional: Could you include the name here in ther narrative docs: https://github.com/astropy/ccdproc/blob/master/docs/ccdproc/image_management.rst#iterating-over-hdus-headers-or-data?

ccdproc/image_collection.py
mirca: I agree, thanks for pointing that out. However, it seems that I still need to set a default unit, otherwise the reader is going to throw this error: ``ValueError: a unit for CCDData must be specified.``

ccdproc/tests/test_image_collection.py
mirca: I just followed the indentation in previous tests. Should I keep it as is?

ccdproc/tests/test_image_collection.py
mirca: Hahaha, makes sense xD

ccdproc/image_collection.py
mirca: If unit is ``None`` then ``_generator`` is always going to throw ``ValueError: a unit for CCDData must be specified.`` even if I just would like to use ``headers`` or ``data``, for instance.

mirca: Tests are passing using

```
if 'unit' not in ccd_kwargs:
                ccd_kwargs['unit'] = 'adu'
```
However, after removing this if block code ``TestImageFileCollection.test_headers`` will fail because 
`fits_ccddata_reader` in
```
return_options = {'header': hdulist[0].header,
                              'hdu': hdulist[0],
                              'data': hdulist[0].data,
                              'ccd': fits_ccddata_reader(full_path,
                                                         **ccd_kwargs)}
```
will throw ``ValueError: a unit for CCDData must be specified``.

MSeifert04: I don't see how `test_headers` can access the `fits_ccddata_reader`. I will have a look.

ccdproc/image_collection.py
```diff
@@ -680,10 +691,14 @@ def _generator(self, return_type,
...
 
+            if 'unit' not in ccd_kwargs:
+                ccd_kwargs['unit'] = 'adu'
+
             return_options = {'header': hdulist[0].header,
                               'hdu': hdulist[0],
-                              'data': hdulist[0].data}
-
+                              'data': hdulist[0].data,
+                              'ccd': fits_ccddata_reader(file_name,
```
mirca: @MSeifert04 thank you very much! haha =D

ccdproc/tests/test_image_collection.py
```diff
@@ -233,12 +233,16 @@ def test_generator_data(self, triage_setup):
...
 
-    def test_generator_ccds(self, triage_setup):
+    def test_generator_ccds_without_unit(self, triage_setup):
         collection = image_collection.ImageFileCollection(
                 location=triage_setup.test_dir, keywords=['imagetyp'])
-        for ccd in collection.ccds():
-            assert isinstance(ccd, CCDData)
 
+        with pytest.raises(ValueError):
+            ccd = collection.ccds()[0]
```
MSeifert04: Even if this test works, it doesn't test what you want because one cannot subscript generators. You need to call `next(collections.ccds())`

ccdproc/image_collection.py
```diff
@@ -691,14 +691,11 @@ def _generator(self, return_type,
...
-
-            return_options = {'header': hdulist[0].header,
-                              'hdu': hdulist[0],
-                              'data': hdulist[0].data,
-                              'ccd': fits_ccddata_reader(file_name,
-                                                         **ccd_kwargs)}
+            return_options = {'header': lambda: hdulist[0].header,
+                              'hdu': lambda: hdulist[0],
+                              'data': lambda: hdulist[0].data,
+                              'ccd': lambda: fits_ccddata_reader(file_name,
```
MSeifert04: Since you change all these lines, you can also change the indentation here to avoid the last line break:

```
            return_options = {
                'header': lambda: hdulist[0].header,
                'hdu': lambda: hdulist[0],
                'data': lambda: hdulist[0].data,
                'ccd': lambda: fits_ccddata_reader(full_path, **ccd_kwargs)
                }
```

MSeifert04:
@mirca I feel rather bad for commenting so much on your PR. 

I just wanted to add: This is a great PR! Thank you.

mirca:
@MSeifert04 Hahaha, don't worry, I'm rather keen on the styles too xD

mirca:
@MSeifert04 Some tests are failing after the new ``return_options``, I'm looking into them.

MSeifert04:
@mirca You need to call the `return_options[option]` result. See also the second part of https://github.com/astropy/ccdproc/pull/405#discussion_r84078776.

mirca:
@MSeifert04 oh sorry, got that now xD

ccdproc/image_collection.py
```diff
@@ -601,6 +603,7 @@ def _generator(self, return_type,
                    overwrite=False,
                    do_not_scale_image_data=True,
                    return_fname=False,
+                   ccd_kwargs={},
```
mwcraig: Please change to `ccd_kwargs={}` to `ccd_kwargs=None` then, early in the function, add a line like 

```python
ccd_kwargs = ccd_kwargs or {}
```
or
```
if ccd_kwargs is None:
    ccd_kwargs = {}
``` 

Though the current implementation does not have any issues (because you are not modifying the dictionary between calls), in principle

ccdproc/image_collection.py
```diff
@@ -748,3 +761,9 @@ def data(self, do_not_scale_image_data=False, **kwd):
     data.__doc__ = _generator.__doc__.format(name='image',
                                              default_scaling='False',
                                              return_type='numpy.ndarray')
+
+    def ccds(self, ccd_kwargs={}, **kwd):
```
mwcraig: Same here as above.

mwcraig:
Thanks, @mirca -- as soon as tests finish I'll merge this!

mwcraig:
@mirca -- would you mind making one more PR to add yourself to `AUTHORS.rst` 😀

PR #408, opened 2016-10-21
Adding licence to the docs

bsipocz:
I think somewhere in the docs we should mention the license

### DIFF
--- a/docs/license.rst
+++ b/docs/license.rst
@@ -1,0 +1,12 @@
+.. _license:
+
+********
+Licenses
+********
+
+Ccdproc License
+===============
+
+Ccdproc is licensed under a 3-clause BSD style license:
+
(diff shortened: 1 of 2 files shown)


MSeifert04:
@bsipocz Thank you for this PR! 

I've left one comment. One additional question because I'm not very familiar with `Manifest.in` : will the License.rst be included in the tarball? If not could you also add it there?

docs/license.rst
```diff
@@ -0,0 +1,10 @@
+********
```
bsipocz: Sure. I've copied it from astropy, didn't look at the ccdproc conventions.

bsipocz:
@MSeifert04 - I believe everything under docs gets included in the tarball.

PR #411, opened 2016-10-25
Removed 1.1.1 section of the changelog.

MSeifert04:
There won't be a 1.1.1 release (I think) so I've moved those entries to 1.2.0

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ ccdproc has now the following additional dependency:
 New Features
 ^^^^^^^^^^^^
 
+- Add an optional attribute named ``filenames`` to ``ImageFileCollection``,
+  so that users can pass a list of FITS files to the collection. [#374]
+
 - Added ``block_replicate``, ``block_reduce`` and ``block_average`` functions.
   [#402]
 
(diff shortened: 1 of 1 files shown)


mwcraig:
Could you please switch `scikit-image` back to a required dependency? it seems like it should be easy for people to install (or at least `scipy` and `numpy` are much harder to compile from source, and anaconda offers packages for scikit-image in its defaults channel).

mwcraig:
As an alternative, add: `setup_requires=['scikit-image']`

MSeifert04:
@mwcraig I've rebased the PR but skipped the CI (conflict was just in Changes.rst).

mwcraig:
Having looked into `extra_requires` a bit more I'm leaning back towards just making `scikit-image` a hard requirement. The rationale is that once we add an `extras_require`, the advice from the [setuptools documentation](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) is to keep the extra requirement around forever (well, to be more specific, `extra_requires` is a dictionary, and the *key* should stay around forever).

Also, the [semantic version docs](http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api) seem to indicate that adding a requirement does not count as a backwards-compatibility-breaking change (assuming the accompanying API change is backwards compatible). 

If this were a patch release it might be different, but for a minor version bump adding a requirement seems fine. If we don't, then we need to update the documentation to indicate that a couple of functions require `scikit-image` and/or change the recommended install command to `pip install ccdproc[all]`.

bsipocz:
(If ``scikit-image`` is an optional dependency, not only the documentation should acknowledge it, but  tests, too. If it's not a hard dependency then tests should always respect that and pass with or without it, and that may turn out to be an unnecessary maintenance overhead.)

MSeifert04:
> Also, the semantic version docs seem to indicate that adding a requirement does not count as a backwards-compatibility-breaking change (assuming the accompanying API change is backwards compatible). 

In that case I think it would be best to put it in the `install_requires` section again. 

@mwcraig Do you want to change it? Also the `_optional_` should be removed from the `Changes.rst` as well.

MSeifert04:
I've removed the removal of `scikit-image` in `install_requires` and removed the word `optional` in the changelog again.

mwcraig:
Thanks, merging!

PR #414, opened 2016-11-01
Add `ccdmask` function to replicate IRAF's ccdmask task

joshwalawender:

For documentation changes:

For bugfixes:

For new functionality:

I've added a `ccdmask` function to replicate [IRAF's ccdmask](http://stsdas.stsci.edu/cgi-bin/gethelp.cgi?ccdmask) task.

I'm not sure this is the optimal way to identify bad pixels, but I needed to match an old IRAF implementation, so I've written this in an attempt to match the description in the IRAF help text for that task.


crawfordsm:
Hey @joshwalawender thanks for the contribution and also testing out our new template! I'm sure we will have a couple of comments, but this looks like a useful submission

ccdproc/core.py
```diff
@@ -1558,6 +1558,137 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+def ccdmask(ratio, findbadcolumns=False, ncmed=7, nlmed=7, ncsig=15, nlsig=15,
+            lsigma=9, hsigma=9, ngood=5):
+    '''
+    Uses method based on the IRAF ccdmask task to generate a mask based on the
+    given input.
+
+    Parameters
+    ----------
+    ratio : `~ccdproc.CCDData`
+        Data to used to form mask.  Typically this is a ratio of two flat field
```
MSeifert04: I think `a ratio` should be `the ratio`?

ccdproc/core.py
```diff
@@ -1558,6 +1558,137 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+        images.
+
+    ncmed, nlmed: `int`, optional
+        The column and line size of a moving median rectangle used to estimate
```
MSeifert04: Again I think it should be `the moving median rectangle` instead of `a ...`

ccdproc/core.py
```diff
@@ -1558,6 +1558,137 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+    nl, nc = ratio.data.shape
+    nbl = int(np.ceil(ratio.data.shape[0] / nlsig))
+    nbc = int(np.ceil(ratio.data.shape[1] / ncsig))
+    for i in range(nbl):
+        for j in range(nbc):
+            l1 = i*nlsig
+            l2 = (i+1)*nlsig
+            if l2 > nl:
+                l2 = nl
+            c1 = j*ncsig
```
MSeifert04: I don't mind the longish variant but I generally combine similar statements:

```
l1, c1 = i*nlsig, j*ncsig
l2, c2 = (i+1)*nlsig, (j+1)*ncsig
...
```
That makes it clearer (at least to me) which statements belong together. This is however a matter of individual style, so use whichever you like more.

ccdproc/core.py
```diff
@@ -1558,6 +1558,137 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+            c2 = (j+1)*ncsig
+            if c2 > nc:
+                c2 = nc
+            block = medsub[l1:l2,c1:c2]
+            high = np.percentile(block.ravel(), 69.1)
```
MSeifert04: scikit-image also has `percentile`-filter. Maybe this whole loop could be done with one of those? All those `ncmed, nlmed, nlsig, ncsig` parameters make it somewhat complicated to understand the code. 😄

ccdproc/core.py
```diff
@@ -1558,6 +1558,137 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+            low = np.percentile(block.ravel(), 30.9)
+            block_sigma = (high-low)/2.0
+            block_mask = ( (block > hsigma*block_sigma) |
+                           (block < -lsigma*block_sigma) )
+            mblock = np.ma.MaskedArray(block, mask=block_mask)
+
+            if findbadcolumns:
+                csum = np.ma.sum(mblock, axis=0)
+                csum_sigma = np.ma.MaskedArray([np.sqrt(c2-c1-x)*block_sigma
```
MSeifert04: You don't need this unrolling of `np.ma.sum`:

```
x = np.ma.sum(mblock.mask, axis=0)
x[x<=0] = 0
csum_sigma = MaskedArray(np.sqrt(c2-c1-x))
```
or if you don't like that the `x` is a scalar so `math.sqrt(c2-c1-x)` should be orders of magnitudes faster.

ccdproc/core.py
```diff
@@ -1558,6 +1558,137 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+    than that given by the ngood parameter all the pixels in the segment are
+    also marked as bad.
+
+
+    Returns
+    -------
+    mask : `numpy.ndarray`
+        A boolean ndarray with the bad pixel positions identified.
+    '''
+    mask = ~np.isfinite(ratio.data)
```
MSeifert04: I think this function needs a check to make sure the input is 2d. I think 1d or 3d could really mess this one up.

ccdproc/core.py
```diff
@@ -1558,6 +1558,137 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+                                   for x in np.ma.sum(mblock.mask, axis=0)])
+                colmask = ( (csum.filled(1) > hsigma*csum_sigma) |
+                            (csum.filled(1)< -lsigma*csum_sigma) )
+                for c in range(c2-c1):
+                    block_mask[:,c] = block_mask[:,c] | np.array([colmask[c]]*(l2-l1))
+
+            mask[l1:l2,c1:c2] = block_mask
+
+    if findbadcolumns:
+        for c in range(0,nc):
```
MSeifert04: 3 nested `for` loops is going to be really slow in Python. Not a dealbreaker but something that should/could be noted as inline comment for future improvements.

ccdproc/core.py
```diff
@@ -1558,6 +1558,137 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+            for l in range(0,nl-ngood-1):
+                if mask[l,c] == True:
+                    for i in range(2,ngood+2):
+                        if (mask[l+i,c] == True) and not np.all(mask[l:l+i+1,c]):
```
MSeifert04: `l+i` is computed several times here without knowing what it indicates. Could you do the computation of it as seperate statement (performance) and give it a meaningful variable name?

ccdproc/core.py
```diff
@@ -1558,6 +1558,137 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+                           (block < -lsigma*block_sigma) )
+            mblock = np.ma.MaskedArray(block, mask=block_mask)
+
+            if findbadcolumns:
+                csum = np.ma.sum(mblock, axis=0)
+                csum_sigma = np.ma.MaskedArray([np.sqrt(c2-c1-x)*block_sigma
+                                   if c2-c1-x > 0 else 0
+                                   for x in np.ma.sum(mblock.mask, axis=0)])
+                colmask = ( (csum.filled(1) > hsigma*csum_sigma) |
+                            (csum.filled(1)< -lsigma*csum_sigma) )
```
MSeifert04: could you insert a whitespace before `<`?

ccdproc/core.py
joshwalawender: If I understand the `percentile_filter` function correctly, then this is a different behavior than the IRAF implementation.  In IRAF, the input image is grouped in to blocks of pixels and the high and low sigmas are calculated for the block.  In the `percentile_filter` the same high and low sigmas can be calculated for a moving window over the image.  The result is that each pixel has a uniquely calculated high and low sigma value, where in the block version every pixel in the block has the same high and low sigma value.

My instant says that the `percentile_filter` behavior is preferable and it will probably look a bit cleaner since the grouping in to blocks in cumbersome in the code.  We could preserve both behaviors and trigger to IRAF-like block calculation with a keyword.  What does everyone think?

CHANGES.rst
```diff
@@ -13,6 +13,8 @@ New Features
 - Added ``block_replicate``, ``block_reduce`` and ``block_average`` functions.
   [#402]
 
+- Added ``ccdmask`` function.
```
MSeifert04: Could you include the PR number? Just append `[#414]`.

ccdproc/core.py
```diff
@@ -1558,6 +1558,134 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+    nbl = int(np.ceil(ratio.data.shape[0] / nlsig))
+    nbc = int(np.ceil(ratio.data.shape[1] / ncsig))
+    for i in range(nbl):
+        for j in range(nbc):
+            l1 = i*nlsig
+            l2 = min((i+1)*nlsig, nl)
+            c1 = j*ncsig
+            c2 = min((j+1)*ncsig, nc)
+            block = medsub[l1:l2,c1:c2]
+            high = np.percentile(block.ravel(), 69.1)
```
MSeifert04: I think this needs some benchmarking for small and big blocks on approximatly real-sized images. If it's really slow (in the order of seconds) we should probably utilize the `percentile_filter` and/or rewrite that part in Cython (or numba, but that would mean an additional dependency - but I like numba and do have a lot of crazy functions that I could include then! 😄)

If you don't have time to benchmark I can also do it (probably not before next week though).

ccdproc/core.py
joshwalawender: This didn't seem to have an effect on the `%timeit` results, but it looks cleaner, so I've made this change (will push edits later).

ccdproc/core.py
joshwalawender: So I tried some benchmarking today.  I added a code path (not yet pushed to github) which optionally uses the `percentile_filter` rather than dividing the image up in to blocks. This makes the code **much** simpler to read (4 lines compared to 28 lines), but it is half as fast.  Both routes are slow.  With a 4096 x 4096 ratio image input, here are the `%timeit` results:

```
In [20]: %timeit ccdproc.ccdmask(ratio, byblocks=False)
1 loop, best of 3: 2min 21s per loop

In [21]: %timeit ccdproc.ccdmask(ratio, byblocks=True)
1 loop, best of 3: 1min per loop
```

Not sure why doing this block by block with nested loops is faster.  Clearly this isn't a very fast operation.

MSeifert04: What block-sizes did you have for the sigma-calculation? 

I ask because the byblocks variant scales with t ~ 1/A (A = block_x * block_y) while the filter variant scales with t ~ A.

ccdproc/tests/test_ccdmask.py
```diff
@@ -0,0 +1,192 @@
...
+
+from numpy.testing import assert_array_equal
+
+import numpy as np
+
+from ..core import ccdmask
+from ..ccddata import CCDData
+
+def test_ccdmask_pixels():
+    flat1 = CCDData(data=np.array([[
```
MSeifert04: This fails because it's not possible to provide the `data` as keyword argument to the `NDDataArray` constructor. Just remove the `data=` and just pass in the array.

ccdproc/tests/test_ccdmask.py
```diff
@@ -0,0 +1,192 @@
...
+        20018, 19935, 20037, 19946, 19998],
+       [20001, 19940, 19994, 19835, 19959, 19895, 20017, 20002, 20007,
+        19851, 19900, 20044, 20354, 19814, 19869, 20148, 20001, 20143,
+        19778, 20146, 19975, 19859, 20008, 20041, 19937, 20072, 20203,
+        19778, 20027, 20075, 19877, 19999],
+       [19753, 19866, 20037, 20149, 20020, 20071, 19955, 20164, 19837,
+        19967, 19959, 20163, 20003, 20127, 20065, 20118, 20104, 19839,
+        20124, 20057, 19943, 20023, 20138, 19996, 19910, 20048, 20070,
+        19833, 19913, 20012, 19897, 19983]]), unit='adu')
+    flat2 = CCDData(data=np.array([[
```
MSeifert04: same here.

ccdproc/core.py
MSeifert04: That's really slow, do you happen to know how fast/slow irafs implementation is?

`astroscrappy` already uses a parallel median filter (just for float32's). Maybe one could leverage it with some template-engine for other dtypes.

joshwalawender: I just ran a quick test and IRAF's version, operating on the same data with the same input parameters takes about the same amount of time.  IRAF took 1m 9s, python took 1m 1s.  It's not quite apples to apples as the IRAF implementation executes the `findbadcolumns` path which was being skipped in the python execution (adding that pushes python to over 2 minutes, but that path isn't optimized).

ccdproc/core.py
```diff
@@ -1558,6 +1558,144 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+        A boolean ndarray with the bad pixel positions identified.
+    '''
+    if ratio.data.ndim !=2:
+        return None
+    mask = ~np.isfinite(ratio.data)
+    medsub = ratio.data - ndimage.filters.median_filter(ratio.data,
+                                                        size=(nlsig, ncsig))
+
+    nlines, ncols = ratio.data.shape
+    if byblocks is True:
```
MSeifert04: you can omit the `is True` here.

ccdproc/core.py
```diff
@@ -1558,6 +1558,144 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+        for i in range(nlinesblock):
+            for j in range(ncolsblock):
+                l1 = i*nlsig
+                l2 = min((i+1)*nlsig, nlines)
+                c1 = j*ncsig
+                c2 = min((j+1)*ncsig, ncols)
+                block = medsub[l1:l2,c1:c2]
+                high = np.percentile(block.ravel(), 69.1)
+                low = np.percentile(block.ravel(), 30.9)
+                block_sigma = (high-low)/2.0
```
MSeifert04: Why not `0.5 * (high - low)`?

ccdproc/core.py
```diff
@@ -1558,6 +1558,144 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+        high = ndimage.percentile_filter(medsub, 69.1, size=(nlsig, ncsig))
+        low = ndimage.percentile_filter(medsub, 30.9, size=(nlsig, ncsig))
+        sigmas = (high-low)/2.0
+        mask = ( (medsub > hsigma*sigmas) | (medsub < -lsigma*sigmas) )
+
+    if findbadcolumns:
+        # Loop through columns and look for short segments (<ngood pixels long)
+        # which are unmasked, but are surrounded by masked pixels and mask them
+        # under the assumption that the column region is bad.
+        for col in range(0,ncols):
```
MSeifert04: could you use six.moves.range here and in the following lines to avoid python2 range?

ccdproc/core.py
MSeifert04: not really, just wanted to know 😄 

I always use multiplication in those cases so I'll never run into py2 floor-division.

ccdproc/core.py
```diff
@@ -1558,6 +1558,144 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+                l2 = min((i+1)*nlsig, nlines)
+                c1 = j*ncsig
+                c2 = min((j+1)*ncsig, ncols)
+                block = medsub[l1:l2,c1:c2]
+                high = np.percentile(block.ravel(), 69.1)
+                low = np.percentile(block.ravel(), 30.9)
+                block_sigma = (high-low)/2.0
+                block_mask = ( (block > hsigma*block_sigma) |
+                               (block < -lsigma*block_sigma) )
+                mblock = np.ma.MaskedArray(block, mask=block_mask)
```
MSeifert04: Does it work if you use `np.ma.MaskedArray(block, mask=block_mask, copy=False)` here? To avoid some copy-operations.

ccdproc/core.py
```diff
@@ -1558,6 +1558,144 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+        # which are unmasked, but are surrounded by masked pixels and mask them
+        # under the assumption that the column region is bad.
+        for col in six.moves.range(0,ncols):
+            for line in six.moves.range(0,nlines-ngood-1):
+                if mask[line,col] == True:
+                    for i in six.moves.range(2,ngood+2):
+                        lend = line+i
+                        if (mask[lend,col] == True) and not np.all(mask[line:lend+1,col]):
+                            mask[line:lend,col] = True
+    nmasked = np.sum(np.array(mask.ravel(), dtype=np.int))
```
MSeifert04: What is this variable for?

joshwalawender: `nmasked` is vestigial, I'll take it out.

ccdproc/core.py
```diff
@@ -1558,6 +1559,142 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+
+    As a final step each column is examined to see if there are small segments
+    of unflagged pixels between bad pixels. If the length of a segment is less
+    than that given by the ngood parameter all the pixels in the segment are
+    also marked as bad.
+
+    Returns
+    -------
+    mask : `numpy.ndarray`
+        A boolean ndarray where the bad pixels hve a value of 1 (True) and valid
```
MSeifert04: minor typo here: `hve`

MSeifert04:
Sorry for this flood of comments, I think for the most parts this is ready now. 
But it wouldn't hurt if someone else took a look at this as well.

Thanks a lot @joshwalawender !

joshwalawender:
Thanks @MSeifert04 the comments were helpful and you work fast!

ccdproc/core.py
```diff
@@ -1558,6 +1559,142 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+    greater to detect only true bad pixels and not the extremes of the noise
+    distribution.
+
+    As a final step each column is examined to see if there are small segments
+    of unflagged pixels between bad pixels. If the length of a segment is less
+    than that given by the ngood parameter all the pixels in the segment are
+    also marked as bad.
+
+    Returns
+    -------
```
crawfordsm: Can this be moved above the return?

ccdproc/core.py
```diff
@@ -1558,6 +1559,142 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
 
 
+def ccdmask(ratio, findbadcolumns=False, byblocks=False, ncmed=7, nlmed=7,
+            ncsig=15, nlsig=15, lsigma=9, hsigma=9, ngood=5):
+    '''
+    Uses method based on the IRAF ccdmask task to generate a mask based on the
+    given input.
+
+    Parameters
+    ----------
```
crawfordsm: Can you order these as they are in the parameter list?   Also there is no entry for byblocks

ccdproc/tests/test_ccdmask.py
```diff
@@ -0,0 +1,192 @@
...
+    ratio = flat1.divide(flat2)
+    mask = ccdmask(ratio, ncsig=11, nlsig=15, findbadcolumns=False)
+    target_mask[0:8,2] = True
+    target_mask[11:,2] = True
+    assert_array_equal(mask, target_mask)
+
+
+    mask = ccdmask(ratio, ncsig=11, nlsig=15, findbadcolumns=True)
+    target_mask[:,2] = True
+    assert_array_equal(mask, target_mask)
```
crawfordsm: can you add a test that includes byblocks?

crawfordsm:
@joshwalawender thanks so much for the contribution -- it will be useful to have in there.   I've added some more comments and sorry it has taken me a while to review

MSeifert04:
Just one additional thought: It's always useful to have some examples in the docstring, so people know what each option does. Maybe you could demonstrate the effect of `findbadcolumns` there:

```
.. plot::
   :include-source:
   >>> import matplotlib.pyplot as plt
   >>> import numpy as np
   >>> import ccdproc
   >>> arr = np.random.random((20, 20))  # using astropy.utils.NumpyRNGContext provides an interface to get identical results each run.
   >>> arr[4,4] = 3
   >>> arr[4,5] = 4
   >>> arr[4,6] = 3
   >>> arr[4,8] = 5
   >>> mask1 = ccdproc.ccdmask(arr, findbalcolumns=False)
   >>> mask2 = ccdproc.ccdmask(arr, findbalcolumns=True)
   >>> ax2 = plt.subplot(1, 3, 1)
   >>> ax2.imshow(arr, origin='lower', interpolation='none', cmap=plt.cm.gray)
   >>> ax1 = plt.subplot(1, 3, 2)
   >>> ax1.imshow(mask1, origin='lower', interpolation='none', cmap=plt.cm.gray)
   >>> ax2 = plt.subplot(1, 3, 3)
   >>> ax2.imshow(mask2, origin='lower', interpolation='none', cmap=plt.cm.gray)
```

This is just some rough idea, I don't even know if it would work 😄

MSeifert04:
Unfortunatly there are some conflicts now that require you to rebase this PR. 

If you need any help, just let me (or us) know! 

Just one point if you try it on your own: make a copy of the current branch so you can always recover if something goes wrong.

ccdproc/core.py
```diff
@@ -1558,6 +1576,151 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0,
...
+        If set to True, the code will search for bad column sections.  Note that
+        this treats columns as special and breaks symmetry between lines and
+        columns and so is likely only appropriate for detectors which have
+        readout directions.
+
+    byblocks : `bool`, optional
+        If set to true, the code will divide the image up in to blocks of size
+        nlsig by ncsig and determine the standard deviation estimate in each
+        block (as described in the original IRAF task, see Notes below).  If
+        set to False, then the code will use `percentile_filter` in
```
MSeifert04: Instead of saying "... in scipy.ndimage" it should link directly to `scipy`:

```
`scipy.ndimage.percentile_filter`
```

Otherwise the sphinx build fails.

mwcraig:
Add the functionality in #432

PR #416, opened 2016-11-02
Add combine_uncertainty_func parameter to combine.

### LINKED ISSUE #224: Allow to chose a function for uncertainty calculation in Combiner

### LINKED ISSUE #424: combine fails with StdDevUncertainty?
It looks like `combine` fails for images with a `StdDevUncertainty`.  I'm working on a simple reduction and am trying to carry uncertainty though the whole process.  While writing the bias file combine step, I got an error which I can reproduce below:

```
import ccdproc
import numpy as np
import astropy.units as u
from astropy.nddata import StdDevUncertainty

readnoise = ccdproc.Keyword('RDNOISE', unit=u.electron)
readnoise.value = 10.
bias_images = []
for i in range(3):
    bias_image = ccdproc.CCDData(np.ones((16,16), dtype=int)*1000, unit='adu')
    bias_image.uncertainty = StdDevUncertainty(np.ones(bias_image.data.shape) * readnoise.value)
    bias_images.append(bias_image)
    
master_bias = ccdproc.combine(bias_images)
```

Yields:

```
Traceback (most recent call last):
  File "quicktest.py", line 22, in <module>
    master_bias = ccdproc.combine(bias_images)
  File "/Users/vysosuser/git/ccdproc/ccdproc/combiner.py", line 585, in combine
    size_of_an_img += ccd.uncertainty.nbytes
AttributeError: 'StdDevUncertainty' object has no attribute 'nbytes'
```

Am I using uncertainty correctly or should I be doing this differently?

---

Don't forget to include the version of astropy, ccdproc and numpy, just copy
this into your Python interpreter (without the backticks):

```
>>> import astropy
>>> print(astropy.__version__)
1.0.4
>>> import ccdproc
>>> print(ccdproc.__version__)
1.2.dev967
>>> import numpy
>>> print(numpy.__version__)
1.9.3
```

MSeifert04:

For documentation changes:

For bugfixes:

For new functionality:

- [x] Fixes #224, maybe also fixes #424

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -449,7 +449,7 @@ def combine(img_list, output_file=None,
             sigma_clip=False,
             sigma_clip_low_thresh=3, sigma_clip_high_thresh=3,
             sigma_clip_func=ma.mean, sigma_clip_dev_func=ma.std,
-            dtype=None, **ccdkwargs):
+            dtype=None, combine_uncertainty_function=None, **ccdkwargs):
     """
     Convenience function for combining multiple images.
 
@@ -540,6 +540,11 @@ def combine(img_list, output_file=None,
         `ccdproc.Combiner`. If ``None`` this is set to ``float64``.
         Default is ``None``.
 
+    combine_uncertainty_function : callable, None, optional
+        If ``None`` use the default uncertainty func when using average or
+        median combine, otherwise use the function provided.
+        Default is ``None``.
+
     ccdkwargs : Other keyword arguments for `ccdproc.fits_ccddata_reader`.
 
     Returns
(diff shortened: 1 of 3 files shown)


ccdproc/combiner.py
```diff
@@ -581,12 +592,23 @@ def combine(img_list, output_file=None,
...
+    except AttributeError:
+        pass
+    # Mask is enforced to be a numpy.array across astropy versions
     if ccd.mask is not None:
         size_of_an_img += ccd.mask.nbytes
-    if ccd.flags is not None:
+    # flags is not necessarily a numpy array so do not fail with an
+    # AttributeError in case something was set!
+    # TODO: Since flags is not supported for combine and read this could
+    #       also be dropped?
```
crawfordsm: let's keep it for right now--flags isn't treated but with the request for handling data quality planes this might be something which is necessary

MSeifert04: Certainly - but at this point this just calculates the necessary memory for `Combiner` and `Combiner` ignores the flags. Should I remove that comment?

crawfordsm: If it's not a problem, leave it in and maybe indicate it is a placeholder for when flags is supported.

MSeifert04: ok, done.

PR #417, opened 2016-11-02
Travis matrix changes

MSeifert04:
This removes some of the tests in `.travis.yml`. I think I've incorporated all of them in the remaining tests.

It also adds one more test for numpy-dev (which will fail 😄 )

This reduces the build-time from 35m -> 25m.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,35 +30,34 @@ env:
         - CONDA_CHANNELS='astropy-ci-extras astropy'
 
     matrix:
-        # Make sure that egg_info works without dependencies
+
         - SETUP_CMD='egg_info'
-        # Try all python versions with the latest numpy
-        - SETUP_CMD='test'
 
 matrix:
     include:
 
-        # Do a coverage test in Python 2.
+        # Try Astropy/NumPy development versions. This requires them to be
+        # compiled during setup which takes some time.
+        - python: 3.5
+          env: ASTROPY_VERSION=development NUMPY_VERSION=dev
+
+        # Do coverage tests for both latest Python versions
+        - python: 3.5
+          env: SETUP_CMD='test --coverage'
+
         - python: 2.7
           env: SETUP_CMD='test --coverage'
 
-        # Check for sphinx doc build warnings - we do this first because it
-        # may run for a long time
+        # Check for sphinx doc build warnings
         - python: 2.7
           env: SETUP_CMD='build_sphinx -w'
-
-        # Try Astropy development version
-        - python: 2.7
-          env: ASTROPY_VERSION=development
-        - python: 3.5
-          env: ASTROPY_VERSION=development
 
         # Check compatibility with the current astropy LTS release
         - python: 2.7
           env: ASTROPY_VERSION=LTS
 
-        # Try older numpy versions
-        - python: 2.7
+        # Try older numpy, python versions
+        - python: 3.4
           env: NUMPY_VERSION=1.10
 
         # Try numpy pre-release version. This runs only when a pre-release
--- a/ccdproc/tests/test_ccdproc.py
+++ b/ccdproc/tests/test_ccdproc.py
@@ -875,10 +875,8 @@ def test_wcs_project_onto_scale_wcs(ccd_data):
     assert new_ccd.wcs.wcs.compare(target_wcs.wcs)
 
     # Define a cutout from the new array that should match the old.
-    new_lower_bound = (np.array(new_ccd.shape) -
-                       np.array(ccd_data.shape)) / 2
-    new_upper_bound = (np.array(new_ccd.shape) +
-                       np.array(ccd_data.shape)) / 2
+    new_lower_bound = (np.array(new_ccd.shape) - np.array(ccd_data.shape)) // 2
+    new_upper_bound = (np.array(new_ccd.shape) + np.array(ccd_data.shape)) // 2
     data_cutout = new_ccd.data[new_lower_bound[0]:new_upper_bound[0],
                                new_lower_bound[1]:new_upper_bound[1]]
 
@@ -890,7 +888,7 @@ def test_wcs_project_onto_scale_wcs(ccd_data):
 
     # Mask should be true for four pixels (all nearest neighbors)
     # of the single pixel we masked initially.
-    new_center = new_ccd.wcs.wcs.crpix
+    new_center = np.array(new_ccd.wcs.wcs.crpix, dtype=int, copy=False)
     assert np.all(new_ccd.mask[new_center[0]:new_center[0]+2,
                                new_center[1]:new_center[1]+2])
 


MSeifert04:
@mwcraig Could you take a look at the failure in numpy dev? It doesn't look like the code is incorrect just the test and GitHub [blames](https://github.com/astropy/ccdproc/blame/master/ccdproc/tests/test_ccdproc.py#L894) you 😄 

<details>
```
=================================== FAILURES ===================================

_______________________ test_wcs_project_onto_scale_wcs ________________________

ccd_data = CCDData([[ 1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,

           1.,  1.,  1.,  1.,  1.,  1.,  1.....,  1.,

           1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,

           1.,  1.,  1.,  1.,  1.]])

    @pytest.mark.data_size(31)

    def test_wcs_project_onto_scale_wcs(ccd_data):

        # Make the target WCS with half the pixel scale and number of pixels

        # and the values should drop by a factor of 4.

    

        ccd_data.wcs = wcs_for_testing(ccd_data.shape)

    

        # Make sure wcs is centered at the center of the center pixel.

        ccd_data.wcs.wcs.crpix += 0.5

    

        # Use uniform input data value for simplicity.

        ccd_data.data = np.ones_like(ccd_data.data)

    

        # Make mask zero...

        ccd_data.mask = np.zeros_like(ccd_data.data)

        # ...except the center pixel, which is one.

        ccd_data.mask[int(ccd_data.wcs.wcs.crpix[0]),

                      int(ccd_data.wcs.wcs.crpix[1])] = 1

    

        target_wcs = wcs_for_testing(ccd_data.shape)

        target_wcs.wcs.cdelt /= 2

    

        # Choice below ensures we are really at the center pixel of an odd range.

        target_shape = 2 * np.array(ccd_data.shape) + 1

        target_wcs.wcs.crpix = 2 * target_wcs.wcs.crpix + 1 + 0.5

    

        # Explicitly set the interpolation method so we know what to

        # expect for the mass.

        new_ccd = wcs_project(ccd_data, target_wcs,

                              target_shape=target_shape,

                              order='nearest-neighbor')

    

        # Make sure new image has correct WCS.

        assert new_ccd.wcs.wcs.compare(target_wcs.wcs)

    

        # Define a cutout from the new array that should match the old.

        new_lower_bound = (np.array(new_ccd.shape) -

                           np.array(ccd_data.shape)) / 2

        new_upper_bound = (np.array(new_ccd.shape) +

                           np.array(ccd_data.shape)) / 2

        data_cutout = new_ccd.data[new_lower_bound[0]:new_upper_bound[0],

>                                  new_lower_bound[1]:new_upper_bound[1]]

E       TypeError: slice indices must be integers or None or have an __index__ method

ccdproc/tests/test_ccdproc.py:883: TypeError

=============== 1 failed, 314 passed, 2 skipped in 10.29 seconds ===============
```
</details>

.travis.yml
```diff
@@ -30,35 +30,34 @@ env:
...
 
 matrix:
     include:
 
-        # Do a coverage test in Python 2.
-        - python: 2.7
+        # Try Astropy/NumPy development versions. This requires them to be
+        # compiled during setup which takes some time.
+        - python: 3.5
+          env: ASTROPY_VERSION=development NUMPY_VERSION=dev
```
bsipocz: I'm not sure I wouldn't mix these two, and whether it's enough to test astropy dev only with python3. However you probably don't need the numpy dev test, there is already a prerelease test that will activated when they are close to a release.

.travis.yml
```diff
@@ -30,35 +30,34 @@ env:
...
+
+        # Do coverage tests for both latest Python versions
+        - python: 3.5
           env: SETUP_CMD='test --coverage'
 
-        # Check for sphinx doc build warnings - we do this first because it
-        # may run for a long time
         - python: 2.7
-          env: SETUP_CMD='build_sphinx -w'
+          env: SETUP_CMD='test --coverage'
```
bsipocz: Why are you doing coverage for both versions? Only one will report back to the PR anyway.

MSeifert04: also it's always good to have py2k and py3k coverage reports. Just to make sure something like `dict.itervalues()` doesn't go unnoticed.

bsipocz: Yes, I suppose you're right, both seems to matter in the status. However I'm not sure about the second comment, you can save a minute by not running it on py2.

.travis.yml
MSeifert04: Astropy-dev tests against numpy-dev so in theory this mixed test should only show problems with one of both? 
I don't know if Python2 with astropy-dev should be tested. Should I re-include it?

bsipocz: Depends on your userbase, probably it's not that critical. However it's worth remembering to run the tests for both when the RC comes out.

.travis.yml
MSeifert04: If it was a minute I'd remove it again but it seems more like 10-20 seconds (or travis fluctuations):

- [2:50 (without coverage)](https://travis-ci.org/astropy/ccdproc/builds/172683391)
- [2:50 and 3:05 (with coverage)](https://travis-ci.org/astropy/ccdproc/builds/172698216)

`ccdproc` does not have as many lines as astropy. I think it will take quite a while until `coverage` significantly slows down the testing.

.travis.yml
MSeifert04: Is there a `Travis` directive for astropy "pre-releases"?

bsipocz: But: fun hackday project is to write a bot that opens temporary PRs to given packages when the RC comes out to test for any braking codes. It's a bit messy as the PRs shouldn't be merged but closed, but doing it once in every half year may be worth it. 
I wonder what @astrofrog and @eteq think about it?

MSeifert04:
I added one commit to test if the failure in numpy-dev is fixed. If the CI passes I'll commit the fix in a seperate PR.

mwcraig:
Looks like travis is pretty backlogged at the moment. I'd be fine with switching the coverage test to python 3.5-only, though I don't feel that strongly about it. We should still test on python 2.7 even if we drop the coverage test.

I think it is fine to leave the commit that fixes numpy-dev in this PR instead of opening a new one.

Just to be totally clear, feel free to merge once we have green lights.

bsipocz:
I think we have to get use to the idea of the backlog, their osx system is under a heavy load normally, they just added a new graph showing the backlog here, it's ~700 jobs atm: https://www.traviscistatus.com/#month

MSeifert04:
@bsipocz Fortunatly `ccdproc` does not test against osx. 

Not sure if that should be included in the matrix but currently the travis osx builds are either broken or they have several hundred jobs backlog.

> Just to be totally clear, feel free to merge once we have green lights.

@crawfordsm Do you have any objections or suggestions? Otherwise I'll merge this tomorrow.

bsipocz:
@MSeifert04 - I don't think it's broken, just terribly overloaded (their capacity seems to be at least 10x smaller than for linux).

MSeifert04:
Yeah, but they also had several issues with OS X lately (at least I've seen some travis status messages regarding OS X issues). I didn't meant currently, currently it's - like you said - _only_ backlog. 😄

mwcraig:
@MSeifert04 -- if you think this will fix https://travis-ci.org/astropy/ccdproc/jobs/176807262 could you please go ahead and merge this?

MSeifert04:
yes, exactly that error should be fixed. I'll merge.

Issue #326, opened 2016-03-14
arithmetic with ndarray


crawfordsm:
Currently arithmetic between an ndarray and CCDData object is not possible, but raises an error when it is attempted.   It does raise a sensible error, but this might be functionality that we do want to enable.

mwcraig:
My main question here would be how to handle units; certainly it ought to be possible at a minimum to do something like `ccd.add(my_numpy_array * u.adu)`. Should we maybe use the `@support_nddata` decorator for this?

MSeifert04:
I've solved it during the `NDArithmetic`-Rafactoring by just converting the operand to another instance of the same class ``operand = self.__class__(operand)``. This would automatically handle units/masks/etc appropriatly while raising errors in case it is not convertible. Which should be fine because I cannot think how such an operation should be defined if the class doesn't know how to interpret it. It's not a possibility with `CCDData` though because you enforce a unit. 

A possible alternative would be to just edit: https://github.com/astropy/ccdproc/blob/master/ccdproc/ccddata.py#L329 to `elif isinstance(other, (numbers.Number, np.ndarray)):`

That would fail for `add`, `subtract` (because of the unit, except the `CCDData` is `dimensionless_unscaled`) but at least work for `multiply` and `divide`.

BrunoSanchez:
Hello, 
I think I have an corner case here, I've found this error while coding some image reduction
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/home/bruno/.virtualenvs/corral/src/corral/corral/run/step.py", line 32, in run
    generator = step.process(obj) or []
  File "/home/bruno/.virtualenvs/corral/src/corral/toritos/toritos/steps.py", line 430, in process
    for apaw in ccdpawlist]
  File "/home/bruno/.virtualenvs/corral/local/lib/python2.7/site-packages/ccdproc/log_meta.py", line 68, in wrapper
    result = func(*args, **kwd)
  File "/home/bruno/.virtualenvs/corral/local/lib/python2.7/site-packages/ccdproc/core.py", line 608, in subtract_dark
    result = ccd.subtract(master_scaled)
  File "/home/bruno/.virtualenvs/corral/local/lib/python2.7/site-packages/ccdproc/ccddata.py", line 444, in subtract
    result = super(CCDData, self).subtract(other)
[...]

PR #426, opened 2016-11-23
Use NDArithmeticMixin methods if astropy >= 1.2

MSeifert04:

For documentation changes:

For bugfixes:

For new functionality:

This is currently work-in-progress. But since the astropy 1.2 release I wanted to give this a shot.

It uses the `NDArithmeticMixin` methods for add/subtract/multiply/divide, which allows to use CCDData in arithmetic operations with numpy-arrays, scalars and some other objects. For this to work I needed to remove the mandatory `CCDData` unit - because `NDArithmeticMixin` converts the second operator to a `CCDData` which fails if it's a numpy.ndarray or scalar.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,7 @@ matrix:
         # Check compatibility with the current astropy LTS release
         - python: 2.7
           env: ASTROPY_VERSION=LTS
+               SETUP_CMD='test --coverage'
 
         # Try older numpy, python versions
         - python: 3.4
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -703,6 +703,7 @@ def gain_correct(ccd, gain, gain_unit=None):
         gain_value = gain
 
     result = ccd.multiply(gain_value)
+    result.meta = ccd.meta.copy()
     return result
 
 
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -684,7 +684,7 @@ def fits_ccddata_reader(filename, hdu=0, unit=None, hdu_uncertainty='UNCERT',
         else:
             mask = None
 
-        if hdu_flags in hdus:
+        if hdu_flags is not None and hdu_flags in hdus:
             raise NotImplementedError('loading flags is currently not '
                                       'supported.')
 
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -25,6 +25,9 @@ from distutils.version import LooseVersion
 
 _ASTROPY_LT_1_2 = LooseVersion(astropy.__version__) < LooseVersion('1.2')
 
+if not _ASTROPY_LT_1_2:
+    from astropy.utils.decorators import sharedmethod
+
 if _ASTROPY_LT_1_2:
 
     class ParentNDDataDescriptor(object):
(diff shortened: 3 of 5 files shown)


MSeifert04:
Related to (and maybe fixes) #326

MSeifert04:
and if the `ValueError` shouldn't be removed when no `unit` is given let me know. That would require hacking into the `NDArithmeticMixin` internals.

mwcraig:
My initial reaction is that it would be nice to preserve the unit requirement, and that perhaps it is a change I'd only want to make in a backwards-incompatible release like 2.0. It isn't backwards-incompatible in the *code* sense, but the *behavior* would be quite a bit different than what it has been in the past.

~~Is there any way to pass the unit information in to the mixin (I honestly do not remember the details). If not, might it be possible to infer the units of the result and add them afterwards?~~

Sorry, just understood what the issue is...will think some more about workarounds.

MSeifert04:
It's not complicated to preserve the unit requirement and just _temporarly_ disable it when the `NDArithmeticMixin` runs (and tries to convert everything to `CCDData`). If that would be better I'll adjust the PR.

mwcraig:
Yes, please! 😀

MSeifert04:
In the end I used a global variable to temporarly disable the unit-checks. I thought about using a contextmanager or a global dictionary but in the end that's much more effort for just one trigger. But I realize that this might not be optimal.

I'll add some tests later.

ccdproc/core.py
```diff
@@ -703,6 +703,7 @@ def gain_correct(ccd, gain, gain_unit=None):
         gain_value = gain
 
     result = ccd.multiply(gain_value)
+    result.meta = ccd.meta.copy()
```
mwcraig: This is also unrelated to the broader change, correct? (agreed it should be fixed) 

Is there a regression test we should add here?

.travis.yml
```diff
@@ -55,6 +55,7 @@ matrix:
         # Check compatibility with the current astropy LTS release
         - python: 2.7
           env: ASTROPY_VERSION=LTS
+               SETUP_CMD='test --coverage'
```
MSeifert04: More to have a coverage of the LTS branch, otherwise either the coverage decreases significantly or I had to put a `# pragma no cover` after each of the arithmetic function.

ccdproc/core.py
MSeifert04: Now that you mention it, I investigated a bit and found why this change was necessary:

The `_ccddata_arithmetic` makes sure the meta is kept when the other operand is a scalar (not CCDData), the `NDArithmeticMixin` however defaults (because of backwards-compatibility-reasons) to no discarding the `meta`. 
Note that currently operations between two CCDData independant of the astropy version discard their meta, it was just scalars that had the special case.

MSeifert04: But I don't know if it's worth keeping the meta only if and only if the other operand has none. That's inconsistent with the case that the other operand is a CCDData, then the meta is always discarded.

mwcraig:
Any idea what is causing the test fail? Doesn't seem related to this at all.

mwcraig:
I think (despite the line numbers reported in the test failure) that this is maybe because the add, etc, are no longer methods?

mwcraig:
Code-wise I like the last commit (less repetition is better), not sure how to fix the references in the documentation, though...

MSeifert04:
I changed it so it's not using `functools.wraps` anymore and write the docstring manually. At least locally that fixed sphinx and did not fail the tests.

MSeifert04:
I guess it was the lazy-loading PR (https://github.com/astropy/astropy/pull/5065) merge which breaks the tests against astropy-dev.

MSeifert04:
The astropy-dev failure bugfix is in #435 so this will fail the corresponding travis test. But the test before the "bugfix" reversal passed.

mwcraig:
Merging, will check the builds in the morning after merging.

PR #428, opened 2016-12-06
Remove unused folders and "CaseInsensitiveOrderedDict".

MSeifert04:
The `CaseInsensitiveOrderedDict` looks generally useful but isn't used anywhere? The other folders and files seem like templates but without being used right now.

Feel free to close this PR if these folders/files/classes don't need deletion.


bsipocz:
The ``cextern`` and ``scripts`` definitely seems to be a leftover from the package template that should be deleted as not being used here.

mwcraig:
This looks pretty straightforward, merging. I think the dict class was left over from some of the early discussions about including a more extensive metadata class in ccpdroc.

PR #422, opened 2016-11-17
Fix ccds generator from ImageFileCollection (see issue #421)

joshwalawender:
Please have a look at the following list and replace the "[ ]" with a "[x]" if
the answer to this question is yes.

- [ ] For new contributors: Did you add yourself to the "Authors.rst" file?

For documentation changes:

- [ ] For documentation changes: Does your commit message include a "[skip ci]"?
      Note that it should not if you changed any examples!

For bugfixes:

- [x] Did you add an entry to the "Changes.rst" file?
- [ ] Did you add a regression test?
- [x] Does the commit message include a "Fixes #issue_number" (replace "issue_number").
- [ ] Does this PR add, rename, move or remove any existing functions or parameters?

For new functionality:

- [ ] Did you add an entry to the "Changes.rst" file?
- [ ] Did you include a meaningful docstring with Parameters, Returns and Examples?
- [ ] Does the commit message include a "Fixes #issue_number" (replace "issue_number").
- [ ] Did you include tests for the new functionality?
- [ ] Does this PR add, rename, move or remove any existing functions or parameters?

Please note that the last point is not a requirement. It is meant as a check if
the pull request potentially breaks backwards-compatibility.

-----------------------------------------

Fixes issue #421

PR #431, opened 2016-12-06
Use full path in ccd generator (replaces #422)

mwcraig:
This is another rebase of #422 and contains the changes originally in (and discussed at) that PR.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,7 +31,6 @@ Other Changes and Additions
 - ``combine`` now accepts a ``dtype`` argument which is passed to
   ``Combiner.__init__``. [#391, #392]
 
-
 Bug Fixes
 ^^^^^^^^^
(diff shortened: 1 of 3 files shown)


MSeifert04:
This definitly needs a regression test but we can also add that later (maybe an effort-easy issue?).

mwcraig:
I'll add a regression test tonight, might as well do it while it is fresh in my mind.

MSeifert04 (approved):
LGTM, thanks for adding the regression test.

Removing the empty line in the changelog is probably asking for a merge conflict but the other open PRs are all mine so it shouldn't be a problem :-)

mwcraig:
I'll revert that change (won't bother squashing) then this can be merged once coveralls is done (will skip CI on the un-removal of the blank line :)).

PR #435, opened 2016-12-10
Only check for HDUs if they are not None (fits reader)

MSeifert04:
I've added no regression test (because `hdu_flags=None` is the default and therefore already checked in several tests) and I added no changelog entry because it shouldn't change any behaviour.

The reason this fails is because `None in HDUList` is not allowed anymore in `astropy.io.fits` but `hdu_mask`, `hdu_uncertainty` and `hdu_flags` are allowed to be `None`.

I reported the issue upstream: https://github.com/astropy/astropy/issues/5583 but I think we should change it even if it's not fixed upstream.

### DIFF
--- a/ccdproc/ccddata.py
+++ b/ccdproc/ccddata.py
@@ -615,18 +615,18 @@ def fits_ccddata_reader(filename, hdu=0, unit=None, hdu_uncertainty='UNCERT',
     with fits.open(filename, **kwd) as hdus:
         hdr = hdus[hdu].header
 
-        if hdu_uncertainty in hdus:
+        if hdu_uncertainty is not None and hdu_uncertainty in hdus:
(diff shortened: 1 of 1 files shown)


juliantaylor:
this is a bug in astropy, `None in container` should imo not raise

embray:
I agree--it's a bug

MSeifert04:
ok, should be fixed eventually by https://github.com/astropy/astropy/pull/5589.

Thanks for the affirmation @juliantaylor @embray

mwcraig:
Hmmm, I was inclined to merge this anyway since there is no need to do the search if the key is `None`.

MSeifert04:
@mwcraig I restored the branch, if you want to reopen and merge it should be possible.

mwcraig:
Thanks, merging.

PR #436, opened 2016-12-13
Add .mailmap and update changelog/authors

mwcraig:
Opening a pull request for this mostly to trigger one last round of testing against astropy dev.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,17 @@
+Steve Crawford      <crawfordsm@gmail.com>
+Matthew Craig       <mattwcraig@gmail.com>
+Hans Moritz Günther <hgunther@mit.edu> <moritz.guenther@gmx.de>
+Hans Moritz Günther <hgunther@mit.edu>
+Anthony Horton      <anthony.horton@aao.gov.au>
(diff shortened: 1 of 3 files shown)


bsipocz:
@mwcraig - You may want to consider to turn on cron testing on travis, once a week tunning the tests on the latest master. You can also use the ``EVENT_TYPE`` env variable in ``.travis.yml`` to control which set of tests to run.

MSeifert04:
What exactly is the `.mailmap`? Does this keep track of mail changes of the gh account?

mwcraig:
@bsipocz -- definitely will. Also going to try actually automating maintenance the conda channels also.

mwcraig:
@MSeifert04 -- it is really only used by `git log` and `git shortlog` to combine contributions made by the same actual person under different names/emails into one consistent name. Completely unnecessary, but I felt like bikeshedding this morning apparently. Actually, its one advantage is that now the names in `git shortlog -s -n` should match the names we have listed in our contributors list.

MSeifert04:
@mwcraig thanks for the explanation. 👍

PR #438, opened 2016-12-13
Tests for #426 (arithmetic methods with CCDData and np.array)

MSeifert04:
This adds tests for the features added in #426.

@crawfordsm Does this close your issue #326?


MSeifert04:
@astropy/ccdproc-maintainers This should be quite simply, so I'm merging this tomorrow if you don't object 😄

PR #440, opened 2016-12-14
Using minversion for version checks and include some new 1.3 features

MSeifert04:
This PR changes manual `LooseVersion` checks in `CCDData` with `astropy.utils.minversion` (more convenience) and adds two new 1.3 features:

- `astropy.io.fits.is_fits` was changed to also accept `fts` so we don't need the custom function anymore
- `astropy.io.registry` can now delay documentation updates when registering reader/writer for the registry. This may reduce the import time _slightly_.


MSeifert04:
The coverage should go up again if 1.3 is officially released (or I add `--coverage` for the dev tests).

PR #441, opened 2016-12-18
Update astropy-helpers to v1.3

bsipocz:
This is an automated update of the astropy-helpers submodule to v1.3. This includes both the update of the astropy-helpers sub-module, and the ``ah_bootstrap.py`` file, if needed.

*This is intended to be helpful, but if you would prefer to manage these updates yourself, or if you notice any issues with this automated update, please let @astrofrog know!*


mwcraig:
Wow, @astrofrog has gone sentient and is masquerading as @bsipocz...

bsipocz:
Well, that may be better to change that handle to be the same as the name on the PR, but there is no harm in reporting issues to @astrofrog :)

PR #449, opened 2016-12-29
Adding python 3.6 to CI builds

bsipocz:
Probably we need to wait a bit more until all the dependencies become available on python 3.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ python:
     - 2.7
     - 3.4
     - 3.5
+    - 3.6
 
 env:
     global:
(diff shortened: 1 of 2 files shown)


MSeifert04:
Some small comments:

- could you include an egg-test for 3.6
- also a test for python 3.5 + np.1.11
- and make the numpy prerelease-test use python 3.6

Otherwise lots of green marks 👍 Seems like we are 3.6 compatible :)

bsipocz:
@MSeifert04 - Done. np1.11 was "stable" when this PR was opened :)

MSeifert04:
@bsipocz Thanks! 👍 

I restarted the failed tests because "only" something went wrong during the `coveralls`-command.

MSeifert04:
I'm going ahead an merge this. Long overdue :)

Thanks @bsipocz

Issue #453, opened 2017-01-25
FITS files not updated for ImageFileCollection.ccds(overwrite=True)


mkelley:
Hello,

When using the `CCDData` generator and the `overwrite=True` option, changes to the returned object are not updated in the FITS file.  However, the hdu generator does work as expected:

```
import numpy as np
from astropy.io import fits
import astropy.units as u
import ccdproc

# generate and save an empty image
im = np.zeros((10, 10))
h = fits.Header()
h['BUNIT'] = 'adu'
print('Original array:', im.mean())
fits.writeto('test.fits', im, h)

# load the image into an image collection, return the CCDData, add 1.0, and overwrite
ic = ccdproc.ImageFileCollection(location='.', filenames=['test.fits'])
for ccd in ic.ccds(overwrite=True):
    ccd += 1 * u.adu
    print('Updated CCDData:', ccd.mean())

print('Saved CCDData:', fits.getdata('test.fits').mean())

# Instead iterate over hdus, add 1.0, and overwrite
for hdu in ic.hdus(overwrite=True):
    hdu.data += 1
    print('Updated HDU:', hdu.data.mean())

print('Saved HDU:', fits.getdata('test.fits').mean())
```

Results:
```
$ python3 test.py
Original array: 0.0
Updated CCDData: 1.0 adu
Saved CCDData: 0.0
Updated HDU: 1.0
Saved HDU: 1.0
```

The saved CCDData mean is 0.0, but I expected 1.0.

astropy 2.0.dev17751, ccdproc 1.3.dev1066, and numpy 1.11.0.

Thanks for the help,
Mike

MSeifert04:
Actually there are two issues here: One is that you are correct and the `ccds`-generator doesn't update the original. However the second problem is that `ccd += 1` creates a numpy-array. So you didn't modify `ccd` but just created a new object.

I'll submit a fix for the first issue but the second one is impossible to fix without creating in-place arithmetic operations (like `__iadd__`).

mkelley:
Ah, I see I should have used `.add()`.

mkelley:
OK, now I get it, `ccd.data += 1` would be the right call.

mwcraig:
@mkelley -- is it ok to close this now?

crawfordsm:
Closing this, but please re-open or create a new issue if something more is needed.

PR #454, opened 2017-01-26
[Don't merge!] fix overwrite=True for ccds generator in ImageFileCollection

MSeifert04:

For bugfixes:

----

Maybe this PR should close the issue #453. However part of the issue was the unintentional conversion to a `np.ndarray`.

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -704,8 +704,11 @@ class ImageFileCollection(object):
 
         for full_path in self._paths():
             no_scale = do_not_scale_image_data
-            hdulist = fits.open(full_path,
-                                do_not_scale_image_data=no_scale)
+            if return_type == 'ccd':
+                ccddata = fits_ccddata_reader(full_path, **ccd_kwargs)
+            else:
+                hdulist = fits.open(full_path,
+                                    do_not_scale_image_data=no_scale)
 
             file_name = path.basename(full_path)
 
@@ -713,7 +716,7 @@ class ImageFileCollection(object):
                     'header': lambda: hdulist[0].header,
                     'hdu': lambda: hdulist[0],
                     'data': lambda: hdulist[0].data,
-                    'ccd': lambda: fits_ccddata_reader(full_path, **ccd_kwargs)
+                    'ccd': lambda: ccddata
                     }
             try:
                 yield (return_options[return_type]()  # pragma: no branch
@@ -739,11 +742,15 @@ class ImageFileCollection(object):
             nuke_existing = clobber or overwrite
             if (new_path != full_path) or nuke_existing:
                 try:
-                    hdulist.writeto(new_path, clobber=nuke_existing)
+                    if return_type == 'ccd':
+                        ccddata.write(new_path, clobber=nuke_existing)
+                    else:
+                        hdulist.writeto(new_path, clobber=nuke_existing)
                 except IOError:
                     logger.error('error writing file %s', new_path)
                     raise
-            hdulist.close()
+            if return_type != 'ccd':
+                hdulist.close()
 
         # reset mask
         for col in self.summary_info.columns:
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -786,6 +786,8 @@ class ImageFileCollection(object):
         name='image', default_scaling='False', return_type='numpy.ndarray')
 
     def ccds(self, ccd_kwargs=None, **kwd):
+        if kwd.get('do_not_scale_image_data', None):
+            raise TypeError('cannot specify "do_not_scale_image_data=True" here.')
         return self._generator('ccd', ccd_kwargs=ccd_kwargs, **kwd)
     ccds.__doc__ = _generator.__doc__.format(
-        name='CCDData', default_scaling='True', return_type='ccdproc.CCDData')
+        name='CCDData', default_scaling='False', return_type='ccdproc.CCDData')
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -705,6 +705,9 @@ class ImageFileCollection(object):
         for full_path in self._paths():
             no_scale = do_not_scale_image_data
             if return_type == 'ccd':
+                # No need to check for "do_not_scale_image_data" because if
+                # it's True then it should be already rejected by the "ccds"
+                # method.
                 ccddata = fits_ccddata_reader(full_path, **ccd_kwargs)
             else:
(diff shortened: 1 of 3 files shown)


MSeifert04:
Very weird that coverage claims that new `ccddata.write` line is not covered. 😅

ccdproc/image_collection.py
```diff
@@ -689,16 +689,19 @@ def _generator(self, return_type,
 
         for full_path in self._paths():
             no_scale = do_not_scale_image_data
-            hdulist = fits.open(full_path,
-                                do_not_scale_image_data=no_scale)
+            if return_type == 'ccd':
+                ccddata = fits_ccddata_reader(full_path, **ccd_kwargs)
```
mwcraig: Should `do_not_scale_image_data` be added to `ccd_kwargs`? It definitely wasn't before, so maybe not...but right now if it gets passed in to the generator it will be ignored for the `ccds` generator, I think.

MSeifert04: Oh, totally overlooked this.

Currently it's not possible to use `do_not_scale_image_data` because it's ["unsupported"](https://github.com/astropy/ccdproc/commit/293e31ed7706bc507d02bc03692442ea7d6eb466#diff-b3346e64cf4de448f0fab8c089b63c0dR189)

mwcraig: Might as well add it, though, in case it is supported in the future.

ccdproc/image_collection.py
```diff
@@ -764,6 +771,8 @@ def data(self, do_not_scale_image_data=False, **kwd):
         name='image', default_scaling='False', return_type='numpy.ndarray')
 
     def ccds(self, ccd_kwargs=None, **kwd):
+        if kwd.get('do_not_scale_image_data', None):
+            raise TypeError('cannot specify "do_not_scale_image_data=True" here.')
```
crawfordsm: replace 'here' with 'for CCDData objects'

crawfordsm:
Two changes:

1. Change the default  do_not_scale_image_data to False

2. And raise error if set to True for ccd objects

MSeifert04:
It seems like ["new" failure](https://travis-ci.org/astropy/ccdproc/jobs/208708854) is because of something upstream.

```
=================================== FAILURES ===================================

_____________________ TestImageFileCollection.test_sorting _____________________

self = <ccdproc.tests.test_image_collection.TestImageFileCollection object at 0x7fc9eb3d8eb8>
triage_setup = <ccdproc.tests.pytest_fixtures.triage_setup.<locals>.Result object at 0x7fc9eb365630>
    def test_sorting(self, triage_setup):
        collection = image_collection.ImageFileCollection(location=triage_setup.test_dir,
                                             keywords=['imagetyp',
                                                       'filter',
                                                       'object'])
        all_elements = []
        for hdu, fname in collection.hdus(return_fname=True):
            all_elements.append((str(hdu.header), fname))
        # Now sort
>       collection.sort(keys=['filter', 'object'])
../ccdproc-test-3mw6c30a/lib.linux-x86_64-3.5/ccdproc/tests/test_image_collection.py:664: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../ccdproc-test-3mw6c30a/lib.linux-x86_64-3.5/ccdproc/image_collection.py:301: in sort
    self._summary_info.sort(keys)
/home/travis/miniconda/envs/test/lib/python3.5/site-packages/astropy/table/table.py:2372: in sort
    indexes = self.argsort(keys)
/home/travis/miniconda/envs/test/lib/python3.5/site-packages/astropy/table/table.py:2328: in argsort
    return data.argsort(**kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = masked_array(data = [('R', --) ('R', --) ('R', --) ('R', --) ('R', --) (--, --) (--, --)],
             mask = [(False... True,  True)],
       fill_value = ('?',   1.00000000e+20),
            dtype = [('filter', 'O'), ('object', '<f8')])
axis = None, kind = 'quicksort', order = ['filter', 'object'], endwith = True
fill_value = (None, inf)
    def argsort(self, axis=None, kind='quicksort', order=None,

                endwith=True, fill_value=None):
        if fill_value is None:
            if endwith:
                # nan > inf
                if np.issubdtype(self.dtype, np.floating):
                    fill_value = np.nan
                else:
                    fill_value = minimum_fill_value(self)
            else:
                fill_value = maximum_fill_value(self)
        filled = self.filled(fill_value)
>       return filled.argsort(axis=axis, kind=kind, order=order)
E       TypeError: unorderable types: NoneType() > NoneType()
```

MSeifert04:
I think the above mentioned failure is due to https://github.com/numpy/numpy/pull/8678, however a fill-value of `(None, inf)` seems weird there and is probably causing the `TypeError`.

ccdproc/image_collection.py
MSeifert04: Fixed. Thanks for catching that.

MSeifert04:
I added a test for the case that "do_not_scale_image_data" raises the exception and the default should already be False.

MSeifert04:
Please don't merge this PR even if the tests pass. Coverage seems to insist that I did not cover the `ccd.write` line while the test should. 😕

MSeifert04:
ok, I've fixed the test and now the write-line with `ccddata` is covered. I think the PR is ready for a final review @mwcraig @crawfordsm :)

MSeifert04:
After thinking about the `CCDData.write` change. What if we read from a MEF file and we choose `overwrite=True` (or `clobber`). That would delete the previous file and only write one extension back. Should `overwrite` be disabled for the `ccds` generator in case there is a chance that one deletes unused extensions?

mwcraig:
Good question. I think we need to give more careful thought to how to handle multi-extension files. The current `CCDData` object is matched well to single-extension files, or to multi-extension files that are describing a single image (so the extensions are mask and uncertainty).

For now disallowing `overwrite` for `ccds` makes sense to me, too.

crawfordsm:
I would agree with Matt's suggestion for right now.   We can come back and revisit this after making a better plan for MEF files.

crawfordsm:
If it is easy to detect unread extensions, then disallow overwrite for those.  Otherwise, we just disallow overwriting for ccds and thrown an exception if a user tries to do it.

Issue #389, opened 2016-08-15
Combiner acception DQ arrays (bitwise value combinations)


SaraOgaz:
Hi,

i was wondering if there are any plans to have Combiner take in a DQ style array for masking, (i.e. scaling to zero).  I've been getting a lot of requests for a mscombine replacement, and I think people use this feature a lot.  If there are no plans for this, I'm thinking about adding a wrapper around Combine to enable this, which would possibly end up in the one of the STScI packages... thoughts?

MSeifert04:
@SaraOgaz What exactly doesn't work for "DQ style mask arrays"? 

`ccdproc` currently automatically casts the mask to a boolean array. Meaning that any "mask" value other than zero (so if it has any DQ flag) is treated as "masked" and ignored during combination:

```
>>> import ccdproc
>>> import numpy as np

>>> # image with "some" quality flags (totally flagged)
>>> ccd = ccdproc.CCDData(np.ones((3,3)), mask=np.ones((3,3), dtype=int), unit='adu')
>>> ccd.mask[1,1] = False  # let's unmask one pixel
>>> # image with no quality flag
>>> ccd2 = ccdproc.CCDData(np.ones((3,3)) * 2, mask=np.zeros((3,3), dtype=int), unit='adu')

>>> ccdcomb = ccdproc.Combiner([ccd, ccd, ccd2])
>>> ccdcomb.average_combine()
CCDData([[ 2.        ,  2.        ,  2.        ],
         [ 2.        ,  1.33333333,  2.        ],
         [ 2.        ,  2.        ,  2.        ]])
```

But I'm not sure if I understood the question correctly. So please correct me if I'm wrong. Or is it about the `combine` function rather than the `Combiner` class?

mwcraig:
There is no objection in principle to supporting that; a little more detail and/or an example would be helpful.

Also, you may want to take a quick look at the [Astropy Proposal for Enhancement for NDData](https://github.com/astropy/astropy-APEs/pull/14), which add supporting for DQ planes at a more fundamental level. Even a simple "+1" or "-1" for that feature would be useful :)

SaraOgaz:
[...]

PR #458, opened 2017-02-14
Bitmasks

MSeifert04:
For new functionality:

---

This is a first draft for bitmask (data quality) support. I haven't included it anywhere yet and it's still missing some (narrative-)documentation/tests/changelog. But it should be good enough for feedback for the API?

xref: #389

### DIFF
--- a/ccdproc/bitmask.py
+++ b/ccdproc/bitmask.py
@@ -30,10 +30,11 @@ _POWERS = (
 # only an unsigned may hold the result.
 _REPRESENTABLE_FLAGS = {
     'int8': 7, 'int16': 15, 'int32': 31, 'int64': 63,
-    'uint64': 64
+    'uint8': 8, 'uint16': 16, 'uint32': 32, 'uint64': 64
 }
 
-# The following could also be calculated using this statement:
+# The following could also be calculated using this statement (except for the
+# lower uint types - which should be avoided!)
 # {i: min((dtype for dtype, maxsize in REPRESENTABLE_FLAGS.items()
 #          if maxsize>=i),
 #         key=REPRESENTABLE_FLAGS.get)
@@ -385,7 +386,9 @@ class Bitmask(object):
         # If bitname or bitvals were given these are now converted to bitvalues
         # so we can do all the real processing in here.
         if bitval is not None:
-            maximum = _REPRESENTABLE_FLAGS[bitmask.dtype.name]
+            # There is no zero'th flag so subtract one from the representable
+            # flags to get the real number of flags
+            maximum = _REPRESENTABLE_FLAGS[bitmask.dtype.name] - 1
             # Make sure the maximum value doesn't exceed the limit given by the
             # dtype of the bitmask
             if max(bitval) > _POWERS[maximum]:
@@ -407,7 +410,8 @@ class Bitmask(object):
             # If we want to reverse the flag interpretation we must subtract
             # the value from the sum of all possible flags:
             if reverse_flag_interpretation:
-                cmpbitval = sum(_POWERS[:maximum+1]) - cmpbitval
+                cmpbitval = np.array(sum(_POWERS[:maximum+1]) - cmpbitval,
+                                     dtype=bitmask.dtype)
 
         # Finally create the mask using a bitwise and operation and return it.
         mask = np.zeros(bitmask.shape, dtype=bool)
--- a/ccdproc/bitmask.py
+++ b/ccdproc/bitmask.py
@@ -1,0 +1,469 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+import collections
+import itertools
+
+import numpy as np
+
+_INTEGER_DTYPE_KINDS = {'u', 'i'}
+
+# Powers of two 2**0 - 2**63
+_POWERS = (
+    1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384,
+    32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608,
+    16777216, 33554432, 67108864, 134217728, 268435456, 536870912, 1073741824,
+    2147483648, 4294967296, 8589934592, 17179869184, 34359738368, 68719476736,
+    137438953472, 274877906944, 549755813888, 1099511627776, 2199023255552,
+    4398046511104, 8796093022208, 17592186044416, 35184372088832,
+    70368744177664, 140737488355328, 281474976710656, 562949953421312,
+    1125899906842624, 2251799813685248, 4503599627370496, 9007199254740992,
+    18014398509481984, 36028797018963968, 72057594037927936,
+    144115188075855872, 288230376151711744, 576460752303423488,
+    1152921504606846976, 2305843009213693952, 4611686018427387904,
+    9223372036854775808
+)
+
+# Use signed integers for more controlled behaviour expect for 64 layers then
+# only an unsigned may hold the result.
+_REPRESENTABLE_FLAGS = {
+    'int8': 7, 'int16': 15, 'int32': 31, 'int64': 63,
+    'uint64': 64
+}
+
+# The following could also be calculated using this statement:
+# {i: min((dtype for dtype, maxsize in REPRESENTABLE_FLAGS.items()
+#          if maxsize>=i),
+#         key=REPRESENTABLE_FLAGS.get)
+#  for i in range(65)}
+_LAYERS_TO_DTYPE = {
+    0:  'int8',   1: 'int8',   2: 'int8',   3: 'int8',   4: 'int8',
+    5:  'int8',   6: 'int8',   7: 'int8',   8: 'int16',  9: 'int16',
+    10: 'int16', 11: 'int16', 12: 'int16', 13: 'int16', 14: 'int16',
+    15: 'int16', 16: 'int32', 17: 'int32', 18: 'int32', 19: 'int32',
+    20: 'int32', 21: 'int32', 22: 'int32', 23: 'int32', 24: 'int32',
+    25: 'int32', 26: 'int32', 27: 'int32', 28: 'int32', 29: 'int32',
+    30: 'int32', 31: 'int32', 32: 'int64', 33: 'int64', 34: 'int64',
+    35: 'int64', 36: 'int64', 37: 'int64', 38: 'int64', 39: 'int64',
+    40: 'int64', 41: 'int64', 42: 'int64', 43: 'int64', 44: 'int64',
+    45: 'int64', 46: 'int64', 47: 'int64', 48: 'int64', 49: 'int64',
+    50: 'int64', 51: 'int64', 52: 'int64', 53: 'int64', 54: 'int64',
+    55: 'int64', 56: 'int64', 57: 'int64', 58: 'int64', 59: 'int64',
+    60: 'int64', 61: 'int64', 62: 'int64', 63: 'int64', 64: 'uint64'
+}
+
+
+class Bitmask(object):
+    """Bitmask representation that can be converted to normal masks.
+
+    Parameters
+    ----------
+    bitmask : `numpy.ndarray` of integer data type
+        The bitmask.
+
+    name_map : sequence or mapping or None, optional
+        Provide names for the different flags of the bitmask. This allows
+        accessing the flags by name rather than by position or value of the
+        bit. If it's a sequence then the names should be listed by increasing
+        bitvalue. If it's a mapping the names should be the keys and the
+        bit position should be the key.
+
+        For example if ``1`` represents ``bad pixel`` and ``2``
+        represents ``hot pixel`` then the ``name_map`` should be either:
+        ``["bad pixel", "hot pixel"]`` or ``{"bad pixel": 0, "hot pixel": 1}``
+
+        Default is ``None``.
+
+    Examples
+    --------
+    Creating an unnamed bitmask::
+
+        >>> import numpy as np
+        >>> from ccdproc import Bitmask
+        >>> Bitmask(np.arange(10).reshape(2,5))
+        Bitmask (
+        array([[0, 1, 2, 3, 4],
+               [5, 6, 7, 8, 9]])
+        name_map={}
+        )
+
+    Or creating a named bitmask::
+
+        >>> Bitmask(np.arange(10),
+        ...         name_map=['bad', 'hot', 'artifact', 'low response'])
+        Bitmask (
+        array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+        name_map=OrderedDict([('bad', 0), ('hot', 1), ('artifact', 2), ('low response', 3)])
+        )
+
+        >>> names = {'bad': 0, 'hot': 1, 'artifact': 2, 'low response': 3}
+        >>> bm = Bitmask(np.arange(10), name_map=names)
+        >>> bm  # doctest: +SKIP
+        Bitmask (
+        array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+        name_map={'hot': 1, 'low response': 3, 'bad': 0, 'artifact': 2}
+        )
+
+    Providing the names may be of help when creating a ``mask``:
+
+        >>> bm.to_mask(bitname=['low response'])
+        array([False, False, False, False, False, False, False, False,  True,  True], dtype=bool)
+    """
+    def __init__(self, bitmask, name_map=None):
+        self.data = bitmask
+        self.name_map = name_map
+
+    def __repr__(self):
+        return ('{self.__class__.__name__} (\n{self.data!r}'
+                '\nname_map={self.name_map}\n)'.format(self=self))
+
+    @property
+    def data(self):
+        return self._data
+
+    @data.setter
+    def data(self, val):
+        # Check if it looks like a "bitmask", meaning that it should be
+        # a `numpy.ndarray` and of integer data type.
+        val = np.asarray(val)
+        if val.dtype.kind not in _INTEGER_DTYPE_KINDS:
+            raise TypeError('"bitmask" must be of integer data type.')
+        self._data = val
+
+    @property
+    def name_map(self):
+        return self._name_map
+
+    @name_map.setter
+    def name_map(self, val):
+        if val is None:
+            val = {}
+        elif isinstance(val, (list, tuple)):
+            val = collections.OrderedDict(zip(val, itertools.count()))
+        elif isinstance(val, dict):
+            val = val
+        # Do the slow checks (collections) after the more common cases
+        elif isinstance(val, collections.Mapping):
+            val = val
+        elif isinstance(val, collections.Sequence):
+            val = collections.OrderedDict(zip(val, itertools.count()))
+        else:
+            raise TypeError('"name_map" must be a list/tuple/dict or another '
+                            'type of sequence/mapping, not a {}.'
+                            .format(type(val)))
+        self._name_map = val
+
+    @classmethod
+    def from_boolarray(cls, boolarray):
+        """Convert a boolean array representation of a bitmask to an integer
+        bitmask.
+
+        Parameters
+        ----------
+        boolarray : `numpy.ndarray` of boolean dtype with at least 2 dimensions
+            The boolean array to convert to a ``bitmask``. The function assumes
+            that the first dimension of the ``boolarray`` represents the
+            different flags.
+
+        Returns
+        -------
+        bitmask : `numpy.ndarray` of integer dtype
+            The bitmask representation of the ``boolarray``. It will have the
+            same shape as ``boolarray.shape[1:]``.
+
+        See also
+        --------
+        to_boolarray
+        ccdproc.boolarray_to_mask
+        """
+        boolarray = _prep_boolarray(boolarray)
+        # Number of layers (first dimension)
+        num = boolarray.shape[0]
+        # Choose the minimal dtype to represent the bitmask
+        dtype = _LAYERS_TO_DTYPE[num]
+        bitmask = np.zeros(boolarray.shape[1:], dtype=dtype)
+        # Calculate the bitmask, note that 1 << exp is equivalent to
+        # 2 ** exp and combine it with bitwise "or"!
+        for exp in range(num):
+            bitmask |= boolarray[exp] << exp
+        return cls(bitmask)
+
+    def to_boolarray(self, num=None):
+        """Convert an integer bitmask to a boolean array representation.
+
+        Parameters
+        ----------
+        num : `int` in the range 0 - 64 (inclusive) or `None`, optional
+            The number of "layers" of the boolean array. This parameter is
+            mainly so that the function doesn't have to guess how many
+            different flags the bitmask could have and to minimize the result
+            of the final boolean array. If ``None`` the number of flags is
+            determined by the ``dtype`` of the ``bitmask``.
+            Default is ``None``.
+
+        Returns
+        -------
+        boolarray : `numpy.ndarray` of boolean dtype
+            The boolean representation of the bitmask. It will have one more
+            dimension that the original ``bitmask`` where the first dimension
+            represents the "flags". So ``boolarray[0]`` will return an array of
+            the size of the original ``bitmask`` and with values ``False`` (if
+            the first flag wasn't set) and ``True`` (if the first flag was
+            set).
+
+        Examples
+        --------
+        This converts the bitmask to its boolean representation. For example
+        if it's easier to manipulate::
+
+            >>> from ccdproc import Bitmask
+            >>> bm = Bitmask(np.arange(7))
+            >>> boolbitmask = bm.to_boolarray(num=3)
+            >>> boolbitmask
+            array([[False,  True, False,  True, False,  True, False],
+                   [False, False,  True,  True, False, False,  True],
+                   [False, False, False, False,  True,  True,  True]], dtype=bool)
+
+        The first index represents a flag, so by accessing the booleans by
+        single integers allows to manipulate, reset, overwrite a given flag
+        for the bitmask easily::
+
+            >>> boolbitmask[1]  # Get the values for the second flag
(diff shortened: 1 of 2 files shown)


MSeifert04:
cc: @mwcraig @crawfordsm @pllim @SaraOgaz

I haven't used DQ arrays often so I could use some feedback on how these are best implemented or which functions are useful in this context. After I get the interface right I check how they can be added to "read", "write", "combine" and so on.

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+        array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+        name_map=OrderedDict([('bad', 0), ('hot', 1), ('artifact', 2), ('low response', 3)])
+        )
+
+        >>> names = {'bad': 0, 'hot': 1, 'artifact': 2, 'low response': 3}
+        >>> bm = Bitmask(np.arange(10), name_map=names)
+        >>> bm  # doctest: +SKIP
+        Bitmask (
+        array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+        name_map={'hot': 1, 'low response': 3, 'bad': 0, 'artifact': 2}
```
pllim: Not sure if I understand this. DQ flag can only be the value of `2**n`. While, what a value represents can differ from instrument to instrument. HST uses `uint16`, so to define for `uint64` might seem like overkill but probably does not hurt?

References:
* [bitmask usage in AstroDrizzle](http://documents.stsci.edu/hst/HST_overview/documents/DrizzlePac/ch64.html#590491) used by HST.
* [HST/ACS DQ flags](http://www.stsci.edu/hst/acs/analysis/reference_files/data_quality_flags.html)
* [HST/WFC3 DQ flags](http://www.stsci.edu/hst/wfc3/design/documents/handbooks/currentIHB/appendixE2.html#49765)
* [DQInspect plugin for Ginga](http://stginga.readthedocs.io/en/latest/stginga/plugins_manual/dqinspect.html)

pllim: Maybe @mcara can explain on the specifics of AstroDrizzle usage. Would be nice if the API is somewhat consistent between the two, as HST users are already probably used to AstroDrizzle API (or maybe not if they only use it like a black box).

MSeifert04: can you elaborate what you don't understand? 

Is the possibility to pass in a `name_map` counter-intuitive because it's too complicated? Is it the values (I chose the exponent in the `name_map` but the `to_mask` method allows bitvalues (2, 4, 8, 16), the bit positions (aka exponents) (0, 1, 2, 3) or `names` (chose the bit position from the dictionary).

The final `dtype` is actually depending on the input, so the `uint64`-arrays shouldn't be common.

pllim: @MSeifert04 , oh, it is the *exponent*... I see now. I don't think that kind of user-facing API is intuitive. In the handbooks, we publish the actual value (`2**n`), not the exponent value. User should not be forced to memorize what `n` maps to `2**n` (while that comes naturally to some people, it increases the chance of user errors). Does this make sense?

mcara: @pllim @MSeifert04 Please give me a chance to take a look at this. Meanwhile, if interested, take a look at our implementation used in STScI code: https://github.com/spacetelescope/stsci.tools/blob/master/lib/stsci/tools/bitmask.py

pllim: Oh, yeah. I forgot I also wrote [interpretdq.py](https://github.com/spacetelescope/reftools/blob/master/reftools/interpretdq.py), which is basically what `DQInspect` plugin for Ginga is using. :sweat_smile:

MSeifert04: @pllim That's why I asked. 

I found it easier to add `to_boolarray` and `from_boolarray` methods which converts the bitmask to a n+1 dimensional boolean array where the first dimension represented the "exponent", which made it easier to manipulate, inspect and extend in interactive use. 

But I've never worked "professionally" with DQ masks. :)

MSeifert04: haha, I should've checked the STScI packages before re-inventing-the-wheel (again). 😄 😅

mcara:
@SaraOgaz should have checked this before opening https://github.com/astropy/ccdproc/issues/389

ccdproc/bitmask.py
mcara: @SaraOgaz should have checked this before opening https://github.com/astropy/ccdproc/issues/389

MSeifert04: @mcara #389 was about implementing bitmasks in the ccdproc convenience function `combine`. This PR was meant as "low-level-interface" to enable that. I just didn't knew there were already existing implementations.

MSeifert04: The "AURA" license is basically BSD-3-clause so should be compatible with `ccdproc`, right?

mcara: I think so, but it is better to ask @stsci-hack

pllim: @MSeifert04 , I am pretty sure it is. PyFITS was ported to `astropy.io.fits` without problem.

SaraOgaz: Hey @MSeifert04, thanks for working on this! Sorry I've been MIA the past week.  Do you want me to test this PR out and give some feedback, or should I wait for an update?

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+        See also
+        --------
+        to_boolarray
+        ccdproc.boolarray_to_mask
+        """
+        boolarray = _prep_boolarray(boolarray)
+        # Number of layers (first dimension)
+        num = boolarray.shape[0]
+        # Choose the minimal dtype to represent the bitmask
+        dtype = _LAYERS_TO_DTYPE[num]
```
mcara: I would suggest getting rid of the ``_LAYERS_TO_DTYPE`` monstrosity 😄, e.g.,
```python
def min_dtype(nplanes):
    nbytes = 1<<int(np.ceil(np.log2(npl / 8.0)))
    if nbytes > 8:
        raise TypeError("Too many layers to hold in a built-in unsigned integer data type")
    return np.dtype('u{:d}'.format(nbytes))
```

pllim:
@mcara , approving a PR means you think it is okay to be merged. Since you will have many more comments, approving it is not the right thing to do. FYI.

ccdproc/bitmask.py
mcara: Also, a note for the future: you may consider returning multi-layer arrays of flags... maybe...

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+    1125899906842624, 2251799813685248, 4503599627370496, 9007199254740992,
+    18014398509481984, 36028797018963968, 72057594037927936,
+    144115188075855872, 288230376151711744, 576460752303423488,
+    1152921504606846976, 2305843009213693952, 4611686018427387904,
+    9223372036854775808
+)
+
+# Use signed integers for more controlled behaviour expect for 64 layers then
+# only an unsigned may hold the result.
+_REPRESENTABLE_FLAGS = {
```
mcara: I personally prefer using unsigned int for flag-type variables (maybe because this is what I used to use in C/C++ - see, e.g., DWORD type often returned by many Windows system calls is an unsigned integer: https://msdn.microsoft.com/en-us/library/windows/desktop/aa383751(v=vs.85).aspx).

In addition, ``_REPRESENTABLE_FLAGS`` probably can be determined programmatically. Also, you say:
> Use signed integers for more controlled behaviour

I would think that using unsigned would be more consistent.

Finally, it is strange that 2 byte int16 can represent only 15 bits... Something is not right here.

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+        bitmask = self.data
+        if num is not None and 0 > num > len(_POWERS) - 1:
+            raise ValueError(
+                '"num" must be between 0 and {} (both inclusive).'
+                .format(len(_POWERS) - 1))
+        # num is a space-optimization, if there is no "num" given
+        # get the number of representable flags by the dtype of
+        # the bitmask. An alternative would be to find the maximum
+        # and use the maximal power of 2 that is smaller or equal
+        # to the maximum. However hat would require a whole pass
```
mcara: **t**hat

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+        # over the array and using the dtype makes sure we use
+        # exactly the same memory as the original array!
```
mcara: I am not sure I get the idea of 
> using the dtype makes sure we use exactly the same memory as the original array!

You are creating a boolean array (probably equivalent to unit8) for every bit in the input. How can it be that you will use the same memory as the original array???

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+        if num is None:
+            num = _REPRESENTABLE_FLAGS[bitmask.dtype.name]
```
mcara: I think this approach is a waste of memory and ``num`` should be minimized, e.g.,
```python
            num = int(np.log2(np.amax(bitmask)))
```

mcara: In addition to saving memory it also saves time due to possibly multiple pass reduction in the last loop of this function.

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+        arr = np.zeros(tuple(itertools.chain([num], bitmask.shape)),
+                       dtype=bool)
+        # Bitwise and checks if the bitmask contains a set flag
+        # for this exponent (layer of the boolmask).
+        for exp in range(num):
+            arr[exp] = bitmask & _POWERS[exp]
```
mcara:
```python
arr[exp] = bitmask & (1<<exp)
```
``1<<exp`` is as fast as accessing array; it is also much safer.

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+
+import collections
+import itertools
+
+import numpy as np
+
+_INTEGER_DTYPE_KINDS = {'u', 'i'}
+
+# Powers of two 2**0 - 2**63
+_POWERS = (
```
mcara: Man, I've never seen anything like this! Please, please, get rid of ``_POWERS``!

MSeifert04: Will do. 😄

ccdproc/bitmask.py
mcara: However, I am not convinced the above formula is "good enough"... It is definitely "good enough" if you assume that signed ints cannot use last bit (that is, if all flags are positive). However, it is not good enough if integer flags can be negative (why not?)

ccdproc/bitmask.py
MSeifert04: I didn't put too much thought into that. It was just that all DQ arrays I had access to were saved as signed integers so I assumed that was the conventional way.

> it is strange that 2 byte int16 can represent only 15 bits

I think that's because I mixed some concepts. That's the highest bit (counting from zero). At least I hope so, I'll definetly check that again.

ccdproc/bitmask.py
MSeifert04: @SaraOgaz If you have some time then that would be great.

I definetly don't know what should be implemented and what would be the best API (methodnames, parameters, expected output) for that.

ccdproc/bitmask.py
MSeifert04: `numpy.uint` arrays behave strange with anything else than "bitwise" operations. It's really easy to get signed integer (or float) arrays by accident. And the DQ FITS extensions I've seen were saved as signed integers.

> returning multi-layer arrays of flags... maybe...

sorry, I don't understand that?

ccdproc/bitmask.py
MSeifert04: ups, I forget the word "amount". That should mean "it uses the same amount of memory". Not the same memory on disk

ccdproc/bitmask.py
MSeifert04: But that discards potentially valid flags when they are all False. I wasn't sure if that's common or a problem. This is the first draft and I tried to keep on the safe side.

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+        boolarray = _prep_boolarray(boolarray)
+        # Number of layers (first dimension)
+        num = boolarray.shape[0]
+        # Choose the minimal dtype to represent the bitmask
+        dtype = _LAYERS_TO_DTYPE[num]
+        bitmask = np.zeros(boolarray.shape[1:], dtype=dtype)
+        # Calculate the bitmask, note that 1 << exp is equivalent to
+        # 2 ** exp and combine it with bitwise "or"!
+        for exp in range(num):
+            bitmask |= boolarray[exp] << exp
```
mcara: Try to see if
```python
bitmask  = boolarray.sum(axis=0, dtype=bool).astype(dtype)
```
is any more efficient.

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+            name_map={}
+            )
+
+        Notes
+        -----
+        The conversion from and to boolean representation can be quite slow so
+        shouldn't be used excessivly.
+        """
+        bitmask = self.data
+        if num is not None and 0 > num > len(_POWERS) - 1:
```
mcara: ```if num is not None and 0 > num > 8*_MAX_UINT_BYTES:```

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+            raise ValueError(
+                '"num" must be between 0 and {} (both inclusive).'
+                .format(len(_POWERS) - 1))
```
mcara: .format(8 * _MAX_UINT_BYTES - 1))

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+        # num is a space-optimization, if there is no "num" given
+        # get the number of representable flags by the dtype of
+        # the bitmask. An alternative would be to find the maximum
+        # and use the maximal power of 2 that is smaller or equal
+        # to the maximum. However hat would require a whole pass
+        # over the array and using the dtype makes sure we use
+        # exactly the same memory as the original array!
+        if num is None:
+            num = _REPRESENTABLE_FLAGS[bitmask.dtype.name]
+        arr = np.zeros(tuple(itertools.chain([num], bitmask.shape)),
```
mcara: arr = np.zeros((num, ) + bitmask.shape), dtype=bool)

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+
+        # So that the logic and validation is only applied once process
+        # first "bitname" to "bitpos" and then "bitpos" to "bitval". So each
+        # step is quite short and well defined and the logic can be implemented
+        # only in the "bitval" branch.
+
+        # Map the names to bit positions
+        if bitname is not None:
+            try:
+                bitpos = [self.name_map[name] for name in bitname]
```
mcara: bitval = [1<<self.name_map[name] for name in bitname]

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+            except KeyError:
+                raise ValueError(
+                    'The "name(s)" {} were not found in "self.name_map" ({})'
+                    .format([name for name in bitname
+                             if name not in self.name_map],
+                            self.name_map))
+            bitname = None
```
mcara: ``bitname = None`` not necessary. I think you can do ``elif`` for the next ``if``.

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+
+        # Map the bit positions to bit values
+        if bitpos is not None:
```
mcara: **el**if bitpos is not None:

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+            try:
+                bitval = [_POWERS[pos] for pos in bitpos]
+            except KeyError:
+                raise ValueError(
+                    'The "bit position(s)" {} are not in the allowed range of '
+                    'values.'.format([pos for pos in bitpos
+                                      if pos not in _POWERS]))
+            bitpos = None
```
mcara: ``bitpos = None`` is not necessary

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+
+        # If bitname or bitvals were given these are now converted to bitvalues
+        # so we can do all the real processing in here.
+        if bitval is not None:
```
mcara: This ``if`` is not necessary as ``bitval`` is always not ``None``. Shift the block below to the left.

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+            # There is no zero'th flag so subtract one from the representable
+            # flags to get the real number of flags
+            maximum = _REPRESENTABLE_FLAGS[bitmask.dtype.name] - 1
```
mcara: maximum = 1<< (8 * bitmask.dtype.itemsize)

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+            # Make sure the maximum value doesn't exceed the limit given by the
+            # dtype of the bitmask
+            if max(bitval) > _POWERS[maximum]:
```
mcara:
```python
if max(bitval) > 1<<maximum:
```
Another idea is to filter out/discard larger bit values without even raising an exception (wo cares about these bit values if input bitmask cannot even have such bits? In my opinion we can safely get rid of them):
```python
maximum_bitval = maximum
bitval = [v for v in bitval if v <= 1<< maximum_bitval]
```

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+                raise ValueError(
+                    'The range of values ({}) exceeds the limit of the data '
+                    'type of the bitmask ({}).'
+                    .format(max(bitval), _POWERS[maximum]))
+            uniques = set(bitval)
+            # Make sure only valid powers of two are inside the bitvals
+            if not uniques.issubset(_POWERS):
```
mcara: This will need rethinking if getting rid of ``_POWERS``

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+                raise ValueError(
+                    'The "bitvals" contain illegal (not powers of two) values:'
+                    ' {}'.format(sorted(uniques.difference(_POWERS))))
+
+            # Independant of the flag interpretation we need to sum all
+            # unique bitvalues
+            cmpbitval = sum(uniques)
```
mcara: I think having users to provide ``cmpbitval`` as an input argument is much more useful then, e.g., ``bitname`` (in fact, I can't even imagine what who and when would like to use ``bitname``).

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+
+            # If we want to reverse the flag interpretation we must subtract
+            # the value from the sum of all possible flags:
+            if reverse_flag_interpretation:
+                cmpbitval = np.array(sum(_POWERS[:maximum+1]) - cmpbitval,
```
mcara:
```python
if reverse_flag_interpretation:
    cmpbitval = -(cmpbitval + 1)

File: ccdproc/bitmask.py (hunk 2 above)
Comment (mcara) [inline ccdproc/bitmask.py]: @MSeifert04 Astrodrizzle (if you do not know what it is - do not worry) creates "context" images (see my last comment on https://forum.stsci.edu/discussion/121/astrodrizzle-segmentation-fault-11) that indicate which input image has contributed to a given output pixel. "Context" image is a sort of bitflag mask. Unfortunately, when there are more than 64 input images (in your case, when user provides a ``boolarray`` with ``shape[0]>64``) we are running out of bits that ``numpy.uint64`` can store. Then, we create a 3D array that has multiple 2D "planes" of ``uint64`` and so next plane extends previous plane(s) bits.

File: ccdproc/bitmask.py (hunk 6 above)
Comment (mcara) [inline ccdproc/bitmask.py]: > discards potentially valid flags when they are all False

Maybe I am missing something, but I do not see the point of creating these extra "planes" that we *know* are all ``False``. Compare:
```python
>>> '{:032b}'.format(2992)
'00000000000000000000101110110000'
>>> '{:b}'.format(2992)
'101110110000'
```
Do you feel that you miss/discard all those leading 0s?

ccdproc/bitmask.py
mcara: Eh, now that I think more of it, it seems that my formulation is too simplistic... Maybe it should also take into account signed/unsigned property of input data and then subtract 1 accordingly from maximum.

ccdproc/bitmask.py
mcara: Actually,
```python
if reverse_flag_interpretation:
    cmpbitval = ~cmpbitval
```

mcara:
@MSeifert04 The following is my sincere opinion and I am hope it will not be interpreted as biased because I wrote STScI's bitmask handling: https://github.com/spacetelescope/stsci.tools/blob/master/lib/stsci/tools/bitmask.py

I just want to note two "general" observations:
1. Personally, I am not a fun of OOP just for the sake of it (and I may be lonely here among astropythonists). I do not think that classes are a good fit for *every situation*. Related to this PR, I do not think ``Bitmask`` is really necessary and it seems to be forced. For example, ``Bitmask`` is essentially used to store ``bitmask`` and ``name_map`` and has only two instance methods: ``to_mask`` and ``to_boolarray`` (about this one see next point) and only ``to_mask()`` really needs ``name_map``.
2. I may not know all the reasons behind this code, but it feels like "bloatware". Specifically,
   - I see no use in ``name_map``. This would be cool in a GUI so that users do not have to remember which bit corresponds to a "hot pixel", but in an API I do not see how this may be useful: users (e.g, I) still have to remember the "names" of "bad bits" specific to an instrument.
   - ``to_boolarray()``: I can't imagine what would be a possible justification for the existence of this method. Why would I ever want to do this? Usually, I would like to get a map of "bad pixels" that a "bad" due to a specific issue (or couple of issues) (e.g, "hot pixels" and "dead pixels") and if I want this map for multiple problems - I still would want a single map. That is, I would use ``to_mask`` instead of ``to_boolarray()`. I think this method is an example of bloatware that makes ``astropy`` so hard to navigate: it is so hard to find the method that I need when I need it because I am overwhelmed with so many method (useful and useless).
   - ``from_boolarray()`` may be of limited use, e.g, to create something like drizzlepac's "context image". But this is so rarely needed (specialized), that I wonder if it needs to be at all in the astropy? I cannot see this method to be of interest for "general use".

ccdproc/bitmask.py
```diff
@@ -0,0 +1,473 @@
...
+            num = _REPRESENTABLE_FLAGS[bitmask.dtype.name]
+        arr = np.zeros(tuple(itertools.chain([num], bitmask.shape)),
+                       dtype=bool)
+        # Bitwise and checks if the bitmask contains a set flag
+        # for this exponent (layer of the boolmask).
+        for exp in range(num):
+            arr[exp] = bitmask & _POWERS[exp]
+        return arr
+
+    def to_mask(self, bitpos=None, bitval=None, bitname=None,
```
mcara: If you want to keep all these options ``bitpos``, ``bitval``, and ``bitname``, maybe it would be more useful to have these in ``kwargs``? That way, only one can be specified and the code will raise an exception if two or more of these arguments are present.

ccdproc/bitmask.py
SaraOgaz: @MSeifert04 Sorry I didn't provide some more guidance on all that stuff when I put in the request.  I'll try and take a look at this next week!

MSeifert04:
> Personally, I am not a fun of OOP just for the sake of it (and I may be lonely here among astropythonists). I do not think that classes are a good fit for every situation. Related to this PR, I do not think Bitmask is really necessary and it seems to be forced. 

@mcara I think that's probably the most important point. I wasn't sure if the intention of DQ flag support entailed modifying bitmasks as well as converting it to a normal mask. Currently I agree that this could be simplified a lot by reducing it to one function similar to the one you created in `stsci.tools`. Given the amount of comments "to be adressed" I'll open an alternative PR just containing the bitmask -> mask functionality. That should be pretty easy to do.

pllim:
Re: bitmask and mask -- Doesn't this discussion belong in astropy/astropy-APEs#14 ?

MSeifert04:
@pllim yes. (damn, I totally forgot about that discussion)

mwcraig:
After some discussion offline, we (ccdproc devs) think the quickest way forward is this:

1. Copy the stsci.tools function `bitmask2mask` into ccdproc. Since we are only using that one function copying it makes more sense than adding scsci.tools as a dependency. 

2. Add documentation suggesting that users:
    + Store DQ arrays in the `flags` attribute.
    + Set the mask on a CCDData object using the output of `bitmask2mask`
    + Call `combiner` after setting the mask.

mcara:
@mwcraig I am not against this. I was wondering if being able to specify ``bitval`` as a list (instead of a sum of bitvalues or a string of comma-separated values as curently stsci.tools.bitmask accepts) is something of value to you? What about ``bitpos``? I was thinking if I could add this kind of capacity to the stsci package (maybe before you port it) but I kind of have no desire do anything about ``name_map``.

Also @pllim mentioned (see https://github.com/astropy/ccdproc/pull/458#issuecomment-283056167) a long discussion about bitmask needs for LSST. Maybe in the context of NDData it would make sense to have a Bitmask class (I have no idea how people think of implementing this) but I think it is better to start with a function that is a workhorse and maybe introduce a class later if astropythonist desires for sophistication prevail. That is, in my opinion, if a function such as the one in stsci.tools satisfies current needs of your project or of @SaraOgaz, I am all for starting with a function.

mwcraig:
@mcara -- sounds like we are on the same page. If you were able to add the ability for `bitval` to be a list that would be great. I don't think `bitpos` is a high priority, and the name idea can wait for the broader discussion.

The class @MSeifert04 is one way to go; IIRC, the LLST scheme for masks is somewhat different, and at some point before astropy 2.0 a discussion about how to proceed will need to happen. I'm pretty sure that will involve some sort of bitmask object.

MSeifert04:
Also closing in favor of #460

PR #460, opened 2017-03-02
Bitmask Vol. III

MSeifert04:
Next try to support bitmasks 😄 

This time I simply copied the `stsci.tools.bitmask` module into `ccdproc` (also including their license into the `licenses` folder) and created a thin wrapper around it.

The thin wrapper isn't necessary, it's mostly so the result is a numpy-like mask.


ccdproc/extern/bitmask.py
```diff
@@ -0,0 +1,235 @@
...
+    bitmask : int or None
+        Returns and integer bit mask formed from the input bit value
+        or `None` if input `val` parameter is `None` or an empty string.
+        If input string value was prepended with '~', then returned
+        value will have its bits flipped (inverse mask).
+
+    Examples
+    --------
+        >>> from ccdproc.extern import bitmask
+        >>> "{0:016b}".format(0xFFFF & bitmask.interpret_bits_value(28) )
```
MSeifert04: @mcara I had to change the name of the function here from `interpret_bit_flags` -> `interpret_bits_value`.

SaraOgaz:
Hey @MSeifert04, sorry for the radio silence once again.  I'm kind of swamped the rest of the week, but if I don't get to this by the end of the week, i will BE SURE to test it out beginning of next week.  I should have plenty of good files to try it out with.

ccdproc/extern/bitmask.py
mcara: @MSeifert04 I am working on redesigning stsci.tools.bitmask to add some new features (e.g., support for Python lists of flags) and improved terminology. In particular I do not like the use of "bit value" instead of flags. In my mind, *bit value* is either 0 or 1 and not ``2**n`` (which is a flag).

MSeifert04: you're right. But I was just changing it because the documentation tests couldn't find any "interpret_bit_flags" function and therefore failed.

bsipocz:
@MSeifert04 - Could you rebase to get rid of the merge commit?

ccdproc/extern/bitmask.py
MSeifert04: Thank you 👍 

Could you let me/us know when you think these changes are ready, I'll update this PR then.

mcara:
@MSeifert04 Please take a look at https://github.com/spacetelescope/stsci.tools/pull/29 and also please feel free to comment over there if you would like something changed before I merge that PR.

MSeifert04:
I updated the PR using the newest version of `stsci.tools.bitmask`.

mwcraig:
Haven't had a chance to take a detailed look...I think the code in `extern` should be excluded from the coverage test. I'm assuming that is most of the source of the decrease in coverage. 

Should we add a test or two for the wrapper?

MSeifert04:
@mwcraig I think there are several issues:

- The code from `stsci.tools` is in fact mostly untested (https://github.com/spacetelescope/stsci.tools/pull/29#issuecomment-288150201). 
- I copied the whole file including the deprecated stuff. I probably should remove those.
- I think something is also wrong with coveralls. I can't see the combined coverage anymore (https://coveralls.io/builds/10706733). Do you know if they have an issue tracker?
- The doctests already cover all 99/117 lines in the extern file and most untested lines are errors: https://coveralls.io/jobs/24078036/source_files/748726476. It's not hard to create a few tests that excercise these lines.

mcara:
1. I agree that you should get rid of the deprecated stuff.
2. Even though I did not write regression tests, the code was "verified" to work correctly.

MSeifert04:
@mcara Yeah, the fact that my previous doctests (for the `bitmask2mask`-function) produce identical results shows that there is no regression (at least not in the branches I covered). What I meant was that there are no tests (besides the doctests) to copy for the `bitmask.py` file :)

mwcraig:
If it is covered by doctests then I think it is fine, but let's wait for @crawfordsm to comment.

@mirca -- thanks for the work you have put into this! It is really nice being able to use the functionality from stsci.tools instead of reinventing the wheel. If you would like us to work on/contribute some unit test, let us know. Not implying that they need to be done beyond the doctests, just an offer to pitch in if it would be useful.

mirca:
@mwcraig I think you meant @mcara, haha ;-)

mcara:
@mwcraig I sincerely appreciate your offer to help with unit tests but I would like to do it myself (this should make @pllim happy).

mcara:
> Not implying that they need to be done beyond the doctests, just an offer to pitch in if it would be useful.

I would not have been upset if you actually implied that doctests are not sufficient and, strangely enough (given my opinion about regression tests), I would agree with you _for this particular function_.

ccdproc/core.py
```diff
@@ -1740,6 +1741,94 @@ def _sigma_mask(baseline, one_sigma_value, lower_sigma, upper_sigma):
     return mask
 
 
+def bitmask2mask(bitmask, ignore_bits=0, flip_bits=None):
```
MSeifert04: I think `bitmask` (or `bitflags`) are more commonly used than `bitfield` so I think we should expose it under that name. however does anyone have a better idea for this function?

crawfordsm: bitmask_to_boolean_mask

crawfordsm:
Needs small change to the name and a changelog and then it can be merged!  Thanks @MSeifert04 and @mcara !

ccdproc/core.py
mcara: The thing is that ``ignore_bits`` argument is actually a mask and it is not the thing that gets  converted to a "boolean mask". It is the input DQ array whose elements are "collections of bit flags" (=[bit fields](https://en.wikipedia.org/wiki/Bit_field)).

mcara:
@crawfordsm Thank you!

MSeifert04:
@mcara In response to your comment (https://github.com/astropy/ccdproc/pull/460#discussion_r110966881), what about `apply_bitmask` as function name?

ccdproc/core.py
crawfordsm: @mcara if your preferred name involves bitfields_to_boolean_mask, then I'd be okay with going with that.

mcara: @crawfordsm Yes, in _my_ opinion (after spending significant time "educating" myself on this), ``bitfields_to_boolean_mask`` seems better than the old name that I have used ``bitmask2mask`` (which was technically incorrect IMO), or the other proposed name ``apply_bitmask`` which suggests the operation of "applying" (e.g., ``logical-AND``) a mask: ``dq&mask``.

However, it is exhausting to discuss semantics and moreover it is your project so please feel free to name things in any way that you feel is more intuitive (and/or correct). In my above comment about masks and fields I just wanted to insert what I have gathered from around the web.

mcara: BTW, masks and bitfields look like very similar things except for the role that each one plays in "applying mask": bits in a [mask](https://en.wikipedia.org/wiki/Mask_(computing)) indicate what bits in the bit field (=DQ field/value) should be modified/masked - at least that's my understanding.

MSeifert04:
ok, I changed (or better unchanged) the name to `bitfield_to_boolean_mask`.

mwcraig:
For cross-reference purposes: this helps address #389

Issue #423, opened 2016-11-17
ImageFileCollection seems to fail on multi extension files


boada:
```
In [1]: ccdproc.__version__
Out[1]: '1.1.0'
```

I have a bunch of imaging data where each of the instrument's CCDs have been written to a different extension in the same file: 

```
Filename: image.fits
No.    Name         Type      Cards   Dimensions   Format
0    PRIMARY     PrimaryHDU     179   ()              
1    im1         ImageHDU       179   (2112, 2048)   int32   
2    im2         ImageHDU       179   (2112, 2048)   int32   
3    im3         ImageHDU       179   (2112, 2048)   int32   
4    im4         ImageHDU       179   (2112, 2048)   int32  

```
Running ImageFileCollection on a directory which contains a single image gives:

```
In [17]: ic1 = ImageFileCollection('.', keywords=keys)
    ...: 
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-17-85b0c52a8b28> in <module>()
----> 1 ic1 = ImageFileCollection('.', keywords=keys)

/home/boada/.local/lib/python3.5/site-packages/ccdproc/image_collection.py in __init__(self, location, keywords, info_file)
    101 
    102         if keywords:
--> 103             self.keywords = keywords
    104 
    105     @property

/home/boada/.local/lib/python3.5/site-packages/ccdproc/image_collection.py in keywords(self, keywords)
    198             # Reorder the keywords to match the initial ordering.
    199             new_keys.sort(key=keywords.index)
--> 200             self._summary_info = self._fits_summary(header_keywords=new_keys)
    201 
    202     @property

/home/boada/.local/lib/python3.5/site-packages/ccdproc/image_collection.py in _fits_summary(self, header_keywords)
    431                 continue
    432 
--> 433         summary_table = Table(summary_dict, masked=True)
    434 
    435         for column in summary_table.colnames:

/home/boada/.local/lib/python3.5/site-packages/astropy/table/table.py in __init__(self, data, masked, names, dtype, meta, copy, rows, copy_indices, **kwargs)
    369 
    370         # Finally do the real initialization
--> 371         init_func(data, names, dtype, n_cols, copy)
    372 
    373         # Whatever happens above, the masked property should be set to a boolean

/home/boada/.local/lib/python3.5/site-packages/astropy/table/table.py in _init_from_dict(self, data, names, dtype, n_cols, copy)
    668 
    669         data_list = [data[name] for name in names]
[...]

PR #463, opened 2017-03-08
Add multi-extension support

janga1997:

- [X] For new contributors: Did you add yourself to the "Authors.rst" file?

For new functionality:

- [X] Did you add an entry to the "Changes.rst" file?
- [X] Did you include a meaningful docstring with Parameters, Returns and Examples?
- [X] Did you include tests for the new functionality?
- [X] Does this PR add, rename, move or remove any existing functions or parameters?

**edit of the edit:**
**edit:** ~~Closes 423~~ Turns out that is not a MEF issue after all :)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -36,6 +36,7 @@ Alphabetical list of contributors
 * Brigitta Sipocz (@bsipocz)
 * Connor Stotts (@stottsco)
 * Ole Streicher (@olebole)
+* JVSN Reddy (@janga1997)
 * Erik Tollerud (@eteq)
 * Zè Vinícius (@mirca)
 * Josh Walawender (@joshwalawender)
@@ -45,4 +46,3 @@ Alphabetical list of contributors
 please send an email to the coordinators, or
 `open a pull request for this page <https://github.com/astropy/ccdproc/edit/master/AUTHORS.rst>`_
 in the `ccdproc repository <https://github.com/astropy/ccdproc>`_)
-
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@
 
 New Features
 ^^^^^^^^^^^^
+
 - added feature to specify extension when calling ImageFileCollection [#463]
+
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@
 
 New Features
 ^^^^^^^^^^^^
-
+- added feature to specify extension when calling ImageFileCollection [#463]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,8 +5,7 @@
 New Features
 ^^^^^^^^^^^^
 
-- added feature to specify extension when calling ImageFileCollection [#463]
-
+- Added ext parameter and property to ImageFileCollection to specify the FITS extension.
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -58,6 +58,11 @@ class ImageFileCollection(object):
         The filenames are assumed to be in ``location``.
         Default is ``None``.
 
+     hdu: str or int, optional
+         Name/Index of the extension, which specifies the extension to extract data
+         from all files in a given directory.
+         Default is 0.
+
     Raises
     ------
     ValueError
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -58,7 +58,7 @@ class ImageFileCollection(object):
         The filenames are assumed to be in ``location``.
         Default is ``None``.
 
-     hdu: str or int, optional
+     ext: str or int, optional
          Name/Index of the extension, which specifies the extension to extract data
          from all files in a given directory.
          Default is 0.
@@ -70,7 +70,7 @@ class ImageFileCollection(object):
         value.
     """
     def __init__(self, location=None, keywords=None, info_file=None,
-                 filenames=None, hdu=0):
+                 filenames=None, ext=0):
         self._location = location
         self._filenames = filenames
         self._files = []
@@ -114,7 +114,7 @@ class ImageFileCollection(object):
         # is always *some* value.
         self._all_keywords = False
 
-        self._hdu = hdu
+        self._ext = ext
 
         if keywords:
             self.keywords = keywords
@@ -222,6 +222,13 @@ class ImageFileCollection(object):
         list of str, Unfiltered list of FITS files in location.
         """
         return self._files
+
+    @property
+    def ext(self):
+        """
+        str,The extension for which all 
+        """
+        return self._ext
 
     def values(self, keyword, unique=False):
         """
@@ -370,7 +377,7 @@ class ImageFileCollection(object):
             summary = input_summary
             n_previous = len(summary['file'])
 
-        h = fits.getheader(file_name, self._hdu)
+        h = fits.getheader(file_name, self._ext)
 
         assert 'file' not in h
 
@@ -702,13 +709,13 @@ class ImageFileCollection(object):
 
             file_name = path.basename(full_path)
 
-            hdu_index = hdulist.index_of(self._hdu)
-            ccd_kwargs.setdefault('hdu', hdu_index)
+            ext_index = hdulist.index_of(self._ext)
+            ccd_kwargs.setdefault('ext', ext_index)
 
             return_options = {
-                    'header': lambda: hdulist[hdu_index].header,
-                    'hdu': lambda: hdulist[hdu_index],
-                    'data': lambda: hdulist[hdu_index].data,
+                    'header': lambda: hdulist[ext_index].header,
+                    'hdu': lambda: hdulist[ext_index],
+                    'data': lambda: hdulist[ext_index].data,
                     'ccd': lambda: fits_ccddata_reader(full_path, **ccd_kwargs)
                     }
             try:
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -264,7 +264,8 @@ class ImageFileCollection(object):
     @property
     def ext(self):
         """
-        str or int, The extension from which the header and data will be read in all files.
+        str or int, The extension from which the header and data will
+        be read in all files.
         """
         return self._ext
 
@@ -769,13 +770,14 @@ class ImageFileCollection(object):
             file_name = path.basename(full_path)
 
             ext_index = hdulist.index_of(self._ext)
-            ccd_kwargs.setdefault('ext', ext_index)
 
             return_options = {
                     'header': lambda: hdulist[ext_index].header,
                     'hdu': lambda: hdulist[ext_index],
                     'data': lambda: hdulist[ext_index].data,
-                    'ccd': lambda: fits_ccddata_reader(full_path, **ccd_kwargs)
+                    'ccd': lambda: fits_ccddata_reader(full_path,
+                                                       hdu=ext_index,
+                                                       **ccd_kwargs)
                     }
             try:
                 yield (return_options[return_type]()  # pragma: no branch
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -65,7 +65,7 @@ class ImageFileCollection(object):
         value.
     """
     def __init__(self, location=None, keywords=None, info_file=None,
-                 filenames=None):
+                 filenames=None, extension=None):
         self._location = location
         self._filenames = filenames
         self._files = []
@@ -108,6 +108,11 @@ class ImageFileCollection(object):
         # actually setting the correct value, this just ensure that there
         # is always *some* value.
         self._all_keywords = False
+
+        if extension is None:
+            self._extension = 0
+        else:
+            self._extension = extension
 
         if keywords:
             self.keywords = keywords
@@ -363,7 +368,8 @@ class ImageFileCollection(object):
             summary = input_summary
             n_previous = len(summary['file'])
 
-        h = fits.getheader(file_name)
+        h = fits.getheader(file_name, self._extension)
+
         assert 'file' not in h
 
         # Try opening header before this so that file name is only added if
@@ -694,10 +700,12 @@ class ImageFileCollection(object):
 
             file_name = path.basename(full_path)
 
+            extension_index = hdulist.index_of(self._extension)
+
             return_options = {
-                    'header': lambda: hdulist[0].header,
-                    'hdu': lambda: hdulist[0],
-                    'data': lambda: hdulist[0].data,
+                    'header': lambda: hdulist[extension_index].header,
+                    'hdu': lambda: hdulist[extension_index],
+                    'data': lambda: hdulist[extension_index].data,
                     'ccd': lambda: fits_ccddata_reader(full_path, **ccd_kwargs)
                     }
             try:
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -59,9 +59,8 @@ class ImageFileCollection(object):
         Default is ``None``.
 
      ext: str or int, optional
-         Name/Index of the extension, which specifies the extension to extract data
-         from all files in a given directory.
-         Default is 0.
+         The extension from which the header and data will be read in all files.
+         Default is ``0``.
 
     Raises
     ------
@@ -226,7 +225,7 @@ class ImageFileCollection(object):
     @property
     def ext(self):
         """
-        str,The extension for which all 
+        str or int, The extension from which the header and data will be read in all files.
         """
         return self._ext
 
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -65,7 +65,7 @@ class ImageFileCollection(object):
         value.
     """
     def __init__(self, location=None, keywords=None, info_file=None,
-                 filenames=None, extension=None):
+                 filenames=None, hdu=0):
         self._location = location
         self._filenames = filenames
         self._files = []
@@ -109,10 +109,7 @@ class ImageFileCollection(object):
         # is always *some* value.
         self._all_keywords = False
 
-        if extension is None:
-            self._extension = 0
-        else:
-            self._extension = extension
+        self._hdu = hdu
 
         if keywords:
             self.keywords = keywords
@@ -368,7 +365,7 @@ class ImageFileCollection(object):
             summary = input_summary
             n_previous = len(summary['file'])
 
-        h = fits.getheader(file_name, self._extension)
+        h = fits.getheader(file_name, self._hdu)
 
         assert 'file' not in h
 
@@ -700,12 +697,13 @@ class ImageFileCollection(object):
 
             file_name = path.basename(full_path)
 
-            extension_index = hdulist.index_of(self._extension)
+            hdu_index = hdulist.index_of(self._hdu)
+            ccd_kwargs.setdefault('hdu', hdu_index)
 
             return_options = {
-                    'header': lambda: hdulist[extension_index].header,
-                    'hdu': lambda: hdulist[extension_index],
-                    'data': lambda: hdulist[extension_index].data,
+                    'header': lambda: hdulist[hdu_index].header,
+                    'hdu': lambda: hdulist[hdu_index],
+                    'data': lambda: hdulist[hdu_index].data,
                     'ccd': lambda: fits_ccddata_reader(full_path, **ccd_kwargs)
                     }
             try:
--- a/ccdproc/tests/test_image_collection.py
+++ b/ccdproc/tests/test_image_collection.py
@@ -155,10 +155,10 @@ class TestImageFileCollection(object):
 
         hdulist.writeto(os.path.join(triage_setup.test_dir, 'multi-extension.fits'))
         ic2 = image_collection.ImageFileCollection(triage_setup.test_dir, keywords='*',
-            filenames=['multi-extension.fits'], hdu='MASK')
+            filenames=['multi-extension.fits'], ext='MASK')
 
         ic1 = image_collection.ImageFileCollection(triage_setup.test_dir, keywords='*',
-            filenames=['multi-extension.fits'], hdu=0)
+            filenames=['multi-extension.fits'], ext=0)
 
         column2 = ic2.summary_info.colnames
         column1 = ic1.summary_info.colnames
--- a/ccdproc/tests/test_image_collection.py
+++ b/ccdproc/tests/test_image_collection.py
@@ -148,6 +148,26 @@ class TestImageFileCollection(object):
         new_data = np.array(collection.summary_info)
         assert (new_data == old_data).all()
 
+    def test_multiple_extensions(self, triage_setup):
+        ext1 = fits.PrimaryHDU()
+        ext2 = fits.ImageHDU(name='MASK')
+        hdulist = fits.hdu.hdulist.HDUList([ext1, ext2])
+
+        hdulist.writeto(os.path.join(triage_setup.test_dir, 'multi-extension.fits'))
+        ic2 = image_collection.ImageFileCollection(triage_setup.test_dir, keywords='*',
+            filenames=['multi-extension.fits'], hdu='MASK')
+
+        ic1 = image_collection.ImageFileCollection(triage_setup.test_dir, keywords='*',
+            filenames=['multi-extension.fits'], hdu=0)
+
+        column2 = ic2.summary_info.colnames
+        column1 = ic1.summary_info.colnames
+
+        list1 = [key.lower() for key in ext2.header]
+        list2 = ic2.summary_info.colnames[1:]
+
+        assert list1 == list2
+
     def test_headers(self, triage_setup):
         collection = image_collection.ImageFileCollection(location=triage_setup.test_dir,
                                              keywords=['imagetyp'])
--- a/ccdproc/tests/test_image_collection.py
+++ b/ccdproc/tests/test_image_collection.py
@@ -155,26 +155,29 @@ class TestImageFileCollection(object):
         new_data = np.array(collection.summary)
         assert (new_data == old_data).all()
 
-    def test_multiple_extensions(self, triage_setup):
+    @pytest.mark.parametrize('extension', ['TESTEXT', 1, ('TESTEXT', 1)])
+    def test_multiple_extensions(self, triage_setup, extension):
         ext1 = fits.PrimaryHDU()
-        ext1.data = np.arange(1,5)
+        ext1.data = np.arange(1, 5)
         # It is important than the name used for this test extension
         # NOT be MASK or UNCERT because both are treated in a special
         # way by the FITS reader.
         test_ext_name = 'TESTEXT'
         ext2 = fits.ImageHDU(name=test_ext_name)
-        ext2.data = np.arange(6,10)
+        ext2.data = np.arange(6, 10)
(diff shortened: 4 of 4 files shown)


janga1997:
@mwcraig

ccdproc/image_collection.py
```diff
@@ -112,6 +112,9 @@ def __init__(self, location=None, keywords=None, info_file=None,
         if keywords:
             self.keywords = keywords
 
+        if extension:
```
mwcraig: This should change to `if extension is not None` because `0` is a valid extension number but will evaluate to `False`.

ccdproc/image_collection.py
```diff
@@ -694,10 +697,12 @@ def _generator(self, return_type,
...
 
+            extension_index = hdulist.index_of(self._extension)
+
             return_options = {
-                    'header': lambda: hdulist[0].header,
-                    'hdu': lambda: hdulist[0],
-                    'data': lambda: hdulist[0].data,
+                    'header': lambda: hdulist[extension_index].header,
+                    'hdu': lambda: hdulist[extension_index],
+                    'data': lambda: hdulist[extension_index].data,
```
mwcraig: This looks like the correct change for the generator.

janga1997:
@mwcraig Please don't mind the commit messages for the review changes. I will squash them later when the PR is ready to merge.

bsipocz:
@janga1997 - Please remove the ``[skip ci]`` for the title, it will only have effect once the PR is merged (and thus skipping CI on the merge and consecutive cron builds). Add it to the individual commit messages instead when you want to skip ci.

vrooje:
Yeah, I think the answer may be "both", but then I'm pretty new to this data myself and haven't fully explored the headers in the other extensions yet...

janga1997:
@mwcraig I think the problem arises in the `_dict_from_fits_header` function, although I can't figure out what's exactly going wrong in that?

mwcraig:
@janga1997 -- taking a look, may still be ~30-45 min until I have a response...

mwcraig:
As a next step I would try printing out (or run it in ipython and drop into the debugger) the contents of the dict returned by `_dict_from_fits_header`. It looks, from #423, like the error is thrown because one of the entries in the dictionary has a different length than the others. That *shouldn't* happen, because missing values are supposed to be filled in with a marker, but apparently it is.

More broadly, I think an `extension` column needs to be added to the table in a way similar to that of  `file`, i.e. always required. For single-extension FITS files it won't make a difference, but for multi-extension FITS (MEFs), it will. Right now there is no way (I think) to indicate which header a keyword in the summary table came from.

ccdproc/image_collection.py
```diff
@@ -112,6 +112,9 @@ def __init__(self, location=None, keywords=None, info_file=None,
         if keywords:
             self.keywords = keywords
 
+        if extension is not None:
```
MSeifert04: That should probably be replaced with:

```
if extension is None:
    self._extension = 0
else:
    self._extension = extension
```

That should fix the failing tests (I hope).

janga1997: You mean the errors on an array being used as a bool?
I was wondering what to do about that. Thanks!

MSeifert04: ok, which error about an array being used? My comment was about fixing all these `AttributeError: 'ImageFileCollection' object has no attribute '_extension'` exceptions 😄

janga1997: I am sorry. I didn't actually test what you had written there.
I was getting some other weird errors when testing locally. :-)

mwcraig:
To keep this PR moving along, I would suggest it be limited to doing a single thing: allowing the user to specify one extension (by name or number) when creating an `ImageFileCollection`. The `summary` table will be built using the keywords from that extension, and the generator will return the data, etc, from that extension.

This would become a valid init: `ic = ImageFileCollection('.', extension=2)`. Then `ic` would build its table using the information in extension 2, and would loop over extension 2 in the `_generator`.

That will not close #423 (or #464) but is a first step towards adding support for multi-extension FITS.

 @janga1997, please post here if you have questions about what I mean. We'll need to generate some small MEF files for testing.

ccdproc/image_collection.py
janga1997: @MSeifert04 Unfortunately the error still persists.
Any reason you can think of?

janga1997: Never mind @MSeifert04 . I really have a bad habit of speaking too soon.

janga1997:
Do I need to add tests for this issue?

ccdproc/image_collection.py
```diff
@@ -694,10 +700,12 @@ def _generator(self, return_type,
...
+            extension_index = hdulist.index_of(self._extension)
+
             return_options = {
-                    'header': lambda: hdulist[0].header,
-                    'hdu': lambda: hdulist[0],
-                    'data': lambda: hdulist[0].data,
+                    'header': lambda: hdulist[extension_index].header,
+                    'hdu': lambda: hdulist[extension_index],
+                    'data': lambda: hdulist[extension_index].data,
                     'ccd': lambda: fits_ccddata_reader(full_path, **ccd_kwargs)
```
MSeifert04: Even though it's more complicated here I think it shouldn't be silently ignored in the `'ccd'`-case. What makes it complicated is that it's also possible to explicitly define a `hdu` by passing it as item-pair in `ccd_kwargs`.

MSeifert04:
@janga1997 Yes, it does need a changelog entry and tests. No complicated tests, just one or two to check if it's working. It also needs an update of the class docstring to explain the new parameter. You can also add yourself as contributor.

ccdproc/image_collection.py
janga1997: Any comments on what could be possibly done about `ccd`?

MSeifert04: I'd say if `ccd_kwargs` contains a `hdu`-keyword that should be used, otherwise one could add `hdu: extension_index` to the `ccd_kwargs`.

janga1997:
@MSeifert04 I've added a test. If this looks good, I will make an addition to changelog.

ccdproc/image_collection.py
```diff
@@ -686,6 +692,8 @@ def _generator(self, return_type,
             self._find_keywords_by_values(**kwd)
 
         ccd_kwargs = ccd_kwargs or {}
+        if 'hdu' not in ccd_kwargs.keys():
+            ccd_kwargs['hdu'] = extension_index
```
MSeifert04: you could also do `ccd_kwargs.setdefault('hdu', extension_index)` instead of checking if it's in the dictionary and adding it manually.

However this needs to be after line 705 (the one which defines the `extension_index`) otherwise you'll get a `NameError`.

ccdproc/tests/test_image_collection.py
```diff
@@ -148,6 +148,19 @@ def test_hdus_masking(self, triage_setup):
...
 
+    def test_multiple_extensions(self, triage_setup):
+        ext1 = fits.PrimaryHDU()
+        ext2 = fits.ImageHDU(name = 'MASK')
+        hdulist = fits.hdu.hdulist.HDUList([ext1, ext2])
+
+        hdulist.writeto(os.path.join(triage_setup.test_dir, 'multi-extension.fits'))
+        ic = image_collection.ImageFileCollection(triage_setup.test_dir, keywords='*',
+                                    filenames = ['multi-extension.fits'], extension='MASK')
+        array1 = list(map(str.lower, list(ext2.header)))
```
MSeifert04: nitpick: Lists are not arrays!

Also I personally find comprehensions more readable, especially in tests where performance isn't critical: `[key.lower() for key in ext2.header]`

ccdproc/tests/test_image_collection.py
```diff
@@ -148,6 +148,19 @@ def test_hdus_masking(self, triage_setup):
...
+        array2 = ic.summary_info.colnames[1:]
+
+        assert np.array_equal(array1, array2)
```
MSeifert04: why not just `list1 == list2`?

janga1997: That is better in this case, Yes. Thanks again!

ccdproc/tests/test_image_collection.py
janga1997: Definitely not a nitpick. I appreciate any and all input. :-)
Changing it now.

janga1997:
@MSeifert04 I don't suppose these errors are in any way related to this PR?
BTW, do you have any further issues with this PR?

MSeifert04:
@janga1997 I think you need to rebase your PR in order to make the tests pass. The basic way to rebase is explained in [the astropy development workflow](http://docs.astropy.org/en/stable/development/workflow/development_workflow.html#rebase) and should be pretty easy because it shouldn't conflict. Let us know if you encounter any problems.

MSeifert04:
Regarding the TODOs, there is a list in the top post that roughly shows the steps involved:

> For new contributors: Did you add yourself to the "Authors.rst" file?
> For new functionality:
> Did you add an entry to the "Changes.rst" file?
> Did you include a meaningful docstring with Parameters, Returns and Examples?
> Does the commit message include a "Fixes #issue_number" (replace "issue_number").
> Did you include tests for the new functionality?

janga1997:
Sure. I wanted to get to the mundane part of the PR, if you are satisfied with the code part of it. I will get to it.

ccdproc/image_collection.py
```diff
@@ -65,7 +65,7 @@ class ImageFileCollection(object):
         value.
     """
     def __init__(self, location=None, keywords=None, info_file=None,
-                 filenames=None):
+                 filenames=None, extension=None):
```
MSeifert04: Not sure but how about naming it `ext` or `hdu`? At least these are the names for similar parameters in `fits.getheader`, etc. I'd say we should go with `hdu` because that's what is used in `CCDData.read`

Also needs a parameter documentation

ccdproc/image_collection.py
```diff
@@ -109,6 +109,11 @@ def __init__(self, location=None, keywords=None, info_file=None,
         # is always *some* value.
         self._all_keywords = False
 
+        if extension is None:
```
MSeifert04: I'm not sure but you could also set the default value to `0`. That should have the same effect as that `if ... else ...`, right?

ccdproc/tests/test_image_collection.py
```diff
@@ -148,6 +148,19 @@ def test_hdus_masking(self, triage_setup):
         new_data = np.array(collection.summary_info)
         assert (new_data == old_data).all()
 
+    def test_multiple_extensions(self, triage_setup):
+        ext1 = fits.PrimaryHDU()
+        ext2 = fits.ImageHDU(name = 'MASK')
```
MSeifert04: Generally one doesn't use spaces around the `=` when it's a keyword parameter.

ccdproc/tests/test_image_collection.py
```diff
@@ -148,6 +148,19 @@ def test_hdus_masking(self, triage_setup):
...
+        hdulist = fits.hdu.hdulist.HDUList([ext1, ext2])
+
+        hdulist.writeto(os.path.join(triage_setup.test_dir, 'multi-extension.fits'))
+        ic = image_collection.ImageFileCollection(triage_setup.test_dir, keywords='*',
+                                    filenames = ['multi-extension.fits'], extension='MASK')
+        list1 = [key.lower() for key in ext2.header]
+        list2 = ic.summary_info.colnames[1:]
+
+        assert list1 == list2
```
MSeifert04: Maybe this could also check that the `ImageFileCollection` of `ext=0` is different from the one in `ext='MASK'`.

ccdproc/tests/test_image_collection.py
```diff
@@ -148,6 +148,19 @@ def test_hdus_masking(self, triage_setup):
...
         assert (new_data == old_data).all()
 
+    def test_multiple_extensions(self, triage_setup):
+        ext1 = fits.PrimaryHDU()
+        ext2 = fits.ImageHDU(name = 'MASK')
+        hdulist = fits.hdu.hdulist.HDUList([ext1, ext2])
+
+        hdulist.writeto(os.path.join(triage_setup.test_dir, 'multi-extension.fits'))
+        ic = image_collection.ImageFileCollection(triage_setup.test_dir, keywords='*',
+                                    filenames = ['multi-extension.fits'], extension='MASK')
```
MSeifert04: Again no spaces around `=` if it's a keyword parameter.

The indentation is also slightly off. The indentation should match the indentation of the previous line or you could use hanging indentation:

```
ic = ...(
    triage_setup.test_dir, 
    keywords='*', 
    filenames=['multi-extension.fits'], 
    extension='MASK')
```

MSeifert04:
> Sure. I wanted to get to the mundane part of the PR, if you are satisfied with the code part of it.

While I agree on the changelog part (because that could easily lead to merge conflicts) the rebase and parameter description are important and not really mundane parts, undocumented or poorly documented parameters are not much better than not having the parameter at all.

Overall this looks fine, but I had some additional comments.

janga1997:
> Something that could be worth investigating is if the writeto-parts of the ImageFileCollection work correctly with the extensions.

@MSeifert04 
I'm not sure I understand.

MSeifert04:
>> Something that could be worth investigating is if the writeto-parts of the ImageFileCollection work correctly with the extensions.

I'm not sure I understand. I've updated according to the rest of your suggestions.

Besides reading the files one can also specify "clobber" or "overwrite" to write the changes back after modifying the data.

Also it looks like something went wrong during rebase, there shouldn't be a merge commit. Do you know what you did so we can help you clean that up? :)

janga1997:
@MSeifert04 I didn't all the commits that came along during the merge. 
Give me 5 minutes

MSeifert04:
No that's not the problem, the problem is that your pull request now contains merge and unrelated commits. To clean these up it would be helpful to know what you did :)

janga1997:
@MSeifert04 But I already merged the master branch into my branch, and squashed the commits!
Would that be a problem?

MSeifert04:
> Would that be a problem?

It would rewrite or redo parts of the history of the master-branch which is generally not advisable because others may have pull requests that may go out of sync (unmergeable) when the pull request would be merged.

Generally you should rebase your PR on master using `git rebase projectrepo/master yourbranch`. That rewrites the history of _your_ branch but keeps the history of the master branch clean.

Probably the best way would be to undo the merge and simply rebase. Did you create a backup-branch before doing the merge and have you read through the link I've posted in https://github.com/astropy/ccdproc/pull/463#issuecomment-286917412?

MSeifert04:
@janga1997 Just to make sure: Do you use command line with git or some other (G)UI?

janga1997:
@MSeifert04 Command line git.

MSeifert04:
This is not a perfect solution, however you probably can restore to a clean state. Important: Make a temporary copy of your branch before you start with those commands.

make sure your branch is in sync with the remote branch:

```
git fetch --all  # get all remote branches
git checkout multi-extension  # switch to your feature branch
git reset --hard {yourrepo}/multi-extension  # undo all local changes compared to the remote branch
```

Then undo the last commit and rebase (no need to squash):

```
git reset --hard HEAD~1  # undoes the merge/squash commit. That also undoes your last changes!
git rebase {ccdprocrepo}/master test  # rebase on master!
```

Then _before_ you push it to the remote branch again you need to include all _your_ changes in the previously [last commit](https://github.com/astropy/ccdproc/pull/463/commits/446408c4f6ff8154a1c4dde9a0d0b6af4494c304). Then force push to the remote branch.

janga1997:
@MSeifert04 I thought it was much more easier to just reset that commit that merged master, and write down my changes (which were simple docs one liners), and push again.

janga1997:
Done.

AUTHORS.rst
```diff
@@ -45,4 +46,3 @@ Alphabetical list of contributors
 please send an email to the coordinators, or
 `open a pull request for this page <https://github.com/astropy/ccdproc/edit/master/AUTHORS.rst>`_
 in the `ccdproc repository <https://github.com/astropy/ccdproc>`_)
-
```
MSeifert04: Any specific reason to remove this blank line?

MSeifert04: Why remove this line?

CHANGES.rst
```diff
@@ -4,7 +4,7 @@
 
 New Features
 ^^^^^^^^^^^^
-
```
MSeifert04: here as well: No need to remove this blank line.

ccdproc/image_collection.py
```diff
@@ -58,14 +58,19 @@ class ImageFileCollection(object):
         The filenames are assumed to be in ``location``.
         Default is ``None``.
 
+     hdu: str or int, optional
+         Name/Index of the extension, which specifies the extension to extract data
+         from all files in a given directory.
+         Default is 0.
+
```
MSeifert04: Could you add a `.. versionadded: 1.3` here? Needs a newline before and after and needs to have the same indentation as the descritpion.

mwcraig:
> There are some things that would be "nice to have" but not mandatory because they could be really tricky:
...
> exposing hdu as property.
>exposing hdu in the summary table.

For now let's expose it as a (read-only) property (once it is renamed to `ext`) but not worry about the summary table yet, since, like the location, the extension is the same for all of the files for now.

ccdproc/image_collection.py
```diff
@@ -58,14 +58,19 @@ class ImageFileCollection(object):
...
     Raises
     ------
     ValueError
         Raised if keywords are set to a combination of '*' and any other
         value.
     """
     def __init__(self, location=None, keywords=None, info_file=None,
-                 filenames=None):
+                 filenames=None, hdu=0):
```
MSeifert04: Oh that one is one me. I recommended to use `hdu` because we use `hdu` in [`CCDData.read`](http://ccdproc.readthedocs.io/en/latest/api/ccdproc.fits_ccddata_reader.html#ccdproc.fits_ccddata_reader). I'm not opposed to `ext` but should we change it (not in this PR!) in the reader as well?

mwcraig: 😳 Guess who chose to call it `hdu` in `CCDData.read`...that would be me 😱 

I guess in the context of `ImageFileCollection` I was thinking that `hdu` was unclear because there are multiple files, there is an `hdus` iterator, and so I immediately wondered which `hdu` of the `hdus` in the collection `ic.hdu` would refer to.

Using `ext` would avoid that potential confusion...though it is entirely possible I'm the source of confusion.

AUTHORS.rst
janga1997: That must be Atom being weird. I have faced one too many issues with empty newlines.

ccdproc/image_collection.py
```diff
@@ -216,6 +223,13 @@ def files(self):
         """
         return self._files
 
+    @property
+    def ext(self):
+        """
+        str,The extension for which all
```
mwcraig: Really minor nitpick: add a space between the comma and `The`

mwcraig: Also, this looks like an incomplete sentence -- maybe `for which all` → `from which the header and data will be read in all files.`?

MSeifert04: And the type doesn't really match what can be used. As far as I know it can be `str` but `int` or a `str, int` tuple.

MSeifert04: The property also needs a test. Should be easily included in the existing test.

CHANGES.rst
```diff
@@ -5,6 +5,8 @@
 New Features
 ^^^^^^^^^^^^
 
+- added feature to specify extension when calling ImageFileCollection [#463]
```
MSeifert04: Maybe this could be specified a bit more: "Added `ext` parameter and property to ImageFileCollection to specify the extension."

ccdproc/image_collection.py
```diff
@@ -58,14 +58,19 @@ class ImageFileCollection(object):
         The filenames are assumed to be in ``location``.
         Default is ``None``.
 
+     ext: str or int, optional
+         Name/Index of the extension, which specifies the extension to extract data
+         from all files in a given directory.
+         Default is 0.
```
MSeifert04: Could you add double backticks around the ``0``.

ccdproc/image_collection.py
```diff
@@ -58,14 +58,19 @@ class ImageFileCollection(object):
         The filenames are assumed to be in ``location``.
         Default is ``None``.
 
+     ext: str or int, optional
+         Name/Index of the extension, which specifies the extension to extract data
```
MSeifert04: The first part contains some redundancy "[...]the extension which specifies the extension[...]". Also "to extract fata from all files in a given directory" is not really specific for the parameter. The `ext` parameter specifies the extension from which to extract the header/data, right?

janga1997: Actually, I fell asleep while writing that, and when I woke up, I forgot and pushed it anyway :)
Correcting it now.

ccdproc/image_collection.py
janga1997: Can it be a tuple? This PR works only for single extensions.

MSeifert04: Not sure, I thought so. You could test it 😄

janga1997: I don't think I accounted for that. The requirement in the original issue was that a single extension would be extracted from all files.
So I don't know if any iterables like a tuple would work.

AUTHORS.rst
MSeifert04: So no possibility to add it again?

CHANGES.rst
```diff
@@ -5,6 +5,7 @@
 New Features
 ^^^^^^^^^^^^
 
+- Added ext parameter and property to ImageFileCollection to specify the FITS extension.
```
MSeifert04: Even though it's not enforced I try to keep lines (if possible also in `rst` files) below 80 characters. Also you lost the pull request number 😄

ccdproc/image_collection.py
```diff
@@ -58,14 +58,18 @@ class ImageFileCollection(object):
         The filenames are assumed to be in ``location``.
         Default is ``None``.
 
+     ext: str or int, optional
+         The extension from which the header and data will be read in all files.
```
MSeifert04: I think you could drop the "in all files" part (simililarly in the property description). However maybe someone else wants to chime in here because I'm not sure if that part clarifies or confuses more.

ccdproc/image_collection.py
```diff
@@ -216,6 +222,13 @@ def files(self):
         """
         return self._files
 
+    @property
+    def ext(self):
+        """
+        str or int, The extension from which the header and data will be read in all files.
```
MSeifert04: Could be split into two lines (because > 80 characters). Also the "in all files" part that I mentioned in the parameter description.

And did you check if one can pass in a tuple of `(str, int)`?

MSeifert04: I suspect that the `tuple` works because it works for [`fits.getheader`](http://docs.astropy.org/en/stable/io/fits/api/files.html#astropy.io.fits.getdata). But I might be wrong.

AUTHORS.rst
janga1997: The line shows up in my editor, but doesn't in the git diff. I will add a newline anyway for good measure then.

MSeifert04:
The inline comments are mostly stilistic, however I think this needs to be tested a bit more thoroughly:

- one test for `headers`, `hdus`, `ccds` and `data` with `ext != 0`. Making sure the `_generator` works with extensions.
- one test using the generator (`ext != 0`), modifying the return value and using `overwrite=True`. Making sure the `_generator` can handle modifying/overwriting one extension but leaving the other extensions alone.

ccdproc/image_collection.py
janga1997: It does work, but only if the str and the int value correspond to the same extension.
and only in the order `(str, int)`

MSeifert04: That's what I said in https://github.com/astropy/ccdproc/pull/463#discussion_r106825591, right? However the important question is: does it also work for the `_generator`? If not we don't need to bother advertising it, but if it works it should be included.

janga1997:
@MSeifert04 This was one area where I wasn't really  sure. I think I have to read a bit about what generator methods are doing here.

MSeifert04:
@janga1997 No rush, just let me know if you need help. You can probably borrow a lot from some existing tests :)

janga1997:
@MSeifert04 I am not sure about this, but maybe this is the way to write those tests
```
        collection = image_collection.ImageFileCollection(
                location=triage_setup.test_dir, keywords='*', filename = 'multi-extension.fits')
        ccd_kwargs = {'ext': 1}
        for ccd in collection.ccds(ccd_kwargs=ccd_kwargs):
            assert isinstance(ccd, CCDData)

```

MSeifert04:
Yes! Instead of the `isinstance` something testing that it actually contains the right data/meta would be better.

janga1997:
@MSeifert04 I am not really sure about the need for these tests. Right now, we can specify only one extension.
Testing for values only in the case of 1 extension seems kind of circular to me. I could definitely be wrong.

MSeifert04:
Well, it's kind of important that one doesn't load the **wrong** extension with the generators.

janga1997:
@MSeifert04 Well, I am not sure as to how to check that.
For example, what do I compare `collection.ccds(ccd_kwargs=ccd_kwargs)` against?

Also, I am not sure how a to test for a possible **wrong** extension, if there is only one extension. I surely see that test becoming necessary in the future, where we can specify a list of extensions to load, and want to load the `ccds` for a specific extension. What am I missing?

MSeifert04:
Ah I think I understand it now: The generators like `ccds` don't yield different extensions for one file, they will yield one item for each file. So if you have multiple files it will give you each file as `CCDData`. So the easiest way to compare would be something like:

```
# create a file containing several extensions, fill them with different data
# create a filecollection for that file and a not-0 extension

# Make sure the generators give appropriate results
for data, hdr, hdu, ccd in zip(ic.data(), ic.headers(), ic.hdus(), ic.ccds()):
    # make sure these are identical to the extension you created
    np.testing.assert_array_equal(data, predefined_data_for_extension_xyz)
    assert hdr == predefined_header_for_extension_xyz
    # Now compare that the generators each give the same stuff
    np.testing.assert_array_equal(data, ccd.data)
    np.testing.assert_array_equal(data, hdu.data)
    assert hdr == hdu.header
    assert hdr == ccd.meta
```

janga1997:
@MSeifert04 There is just one single problem. `ic.ccds()` seems to always raise the erro `ValueError: a unit for CCDData must be specified.` 
Any possible reason I'm running into this?

MSeifert04:
You need to specify a `unit` for `CCDData`, so you need to pass in `ccd_kwargs` containing a unit. Doesn't matter which one (I think we mostly use `'adu'`).

MSeifert04:
@janga1997 Do you need more help for the tests?

janga1997:
@MSeifert04 Sorry for the LONG delay. Just got finished with my exams today.

About the tests, everything else except `ccds()` seems to be fine.
the `ccds()` generator seems to only generate data from the first extension.

I think it surely has something to with how the generator generates `ccds()`

```
            return_options = {
                    'header': lambda: hdulist[ext_index].header,
                    'hdu': lambda: hdulist[ext_index],
                    'data': lambda: hdulist[ext_index].data,
                    'ccd': lambda: fits_ccddata_reader(full_path, **ccd_kwargs)
                    }
```
Does `fits_ccddata_reader` have a provision for different extensions?

ccdproc/image_collection.py
mwcraig: Let's leave it in to emphasize that the data will be read from the same extension for all of the files.

mwcraig:
@MSeifert04 -- can you please take another look at this and merge if it is ok?

MSeifert04:
I'm travelling but I will take a look at the weekend. But I think the only remaining issue were the tests and we could also include them later.

mwcraig:
@MSeifert04 -- friendly bump -- feel free to merge if this looks ok now.

MSeifert04:
Thanks @janga1997 :)

PR #465, opened 2017-03-11
Handle duplicate keywords in Headers

### LINKED ISSUE #464: Handle duplicate keywords in headers in ImageFileCollection
What happens:

If a fits file that has the same keyword more than once in the header is added to an `ImageFileCollection`, building the collection fails as reported in #423. @boada was on the right track in his comment https://github.com/astropy/ccdproc/issues/423#issuecomment-261349459, which illustrates that the reported issue has to do with the contents of the headers, not with the fact that the files are multi-extension. This [example file from @vrooje](https://github.com/astropy/ccdproc/files/825931/b207_os_bs_ff_cr.fits.zip) demonstrates the issue. 

In that file, the keyword `OBSTYPE` appears twice in the primary header, once with value `'OBJECT'` and again with value `1`. Base on @boada's comment https://github.com/astropy/ccdproc/issues/423#issuecomment-261351015 I'm guessing the same issue occurred there, too.

Not sure yet what the expected behavior should be in this case. 😕 

Able to reproduce with astropy 1.3, ccdproc v1.2.0.

janga1997:
Fixes #464

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -351,7 +351,13 @@ class ImageFileCollection(object):
         def _add_val_to_dict(key, value, tbl_dict, n_previous, missing_marker):
             key = key.lower()
             try:
-                tbl_dict[key].append(value)
+                if key not in tbl_dict.keys():
+                    tbl_dict[key].append(value)
+                else:
+                    warnings.warn('The keyword "' + key.upper() +
+                    '" appears more than once in header' +
+                    '\nConsidering only the first occurence', AstropyUserWarning)
+
             except KeyError:
                 tbl_dict[key] = [missing_marker] * n_previous
                 tbl_dict[key].append(value)


MSeifert04:
I haven't had the time to do a proper review but the patch seems to be at the wrong place. The dictionary should still append new header-items to the values of the dict, just not multiple entries for one header.

janga1997:
@MSeifert04 I'm not sure i understand what you mean by `new header items to the values of the dict`.

MSeifert04:
The root-cause of the problem is that if a Header contains a keyword twice it will be appended 2 times. That leads to an inconsistent table because one column has more entries than the others.

Your patch however only ever adds one column for each keyword. That's why all the existing tests on `ImageFileCollection` fail.

janga1997:
@MSeifert04 I am not able to understand what is different about checking whether a keyword is in `alreadyencountered` , or in `tbl_dict.keys()`
Aren't they necessarily the same thing?

janga1997:
@MSeifert04 Could you give an example of a case where my approach would yield an different result than yours?

MSeifert04:
The difference is that `tbl_dict` contains as keys a collection of all already processed headers something like: `{'object': ['target', 'target'], 'exptime': [None, 10]}` so you shouldn't check if the key exists but could check the length of these lists. However `alreadyencountered` just is a set of all keys for _one_ header.

You can check the test-case I submitted with my pull request as example (except for the `triage...` parts which are only responsible for cleanup after the tests). 

If you want to fix this PR I will close my pull request again.

janga1997:
@MSeifert04 I do want to fix this PR, partly because as a part of my gsoc application.

bsipocz:
IMO there are plenty of other issues to fix, so if the other PR does the job what about working on some of those?

janga1997:
@bsipocz Actually I am working on another issue. Maybe I will just focus on wrapping that up.

PR #466, opened 2017-03-13
include conda forge as appveyor channel

### LINKED ISSUE #462: AppVeyor Python 3.5 package conflict
The Python 3.5 AppVeyor tests fail because of conflicting package requirements:

```
UnsatisfiableError: The following specifications were found to be in conflict:
  - numpy 1.12*
  - python 3.5*
  - reproject
Use "conda info <package>" to see the dependencies for each package.
```

@mwcraig Is this something that can be fixed with new builds in [astropy-conda-channel](https://github.com/astropy/conda-channel-astropy)?

MSeifert04:
Could fix #462


MSeifert04:
@mwcraig @bsipocz I'm not sure how this will affect the channel priorities on AppVeyor. Was the change here correct?

bsipocz:
@MSeifert04 - This should be fine I think.

bsipocz:
@MSeifert04 - Can you merge this? Then the py36 upgrade PR will probably be passing, too.

MSeifert04:
I hoped @mwcraig or @crawfordsm would have a look at the changes before I merge it (I'm not really comfortable with merging my own PRs).

docs/ccdproc/image_management.rst
```diff
@@ -25,6 +25,10 @@ example initialization looks like::
     >>> keys = ['imagetyp', 'object', 'filter', 'exposure']
     >>> ic1 = ImageFileCollection('.', keywords=keys) # only keep track of keys
 
+    >>> # the following line shouldn't be needed in real files but it's needed
+    >>> # because the testfile has no header entry for "filter".
+    >>> ic1.summary['filter'] = ic1.summary['filter'].astype(str)
+
```
crawfordsm: would it be better to fix test file rather than insert this line into the docs?

crawfordsm:
@MSeifert04 Can you give some more context for the changes in the testing scripts?  Is that also to fix something with the test file?

MSeifert04:
@crawfordsm The basic issue is that the test file header doesn't contain a `'FILTER'`. I opened a seperate issue because most of the test suite uses those keywords #468. The purpose of this PR is mostly to make the test suite passing again. Changing the FITS file and going through the tests is quite an effort but should be made, eventually.

ccdproc/tests/test_image_collection.py
```diff
@@ -642,16 +642,15 @@ def test_keyword_order_is_preserved(self, triage_setup):
...
+        collection = image_collection.ImageFileCollection(
+            location=triage_setup.test_dir,
+            keywords=['imagetyp', 'filter', 'object'])
 
         all_elements = []
         for hdu, fname in collection.hdus(return_fname=True):
             all_elements.append((str(hdu.header), fname))
         # Now sort
-        collection.sort(keys=['filter', 'object'])
+        collection.sort(keys=['imagetyp', 'object'])
```
MSeifert04: Recently `np.ma.argsort` was changed, it didn't compare masked entries before (which are `None` because no header contains a `filter`) but now they do. However `None`s can't be compared.

ccdproc/tests/test_image_collection.py
```diff
@@ -642,16 +642,15 @@ def test_keyword_order_is_preserved(self, triage_setup):
         assert ic.keywords == ['file'] + keywords
 
     def test_sorting(self, triage_setup):
-        collection = image_collection.ImageFileCollection(location=triage_setup.test_dir,
-                                             keywords=['imagetyp',
-                                                       'filter',
-                                                       'object'])
+        collection = image_collection.ImageFileCollection(
+            location=triage_setup.test_dir,
+            keywords=['imagetyp', 'filter', 'object'])
```
MSeifert04: This is not a real change it just refactors the way the code is organized. It's still PEP8 compliant (according to pyflakes).

docs/ccdproc/image_management.rst
MSeifert04: You're right but there are a lot of places where inexistant header keywords (in particular `filter`) are used. I think that should be fixed in the complete test suite. That's a bit of effort, however requires very little knowledge about `ccdproc` itself so I figured it would make a good "first fix". The main objective of this PR is just to make the tests pass again.

bsipocz:
@mwcraig - The reproject PR just passed, so maybe we're OK without conda-forge?

ccdproc/tests/test_image_collection.py
mwcraig: Ah, that makes sense.

docs/ccdproc/image_management.rst
MSeifert04: @mwcraig You mean skipping the doctests on "image management" completly?

mwcraig: if we need to for now, yes....to summarize, as I understand it, the sort fails now because of a change in `np.ma.argsort`. I think we should open a separate issue for that, because it will conceivably affect real use cases.

MSeifert04:
I pushed a commit skipping the doctests in image-management. Unfortunatly almost every line did some comparisons with `filters == 'I'` that fails (or currently raises a Warning). That needs to be reverted again when #468 is adressed.

bsipocz:
The scipy/numpy intersphinx inventories are still unaccessible, so the travis fail is not a real one.

mwcraig:
Thanks @bsipocz, merging!

MSeifert04:
I guess I have to rebase my PRs now 😄 Thanks for merging @mwcraig

PR #467, opened 2017-03-13
Handle duplicated header-keywords in ImageFileCollection

### LINKED ISSUE #464: Handle duplicate keywords in headers in ImageFileCollection

MSeifert04:
Checklist:

- [x] Fixes #464 and Closes #465 

This implements option 3 from https://github.com/astropy/ccdproc/issues/464#issuecomment-285733207:

> Ignore these duplicate entries with a warning.

I haven't implemented a check if these entries are equal but the warning includes a message about the file and the ignored keyword value pair. Is that enough information?

@janga1997 I hope you don't mind that I made another pull request. I started reviewing #465 but that did not fix the bug and introduced a lot of new bugs.

### DIFF
--- a/ccdproc/tests/test_image_collection.py
+++ b/ccdproc/tests/test_image_collection.py
@@ -657,6 +657,20 @@ class TestImageFileCollection(object):
         for i in range(len(collection.summary)):
             assert(collection.summary['file'][i] == collection.files[i])
 
+    def test_duplicate_keywords(self, triage_setup):
+        # Make sure duplicated keywords don't make the imagefilecollection
+        # fail.
+        hdu = fits.PrimaryHDU()
+        hdu.data = np.zeros((5, 5))
+        hdu.header['stupid'] = 'fun'
+        hdu.header.append(('stupid', 'nofun'))
+
+        hdu.writeto(os.path.join(triage_setup.test_dir, 'duplicated.fits'))
+
+        ic = image_collection.ImageFileCollection(triage_setup.test_dir,
+                                                  keywords='*')
+        assert 'stupid' in ic.summary.colnames
+
     def test_ccds_generator_in_different_directory(self, triage_setup, tmpdir):
         """
         Regression test for https://github.com/astropy/ccdproc/issues/421 in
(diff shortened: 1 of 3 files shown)


ccdproc/tests/test_image_collection.py
```diff
@@ -657,6 +657,20 @@ def test_sorting(self, triage_setup):
         for i in range(len(collection.summary)):
             assert(collection.summary['file'][i] == collection.files[i])
 
+    def test_duplicate_keywords(self, triage_setup):
+        # Make sure duplicated keywords don't make the imagefilecollection
+        # fail.
+        hdu = fits.PrimaryHDU()
+        hdu.data = np.zeros((5, 5))
+        hdu.header['stupid'] = 'fun'
+        hdu.header.append(('stupid', 'nofun'))
```
mwcraig: :grin:

ccdproc/tests/test_image_collection.py
```diff
@@ -657,6 +657,20 @@ def test_sorting(self, triage_setup):
...
+
+        hdu.writeto(os.path.join(triage_setup.test_dir, 'duplicated.fits'))
+
+        ic = image_collection.ImageFileCollection(triage_setup.test_dir,
+                                                  keywords='*')
+        assert 'stupid' in ic.summary.colnames
```
mwcraig: Maybe also `assert 'nofun' not in ic.summary['stupid']`?

MSeifert04: Sounds good 👍. I added the test and one for `assert 'fun' in ic.summary['stupid']`.

mwcraig:
Thanks, merging!

PR #469, opened 2017-03-14
removed several list->set or set->list operations in imagefilecollection.keywords setter

MSeifert04:
This is simply a refactor of the keywords setter in ImageFileCollection.

It saves some unnecessary list (2 before, 1 in this PR) / set (3 before, 1 in this PR) casts.


MSeifert04:
ok, the tests passed so I'm merging it now.

PR #475, opened 2017-03-17
Add __repr__ method to ImageFileCollection

janga1997:
For new functionality:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,7 @@
 
 New Features
 ^^^^^^^^^^^^
-
+- Add representation for ImageFileCollection [#475]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -69,6 +69,7 @@ class ImageFileCollection(object):
         self._location = location
         self._filenames = filenames
         self._files = []
+        self._info_file = info_file
         if location:
             self._files = self._get_files()
 
@@ -111,6 +112,18 @@ class ImageFileCollection(object):
 
         if keywords:
             self.keywords = keywords
+
+    def __repr__(self):
+
+        kw = "'*'" if self._all_keywords else self.keywords[1:]
+        str_repr = ("{self.__class__.__name__}(location='{self._location}', keywords={kw}, ").format(self=self, kw=kw)
+
+        if self._info_file is not None:
+            str_repr += ("info_file={}, ").format(self._info_file)
+
+        str_repr +=("filenames={})").format(self._filenames)
+
+        return str_repr
 
     @property
     def summary(self):
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -146,7 +146,7 @@ class ImageFileCollection(object):
         if self._info_file is not None:
             str_repr += ("info_file={}, ").format(self._info_file)
 
-        str_repr +=("filenames={})").format(self._filenames)
+        str_repr += ("filenames={})").format(self._filenames)
 
         return str_repr
 
--- a/ccdproc/tests/test_image_collection.py
+++ b/ccdproc/tests/test_image_collection.py
@@ -151,8 +151,6 @@ class TestImageFileCollection(object):
         assert repr(ic) == "ImageFileCollection(location='{0}', keywords='*', filenames={1})".format(
             triage_setup.test_dir, ic._filenames)
 
-
-
     def test_hdus_masking(self, triage_setup):
         collection = image_collection.ImageFileCollection(location=triage_setup.test_dir,
                                              keywords=['imagetyp', 'exposure'])
--- a/ccdproc/tests/test_image_collection.py
+++ b/ccdproc/tests/test_image_collection.py
@@ -139,6 +139,13 @@ class TestImageFileCollection(object):
             n_hdus += 1
         assert n_hdus == triage_setup.n_test['files']
 
+    def test_repr_method(self, triage_setup):
+        ic = image_collection.ImageFileCollection(location=triage_setup.test_dir)
+        assert repr(ic) == "ImageFileCollection(location='{0}', keywords='*', filenames={1})".format(
+            triage_setup.test_dir, ic._filenames)
+
+
+
     def test_hdus_masking(self, triage_setup):
         collection = image_collection.ImageFileCollection(location=triage_setup.test_dir,
                                              keywords=['imagetyp', 'exposure'])


MSeifert04:
it has been a while since I've seen so much string addition. I generally favor `str.format` because that's faster and more readable. However it seems like a very uncommon way to represent a class. You may want to have a look how other classes implement their representation.

janga1997:
@MSeifert04 I don't know if you meant that comment to be funny, but it really was. :-)
And it does look a lot cleaner with `str.format`. My bad.

MSeifert04:
Yeah, format looks cleaner. 👍 

But it doesn't really look like a [representation](https://docs.python.org/3/reference/datamodel.html#object.__repr__). That may just be my interpretation, though.

janga1997:
@MSeifert04 What's in your mind?

MSeifert04:
When you read the above linked description it states: The representation should "look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). ". That's definetly not the case here. I'm not sure if it's impossible, I haven't took a closer look.

janga1997:
So that would be just
`ImageFileCollection(location='.',filenames = ['{file_name}.fits'], keywords= '*')`, or something like that.?

MSeifert04:
Yeah, something like that! However it definetly needs some tests (either locally or in the test suite) that test if it produces an identical instance.

janga1997:
@MSeifert04 
One little doubt. For the user to be able to recreate that object, we have to print either `ccdproc.ImageFileCollection` or `ImageFileCollection`, or something other, depending on how the user imports the module and this class. How do we do that?

MSeifert04:
> How do we do that?

Like I said before "You may want to have a look how other classes implement their representation". 

However if you're really stuck: Generally it's solved by using `self.__class__.__name__` maybe also using `self.__class__.__module__.split('.')[0]` to get the top-level module for the class.

janga1997:
@MSeifert04 I did not mean that. What I meant was, if the user did `import ccdproc as ccd`
and then
 `test = ccd.ImageFileCollection(location='.',filenames = ['{file_name}.fits'], keywords= '*')`
But `self.__class__.__module__.split('.')[0]` returns `ccdproc`, and `self.__class__.__name__` returns 
`ImageFileCollection`, none of which directly helps me to recreate that object.

In this case, I need it to be 
`ccd.ImageFileCollection(location='.',filenames = ['{file_name}.fits'], keywords= '*')`
and not 
`ImageFileCollection(...)` or `ccdproc.ImageFileCollection(...)`.

Do we even need to do that, btw?

MSeifert04:
You're overthinking this and I totally misunderstood your previous comment 😄. Sorry.

The representation is not required to _always_ recreate the object. Just in _"given an appropriate environment"_. That means how the object should be loaded. For example if we expect users to do `from ccdproc import ImageFileCollection` then it should just be `ImageFileCollection(...)`, when most docstrings (and such like) recommend `import ccdproc` then it should be `ccdproc.ImageFileCollection(...)`. However generally just the classname should be enough. The representation will be long enough already.

janga1997:
@MSeifert04 I hope it looks good enough now. If so, I will add the tests.

MSeifert04:
> I hope it looks good enough now.

Could you show some exemplaric representations for different instances? From the code alone it's quite hard to figure that out :)

janga1997:
```
In [5]: test = ccdproc.ImageFileCollection(location='.',filenames = ['janga.fits'], keywords= '*')

In [6]: test
Out[6]: ImageFileCollection(location='.', keywords='*', filenames=['janga.fits'])

In [7]: test.keywords = ['history', 'lonpole', 'bunit']

In [8]: test
Out[8]: ImageFileCollection(location='.', keywords=['bunit', 'lonpole', 'history'], filenames=['janga.fits'])
```

I added a conditional for the keywords, because for a file with a large header, it becomes really ugly to see a large representation.

CHANGES.rst
```diff
@@ -4,7 +4,7 @@
 
 New Features
 ^^^^^^^^^^^^
-
+- Add Representation for ImageFileCollection []
```
MSeifert04: You'll need to fill in the PR number. Also no need to capitalize `representation` (I guess, not sure about that).

ccdproc/image_collection.py
```diff
@@ -69,6 +69,7 @@ def __init__(self, location=None, keywords=None, info_file=None,
         self._location = location
         self._filenames = filenames
         self._files = []
+        self._info_file = info_file
```
MSeifert04: Is it really necessary to keep the info-file as attribute? I'm not sure but couldn't that interfere in the representation with the other attributes?

janga1997: @MSeifert04 But is there another way to reference `info_file`?

ccdproc/image_collection.py
```diff
@@ -112,6 +113,19 @@ def __init__(self, location=None, keywords=None, info_file=None,
         if keywords:
             self.keywords = keywords
 
+    def __repr__(self):
+
+        str_repr = ("{0}(location='{1}', keywords={2}, ").format(
```
MSeifert04: Instead of positional parameters you can also use named parameters, for example:

```
kw = '*' if self._all_keywords else self.keywords[1:]
r = "{self.__class__.__name__}(location={self._location}, keywords={kw})".format(self=self, kw=kw)
```

That also allows to make the base-representation (before formatting) depend on `self._info_file` and `self._filenames` instead of adding these.

ccdproc/image_collection.py
MSeifert04: New attributes are always tricky because all sort of modifications on the instance need to keep the class invariants. Each added invariant makes the class more difficult. For example can you garantuee that any change on an instance correctly handles this attribute?

I think that may be much more trouble than it's beneficial.

janga1997: So I should just remove `info_file` from the `repr`?

MSeifert04: Maybe wait for someone else to chime in. I actually haven't used the `info_file` ever so I'm not sure. 😄

janga1997:
@MSeifert04 About the tests, if I am comparing an object with the object generated from its string representation, I suppose I should compare the summary_info attributes, which returns 
```
masked_array(data = [ True],
             mask = False,
       fill_value = True)
```. 
And I should just mind about the the value of `data`?

**Comment by MSeifert04** [comment]
I'm not exactly sure I understand what you mean.

**Comment by mwcraig** [comment]
A far easier test than generating an `ImageFileCollection` from its `__repr__` and then testing whether the generated objects are identical would be to just check that the expected string gets returned. So, for example, in the body of the test

```python
     ic = ImageFileCollection(location=triage_setup.test_dir)
     assert repr(ic) == 'ImageFileCollection(location={})'.format(triage_setup.test_dir
```

@MSeifert04 if this style test seems OK to you can +1 this?

MSeifert04:
yeah, that's what I thought.

Just one thought: This PR should be merged after #463 because the `ext` needs to be added as well, right?

mwcraig:
Yes, good catch!

mwcraig:
Merging, thanks @janga1997!

(the remaining test failure is due to an unrelated issue).

PR #482, opened 2017-03-30
License update and move

bsipocz:
This PR moves the license file to the root dir (discussion e.g. https://github.com/astropy/astropy/issues/5914), and also contains a few updates to make it more consistent over the package. 

Kept the two things in separate commits, so if you don't want to move can just omit the latter.


licenses/README.rst
```diff
@@ -1,5 +1,8 @@
 Licenses
 ========
 
-This directory holds license and credit information for the affiliated package,
-works the affiliated package is derived from, and/or datasets.
+This directory holds license and credit information for works the ccdproc
+package is derived from or distributes, and/or datasets.
+
+License file for the ccdproc package itself is placed in the root directory
```
MSeifert04: Minor suggestion: A "The" at the start of the sentence?

docs/license.rst
```diff
@@ -9,4 +9,4 @@ Ccdproc License
 
 Ccdproc is licensed under a 3-clause BSD style license:
 
-.. include:: ../licenses/LICENSE.rst
+.. include:: ../LICENSE.rst
```
MSeifert04: hm, is it a good idea to `include` the license file? Wouldn't `literalinclude` make more sense?

MSeifert04:
LGTM :) Modulo two comments (even if one isn't even related to the PR) 👍

docs/license.rst
bsipocz: I leave you to decide on this for ccdproc, astropy core does exactly the same as above.

MSeifert04:
Thanks @bsipocz!

PR #483, opened 2017-04-08
Remove prints in tests

MSeifert04:
Actually I don't know why these `print`s were in these tests. `pytest` actually does a great job at displaying differences so it seems they were used to create the tests but should've been removed afterwards.

Correct me if I'm wrong.


mwcraig:
Oops. Part of the reason they stayed in this long is that `pytest` is also great at suppressing print statements when the tests pass.

PR #484, opened 2017-04-09
Add Unix-style filename inclusion/exclusion to imagecollection

adrn:
- [ ] Did you add yourself to the "Authors.rst" file?

For new functionality:

I've been using `ccdproc` quite a bit lately for doing spectroscopic data reduction. I've found that I often want to be able to select subsets of my data based on filename pattern matching to test out certain pieces of the reduction pipeline, so I had implemented something like the features in this PR locally on a subclass of `ImageFileCollection`. If you like it, I'll add a changelog entry.

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -70,6 +70,17 @@ class ImageFileCollection(object):
     """
     def __init__(self, location=None, keywords=None, info_file=None,
                  filenames=None, glob_include=None, glob_exclude=None):
+
+        # Include or exclude files from the collection based on glob pattern
+        # matching - has to go above call to _get_files()
+        if glob_exclude is not None:
+            glob_exclude = str(glob_exclude) # some minimal validation
+        self.glob_exclude = glob_exclude
+
+        if glob_include is not None:
+            glob_include = str(glob_include)
+        self.glob_include = glob_include
+
         self._location = location
         self._filenames = filenames
         self._files = []
@@ -106,16 +117,6 @@ class ImageFileCollection(object):
                                    info_path)
                 else:
                     raise
-
-        # Include or exclude files from the collection based on glob pattern
-        # matching
-        if glob_exclude is not None:
-            glob_exclude = str(glob_exclude) # some minimal validation
-        self.glob_exclude = glob_exclude
-
-        if glob_include is not None:
-            glob_include = str(glob_include)
-        self.glob_include = glob_include
 
         # Used internally to keep track of whether the user asked for all
         # keywords or a specific list. The keywords setter takes care of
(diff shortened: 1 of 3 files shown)


ccdproc/image_collection.py
```diff
@@ -330,7 +351,22 @@ def _get_files(self):
             else:
                 files = self._filenames
         else:
-            files = self._fits_files_in_directory()
+            _files = self._fits_files_in_directory()
+
+            files = []
+            for fn in _files:
```
MSeifert04: Personally I prefer list comprehensions because the logic is more straight-forward:

```
if self.glob_include is not None:
    files = fnmatch.filter(files, self.glob_include)
if self.glob_exclude is not None:
    files = [file for file in files if not fnmatch.fnmatch(fn, self.glob_exclude)]
```

It's probably also faster because it doesn't need to check the `self.glob_* is not None` each iteration and [`fnmatch.filter`](https://docs.python.org/library/fnmatch.html#fnmatch.filter) advertises itself as "being faster".

ccdproc/image_collection.py
```diff
@@ -62,14 +62,35 @@ class ImageFileCollection(object):
...
         value.
     """
     def __init__(self, location=None, keywords=None, info_file=None,
-                 filenames=None):
+                 filenames=None, glob_include=None, glob_exclude=None):
+
+        # Include or exclude files from the collection based on glob pattern
+        # matching - has to go above call to _get_files()
+        if glob_exclude is not None:
+            glob_exclude = str(glob_exclude) # some minimal validation
```
MSeifert04: `str` almost never fails, at least on Python3. Wouldn't it make more sense to let it fail when it actually does the `fnmatch`? Also the exception message when `str` fails is probably not helpful either so I don't see a compelling argument for this "minimal validation".

Or am I misunderstanding this?

ccdproc/tests/test_image_collection.py
```diff
@@ -700,3 +700,22 @@ def test_ccds_generator_in_different_directory(self, triage_setup, tmpdir):
         # This generated an IOError before the issue was fixed.
         for _ in coll.ccds(ccd_kwargs={'unit': 'adu'}):
             pass
+
+    def test_glob_matching(self, triage_setup):
+        # We'll create two files with strange names to test glob
+        #   includes / excludes
+        one = fits.PrimaryHDU()
+        one.data = np.zeros((100, 100))
```
MSeifert04: This is possibly just a copy&paste but there's no point in creating such a "big image" here, could you change it to `(5, 5)` (or something like that)?

MSeifert04:
Thanks @adrn for the Pull request. I think this would be a great enhancement. 👍

PR #486, opened 2017-04-11
remove deprecated summary_info propery of ImageFileCollection

MSeifert04:

The attribute is marked as deprecated in the docstring and it was since the `ImageFileCollection` was implemented in #217 . I had to change quite a lot to actually remove it.

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -100,7 +100,7 @@ class ImageFileCollection(object):
         if self._files == []:
             warnings.warn("no FITS files in the collection.",
                           AstropyUserWarning)
-        self._summary_info = {}
+        self._summary = {}
         if keywords is None:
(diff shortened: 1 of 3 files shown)


ccdproc/image_collection.py
```diff
@@ -142,18 +142,6 @@ def summary(self):
         return self._summary_info
 
     @property
-    def summary_info(self):
```
mwcraig: Add astropy deprecation warning to function.

MSeifert04:
The property and the test are restored. It will now warn the user if he accessed the property about the future removal.

MSeifert04:
The tests passed. So this is ready for a final review 😄

mwcraig:
Thanks again for catching this and cleaning it up...and sorry it made it into the codebase in the first place 😊

PR #491, opened 2017-04-22
gain_corrected in ccd_process

### LINKED ISSUE #490: Why do we need unit conversion to bias/flat before doing ccd_process?
I have asked this to the "Python users in Astronomy" group page of FB and M. Durant and M. Craig kindly suggested me to open an issue here.

`ccdproc.ccd_process` takes the `master_bias` and `master_flat` in the unit of "after gain correction", which usually means "electron" unit. But the bias/flat I have is in ADU, which is usually the case, and of course the input image (`CCDData` object) is in ADU too.

What I learnt from M. Craig's answer is that I have to multiply gain (`gain=2.1 * u.electron/u.adu`) to bias and flat but NOT to the object image, before plugging those in to `ccd_process`:

```python
bias = CCDData.read('../bias/bias.fits', unit='adu')
obj  = CCDData.read('obj.fits', unit='adu')
reduced = ccd_process(obj, master_bias=bias) # This works correctly
reduced = ccd_process(obj, master_bias=bias,  # This gives error
                      gain=2.1*u.electron/u.adu, 
                      readnoise=11*u.electron))
```

But the second one gives following error:
```
UnitConversionError: Can only apply 'subtract' function to quantities with compatible dimensions
```

For me this seems quite strange that I have to multiply gain to the bias and flat before plugging them in to `ccd_process`. It is clear that it should multiply gain to get electron unit if the unit of bias is in adu. Is there any specific reason for this, i.e., why do we have to multiply gain to bias/flat before using them?

This happens when I plug "gain" factor into the ccd_process.

Thanks in advance :)

crawfordsm:
This Fixes #490 and adds the option to include ``gain_corrected``.  If set to False, the calibrations files will be assumed to be un-gained corrected and the gain correction will only occur after those corrections are applied to the gain correction

For new functionality:

Default value for ``gain_corrected`` is True, which is inline with backward compatibility

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,9 +13,8 @@ New Features
 - Add ``bitfield_to_boolean_mask`` function to convert a ``bitfield`` to a
   boolean mask (following the numpy conventions). [#460]
 
-- Added ``gain_corrected`` option in ccd_process so that 
-  calibration files do not need to previously been gain
-  corrected [#491]
+- Added ``gain_corrected`` option in ccd_process so that calibration 
+  files do not need to previously been gain corrected [#491]
 
 
 Other Changes and Additions
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,10 @@ New Features
 - Added ``glob_include`` and ``glob_exclude`` parameter to
   ``ImageFileCollection`` [#484]
 
+- Added ``gain_corrected`` option in ccd_process so that 
+  calibration files do not need to previously been gain
+  corrected 
+
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -166,7 +166,7 @@ def ccd_process(ccd, oscan=None, trim=None, error=False, master_bias=None,
         Default is ``False``.
 
     gain_corrected : bool, optional
-        If True, the ``master_bias``, ``master_flat``, and ``dark_frame`` 
+        If True, the ``master_bias``, ``master_flat``, and ``dark_frame``
         have already been gain corrected.  Default is ``True``.
 
     Returns
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -56,7 +56,7 @@ def ccd_process(ccd, oscan=None, trim=None, error=False, master_bias=None,
                 gain=None, readnoise=None, oscan_median=True, oscan_model=None,
                 min_value=None, dark_exposure=None, data_exposure=None,
                 exposure_key=None, exposure_unit=None,
-                dark_scale=False):
+                dark_scale=False, gain_corrected=True):
     """Perform basic processing on ccd data.
 
     The following steps can be included:
@@ -164,6 +164,10 @@ def ccd_process(ccd, oscan=None, trim=None, error=False, master_bias=None,
         If True, scale the dark frame by the exposure times.
         Default is ``False``.
 
+    gain_corrected : bool, optional
+        If True, the calibration frames have already been gain. 
+        Default is ``True``.
+
     Returns
     -------
     occd : `~ccdproc.CCDData`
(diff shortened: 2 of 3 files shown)


MSeifert04:
@crawfordsm It seems you're not linked to your commits. I had a similar issue lately and even though I don't know what exactly fixed it I bookmarked [this github help page on "Commits are not linked to any user"](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user). Or is that intended?

crawfordsm:
Yes, apparently I've had my email set wrong on this new machine due to a typo -- I'm amazed by how many commits I've been allowed to make with that set wrong

bsipocz:
@crawfordsm - You can add that email address to your github account, and the it will be recognized. Alternatively listing it in the ``.mailmap`` file may be enough, too.

ccdproc/tests/test_ccdproc.py
```diff
@@ -915,3 +915,39 @@ def test_ccd_process():
     assert(occd.unit == u.electron)
     # Make sure the original keyword is still present. Regression test for #401
     assert occd.meta['testkw'] == 100
+
+def test_ccd_process_gain_corrected():
+    # test the through ccd_process
```
MSeifert04: This sentence is missing something :)

CHANGES.rst
```diff
@@ -13,6 +13,10 @@ New Features
 - Add ``bitfield_to_boolean_mask`` function to convert a ``bitfield`` to a
   boolean mask (following the numpy conventions). [#460]
 
+- Added ``gain_corrected`` option in ccd_process so that
```
MSeifert04: Were the line breaks at 60 intentional?

ccdproc/core.py
```diff
@@ -262,6 +265,10 @@ def ccd_process(ccd, oscan=None, trim=None, error=False, master_bias=None,
         raise TypeError(
             'master_flat is not None or a CCDData object.')
 
+    # apply the gain correction only at the end if gain_corrected is False
+    if isinstance(gain, Quantity) and gain_corrected is False:
```
MSeifert04: `not gain_corrected` seems better than `gain_corrected is False` and it would prevent accidentally not triggering the branch.

ccdproc/core.py
```diff
@@ -224,13 +228,12 @@ def ccd_process(ccd, oscan=None, trim=None, error=False, master_bias=None,
         raise TypeError('bad_pixel_mask is not None or numpy.ndarray.')
 
     # apply the gain correction
-    if isinstance(gain, Quantity):
-        nccd = gain_correct(nccd, gain)
-    elif gain is None:
-        pass
-    else:
+    if not (gain is None or isinstance(gain, Quantity)):
```
MSeifert04: the `isinstance(gain, Quantity)` is used 3 times in the function now. You could use an intermediate variable so it's only computed once.

ccdproc/core.py
```diff
@@ -165,6 +165,10 @@ def ccd_process(ccd, oscan=None, trim=None, error=False, master_bias=None,
         If True, scale the dark frame by the exposure times.
         Default is ``False``.
 
+    gain_corrected : bool, optional
```
MSeifert04: I'm not exactly sure that's the best approach because the name could also mean that the "object" has been gain corrected.

ccdproc/core.py
```diff
@@ -165,6 +165,10 @@ def ccd_process(ccd, oscan=None, trim=None, error=False, master_bias=None,
...
+        If True, the calibration frames have already been gain corrected.
```
MSeifert04: Maybe a hint which parameters correspond to the "calibration frames" could be added.

ccdproc/core.py
crawfordsm: yes, can be removed later on as it is already checked and simplified the code

ccdproc/core.py
mwcraig: what about calling the parameter `gain_corrected_masters`?

PR #497, opened 2017-05-22
Fix failures in astropy-dev tests

### LINKED ISSUE #496: Fix errors on cron builds against astropy master
See, e.g., https://travis-ci.org/astropy/ccdproc/jobs/232954337#L635

Two things to fix (copied from that error log):

```
ImportError while importing test module '/tmp/ccdproc-test-j_l9nth0/lib.linux-x86_64-3.6/ccdproc/tests/test_combiner.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
ccdproc/tests/test_combiner.py:13: in <module>
    from astropy.utils.compat import NUMPY_LT_1_9
E   ImportError: cannot import name 'NUMPY_LT_1_9'
```

and

```
============================ pytest-warning summary ============================
WC1 None [pytest] section in setup.cfg files is deprecated, use [tool:pytest] instead.
```

mwcraig:
Closes #496

This fixes the issue in our tests with astropy dev -- the definition of `NUMPY_LT_1_9` was removed recently. It was used only to skip a test in one case that we no longer test against anyway...

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -631,7 +631,7 @@ class ImageFileCollection(object):
         # be omitted, hence use ~matches.
         logger.debug('Matches: %s', matches)
         self.summary['file'].mask = ma.nomask
-        self.summary['file'][~matches] = ma.masked
+        self.summary['file'].mask[~matches] = True
 
     def _fits_files_in_directory(self, extensions=None,
                                  compressed=True):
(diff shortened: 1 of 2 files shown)


mwcraig:
It looks like there is an additional issue, as seen in the travis log for the build against numpy-dev/astropy-dev. The fail in question is related to something in astropy, not numpy, since the tests pass when run against astropy 1.3.2 and numpy-dev.

Somehow running this in the test that fails (`test_image_collection.TestImageFileCollection.test_hdus_masking`) changes the names of the non-bias files to `'0.0'`:

```python
for hdu in collection.hdus(imagetyp='bias'):
    pass
```

mwcraig:
Ah, so whether the change is in numpy or astropy, masked file names are now being converted to zero when a numpy array is created form a masked table.

mwcraig:
Merging so that tests pass again (not much to this PR, either). Made no changelog entry because neither of these is user-facing.

PR #502, opened 2017-06-03
Using system pytest rather than astropy bundled

bsipocz:
As the bundled version gets deprecation with v2.0


MSeifert04:
The failure with LTS seems to be real, the last build passed [just one day ago](https://travis-ci.org/astropy/ccdproc/builds/238604495).

mwcraig:
Thanks, @bsipocz!

PR #473, opened 2017-03-16
[Do not merge] Enable deprecation-checks and fix them

MSeifert04:
This fixes all deprecations (not including python 3.6 deprecations) encountered in the test suite.

The PR isn't intended to be merged yet, I want to rebase and check for 3.6 as soon as #449 is merged.

However locally I get a lot of 

```
________________ [doctest] docs/ccdproc/reduction_toolbox.rst _________________
_pytest.runner:149: in __init__
    ???
_pytest.runner:137: in <lambda>
    ???
_pytest.vendored_packages.pluggy:724: in __call__
    return self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
_pytest.vendored_packages.pluggy:338: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
_pytest.vendored_packages.pluggy:333: in <lambda>
    _MultiCall(methods, kwargs, hook.spec_opts).execute()
_pytest.vendored_packages.pluggy:595: in execute
    return _wrapped_call(hook_impl.function(*args), self.execute)
_pytest.vendored_packages.pluggy:253: in _wrapped_call
    return call_outcome.get_result()
_pytest.vendored_packages.pluggy:278: in get_result
    raise ex[1].with_traceback(ex[2])
_pytest.vendored_packages.pluggy:264: in __init__
    self.result = func()
_pytest.vendored_packages.pluggy:596: in execute
    res = hook_impl.function(*args)
_pytest.runner:90: in pytest_runtest_call
    ???
C:\...\lib\site-packages\astropy-2.0.dev17912-py3.5-win-amd64.egg\astropy\tests\pytest_plugins.py:204: in runtest
    raise_on_error=True, verbose=False, encoding='utf-8')
C:\...\lib\doctest.py:2065: in testfile
    runner.run(test)
C:\...\lib\doctest.py:1830: in run
    r = DocTestRunner.run(self, test, compileflags, out, False)
C:\...\lib\doctest.py:1453: in run
    self.debugger = _OutputRedirectingPdb(save_stdout)
C:\...\lib\doctest.py:357: in __init__
    pdb.Pdb.__init__(self, stdout=out, nosigint=True)
C:\...\lib\pdb.py:153: in __init__
    import readline
C:\...\lib\site-packages\readline.py:34: in <module>
    rl = Readline()
C:\...\lib\site-packages\pyreadline\rlmain.py:422: in __init__
    BaseReadline.__init__(self)
[...]

PR #507, opened 2017-06-07
fix unneeded compatibility import

MSeifert04:
This fixes a failure with astropy dev because we still imported the (now removed: https://github.com/astropy/astropy/pull/6164) `astropy.utils.compat.gzip` module.

### DIFF
--- a/ccdproc/tests/pytest_fixtures.py
+++ b/ccdproc/tests/pytest_fixtures.py
@@ -2,6 +2,7 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
+import gzip
 from tempfile import mkdtemp
 import os
 from shutil import rmtree
@@ -13,7 +14,6 @@ from astropy import units as u
 from astropy.utils import NumpyRNGContext
 from astropy.io import fits
 
-from astropy.utils.compat import gzip
 
 from ..ccddata import CCDData
 


bsipocz:
Ideally the deprecations should have been raised as exceptions, so the discovery of such things comes earlier. Is there any reason that you don't do that in ccdproc?

MSeifert04:
I regularly test with deprecations turned on. However it's a terrible mess on Windows because there are lots of doctest problems there (see https://github.com/astropy/ccdproc/pull/473#issue-214565757). But it never catched that one. No idea why.

MSeifert04:
Maybe because it's an `AstropyDeprecationWarning`. I think those aren't caught by the `enable_deprecations_as_exceptions`.

bsipocz:
Hmm, then it's a bug, they should definitely be raised in affiliates.

MSeifert04:
@bsipocz That's definetly weird. But as you can see in #473 (especially this log: https://travis-ci.org/astropy/ccdproc/builds/211587272, where the deprecations were active) it did not raise an Exception.

MSeifert04:
@mwcraig @crawfordsm The change itself should be pretty uncontroversial and passed the tests (fixed the test against astropy master branch). 

Do you have any objections?

MSeifert04:
@bsipocz Thanks for opening the issue! :)

bsipocz:
I would say go ahead and merge this, as it's rather trivial and now there is no CI queue unlike this evening.

MSeifert04:
I don't really like to merge my own PRs. But I guess you're right. :)

Issue #495, opened 2017-05-16
Attributes of imagefilecollection


MSeifert04:
There have been some new attributes `glob_include` and `glob_exclude` and maybe the `extension` attribute from #463 especially if #475 is merged.

MSeifert04:
Fixed in #515

PR #515, opened 2017-07-18
Added new attributes to ImageFileCollection representation

MSeifert04:
This partially adresses #495.

This adds all new attributes `ext, globs` to the representation of `ImageFileCollection`.

TODO: Could needs some more tests :)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@
 
 New Features
 ^^^^^^^^^^^^
-- Add representation for ImageFileCollection [#475]
 
+- Add representation for ImageFileCollection [#475, #515]
 
 - Added ext parameter and property to ImageFileCollection to specify the FITS extension.[#463]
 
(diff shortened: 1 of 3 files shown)


MSeifert04:
The Sphinx errors are due to a Connection issue:

```
WARNING: intersphinx inventory 'https://docs.scipy.org/doc/scipy/reference/objects.inv' not fetchable due to <class 'requests.exceptions.ConnectionError'>: HTTPSConnectionPool(host='docs.scipy.org', port=443): Max retries exceeded with url: /doc/scipy/reference/objects.inv (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7f1ff0b11050>: Failed to establish a new connection: [Errno 110] Connection timed out',))
```

The AppVeyor failure is because of missing packages on windows with python 3.6.

But the coverage failure is because I was too lazy to make more tests.

bsipocz:
@MSeifert04 - Hmm, this is unfortunate. The scipy and numpy intersphinx are actually picked up from the stsci server, so that warnings should have been caught and never have made the build fail. Could you please open an issue for it in astropy-helpers?

MSeifert04:
I added tests for the branches in `__repr__`. This should be ready for a final-review. :)

PR #517, opened 2017-07-18
Use build_docs instead of the deprecated build_sphinx in travis

MSeifert04:
While looking into a failure with the sphinx build I noticed that we use the deprecated `build_sphinx` command there: https://travis-ci.org/astropy/ccdproc/jobs/255026568#L621

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ matrix:
 
         # Check for sphinx doc build warnings
         - python: 2.7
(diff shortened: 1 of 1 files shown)


MSeifert04:
@mwcraig @crawfordsm Can we merge this?

mwcraig:
Merging, and we should expect tests on windows + python 3 to pass going forward...

PR #522, opened 2017-08-12
Minor cleanup in imagefilecollection tests

MSeifert04:
This imports `ImageFileCollection` directly in the test suite and for a few functions I removed the `tmpdir` fixture.

Overall this makes the test code just a bit more compact.


MSeifert04:
@mwcraig @crawfordsm That PR should be fairly straight-forward. The question is just if that's an improvement or not.

mwcraig:
It reduces the number of lines/characters in the code and is easier to read...I'm all for that!

PR #523, opened 2017-08-13
Run ImageFileCollection tests on AppVeyor

MSeifert04:
They are passing locally on my windows computer so I thought it might be worth testing if they really fail on AppVeyor.


mwcraig:
Thanks for checking this!

Issue #511, opened 2017-06-27
Clean up travis testing


crawfordsm:
With the release of 1.3, we should review the libraries tested against and update them accordingly.

MSeifert04:
Possible candidates for dropping after 1.3:

- Python 3.4
- NumPy 1.10
- astropy 1.0 and 1.3

MSeifert04:
xref #615 

We also probably should add a changelog when we figured out what has been dropped.

PR #524, opened 2017-08-13
Include a numpy 1.12 test in travis

MSeifert04:
NumPy 1.13 has been released for a few weeks so don't test against NumPy 1.12. 

In case we want to drop NumPy 1.10 support (or at least don't test against it) it could also be shuffled:

- Python 3.4 + NumPy 1.11 (previously NumPy 1.10)
- Python 3.5 + NumPy 1.12 (previously NumPy 1.11)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,9 +65,11 @@ matrix:
         - python: 3.4
           env: NUMPY_VERSION=1.10
 
-        # Try older numpy, python versions
         - python: 3.5
           env: NUMPY_VERSION=1.11
+
+        - python: 3.6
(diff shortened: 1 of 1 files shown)


mwcraig:
I think we could drop Python 3.4/numpy 1.10 altogether. I don't have strong feelings about what version of numpy we test against on python 3.5, but 1.11 seems fine.

MSeifert04:
I wasn't sure if we agreed to revisit the testing matrix before or after the next release (see https://github.com/astropy/ccdproc/issues/511#issue-238960015).

Issue #516, opened 2017-07-18
Make the glob attributes for ImageFileCollection readonly?


MSeifert04:
The `ImageFileCollection` has `glob_include` and `glob_exclude`. These are currently settable but they only take effect if one `refreshs` the instance. The `location`, `files` and `ext` attributes are readonly because they also would take effect only when refreshed.

Should the `glob_include/exclude` also be readonly?

crawfordsm:
Yes, they should be readonly.

PR #526, opened 2017-08-15
Make glob attributes of ImageFileCollection readonly

MSeifert04:
Fixes: #516 

This only affects the development branch so it shouldn't need a changelog.


mwcraig:
Thanks!

PR #527, opened 2017-08-15
Raise an exception instead of silently ignoring overwrite in ccd generator

### LINKED ISSUE #453: FITS files not updated for ImageFileCollection.ccds(overwrite=True)

MSeifert04:
It might make sense to think how one could use ``overwrite=True`` with the `ccds` generator in `ImageFileCollection` but given that there could be multiple extensions involved (which the generator doesn't know about) and because `CCDData` doesn't support many in-place operations that won't be straight-forward. So I'm not sure if #453 can be closed afterwards.

However currently the parameter is silently ignored without warning or exception which this PR fixes by raising a `NotImplementedError`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,10 @@ Bug Fixes
 - ``ImageFileCollection`` now handles Headers with duplicated keywords
   (other than ``COMMENT`` and ``HISTORY``) by ignoring all but the first. [#467]
 
+- The ``ccd`` method of ``ImageFileCollection`` will raise an
+  ``NotImplementedError`` in case the parameter ``overwrite=True`` or
+  ``clobber=True`` is used instead of silently ignoring the parameter.
+
 
 1.2.0 (2016-12-13)
 ------------------
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -895,6 +895,9 @@ class ImageFileCollection(object):
         name='image', default_scaling='False', return_type='numpy.ndarray')
 
     def ccds(self, ccd_kwargs=None, **kwd):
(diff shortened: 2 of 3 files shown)


mwcraig:
Good catch; merging!

PR #533, opened 2017-08-17
Remove some unnecessary set calls and replace some with literals

MSeifert04:


mwcraig:
Thanks for all the code cleanup, @MSeifert04!

PR #534, opened 2017-08-20
Make key argument for ImageFileCollection.sort mandatory

MSeifert04:
Just found out by coincidence that `sort` without argument always throws a `ValueError` so it's probably better to make the `keys` argument mandatory.


mwcraig:
Funny but true...I completely forgot there was a `sort` method!

PR #535, opened 2017-08-20
Make "header_keywords" argument for "_fits_summary" method mandatory

MSeifert04:
Almost immediately the argument will be passed to `list` which would throw a `TypeError` because `None` isn't iterable.

Given that it's a private method I didn't add a test and changelog.


mwcraig:
Looks good, thanks for this and the other cleanup PRs.

PR #537, opened 2017-08-20
[do not merge] Use the context manager inside ImageFileCollection generator method

MSeifert04:
Given that we explicitly raise an Exception (two places) after opening the file it's probably not a bad idea to use `fits.open` as context manager in the `_generator` method.

Unfortunately this is quite a big changeset but it actually just indents lots of lines...

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -821,57 +821,56 @@ class ImageFileCollection(object):
 
         for full_path in self._paths():
             no_scale = do_not_scale_image_data
-            hdulist = fits.open(full_path,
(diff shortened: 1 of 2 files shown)


MSeifert04:
This breaks because of unknown reasons on python 2.7 so I've put a WIP in the title and opened an issue for astropy (cross-link: https://github.com/astropy/astropy/issues/6465).

MSeifert04:
It seems like there's something going really wrong when the `_generators` are zipped I changed the test to do the iterations individually. That seems to work without problems.

MSeifert04:
Superceded by #545

Issue #481, opened 2017-03-29
ImageFileCollection's methods to accept multiple word keywords


bsipocz:
Most of the data in ESO produced headers are stored in HIERARCH keywords with multiple words. Currently it’s not possible to provide these nicely as keywords for methods of ImageCollection. 

My current workaround is to provide a dict with the keyword requirements and unpacking it at method call:

```
hdu, fname = image_coll.hdus(return_fname=True, **{'cdelt1':1, 'ESO TPL ID':'EFOSC_spec_obs_Spectrum’})
```

It would be nice to have be able to provide this dict directly via a keyword (e.g. kwd_dict=None), or mentioned this in the docs.

MSeifert04:
That's definetly annoying. But if one has to create the dictionary anyhow then a new parameter would be just a convenience to omit the unpacking (`**`). Also having a `kwd_dict` besides `kwd` and `ccd_kwargs` as parameters makes a function signature really confusing.

But it should be documented, I wonder where [the promised Examples](https://github.com/astropy/ccdproc/blob/master/ccdproc/image_collection.py#L692) are...

PR #539, opened 2017-08-29
Add replace_ argument for files_filtered and _generator

MSeifert04:

Fixes: #481 

CC: @bsipocz

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,10 @@ New Features
 
 - Added ``sum`` option in method for ``combime``. [#500, #508]
 
+- Add ``_replace_with_whitespace`` argument for the ``files_filtered`` and
+  generator methods of ``ImageFileCollection`` to allow filtering for whitespace
+  seperated keys. [#539]
+
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,9 +28,8 @@ New Features
 
 - Added ``sum`` option in method for ``combime``. [#500, #508]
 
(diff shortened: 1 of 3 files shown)


MSeifert04:
I'm not sure if it's good style to have an argument that starts with an underscore, I just wanted to avoid unnecessary conflicts with valid argument names and it distinguishes it from regular filtering argument names. But it could also be considered private (Python convention) which wouldn't be good.

MSeifert04:
Possible alternatives:

- instead of making this an argument it could also be an instance attribute.
- other argument name, not starting with an underscore, but that *could* conflict with existing header keywords.

MSeifert04:
I have thought about the alternatives some more and I would like to keep it an argument and I don't think it would be a problem with the leading underscore. However I would prefer a shorter name, e.g. `_repl` or `_replace` or such like.

bsipocz:
Thanks @MSeifert04 for picking this up. 
I have to admit that I I won't have time in the coming weeks to go back to that piece of work that triggered this feature, but this definitely seems to be an improvement.

mwcraig:
I think `replace_with_whitespace` would work as a keyword without a leading underscore (it seems unlikely to be a keyword). If we want to underscore it, could we go with `replace_with_whitespace_`?

This is good to go otherwise...

MSeifert04:
Yes, it shouldn't create conflicts, I don't think there are many files with a `replace_with_whitespace` header card around. But also the length of the argument is a bit annoying but the shorter it gets the more likely there will be collisions.

mwcraig:
What about going with `replace_`?

MSeifert04:
I really like `replace_` and changed the argument name. :)

PR #545, opened 2017-08-30
Fix RessourceWarnings due to open files in ImageFileCollection._generator

MSeifert04:
These `ResourceWarnings` were because files were still open and the reason is that the `_generator` opens a file, then yields the contents but only closes the file **if** the next element is requested. That works fine for loops but actually doesn't close the file when the generator isn't advanced.

So this approach opens the file to extract the relevant data, then closes it before yielding the "object" and if overwriting or a save-location was specified opens the file again to write the stuff into the file.

It's still fragile because not advancing the iterator may cause the file not to be written but that was also the case before. However like this it won't leave files open.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,10 @@ New Features
 
 - Added ``sum`` option in method for ``combime``. [#500, #508]
 
+- Add ``_replace_with_whitespace`` argument for the ``files_filtered`` and
+  generator methods of ``ImageFileCollection`` to allow filtering for whitespace
+  seperated keys. [#539]
+
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -70,6 +70,10 @@ Bug Fixes
 
 - Suppress errors during WCS creation in CCDData.read(). [#552]
 
+- The generator methods in ``ImageFileCollection`` now don't leave open file
+  handles in case the iterator wasn't advanced or an exception was raised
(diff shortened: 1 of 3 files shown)


MSeifert04:
Test failures are because #541 was merged but if the `astropy.nddata.CCDData` is used the upstream patch isn't applied yet so the new test from #541 is failing.

MSeifert04:
I think I accidentally messed up this PR. I'm not sure how I can resolve this. The first two commits are rogues. The first is actually from another of my PRs and the second is from master...

bsipocz:
I suspect an interactive rebase should solve it. If not, you can always make a new branch from master (preferably rename this one locally first, and use the same name for the new branch), cherry-pick the relevant changes and force push it back here.

MSeifert04:
Ah, I could've overridden the branch just by pushing a new branch with the same name here? :anguished:

I cherry-picked the commits in a new branch and made a new PR because some changes propagated into the wrong commits and I didn't trust the interactive rebase (it showed some strange commit history after the first try).

bsipocz:
yes, you can basically force push anything, it doesn't even has to have the same name (it just makes it a bit clearer locally). What I regularly do when e.g. rebasing someone else's "featurebranch" is:
``git push -f someoneelsename HEAD:featurebrach``

Then, of course it can be another branch name rather than HEAD, etc as github doesn't track the name of your local branches.

MSeifert04:
Hah, I should've waited just a bit longer. That would have avoided having two PRs. I just tried it locally and it worked fine.

I added a bookmark to your comment and next time I won't duplicate the PR.

Thank you for the explanation. I really should invest some more time into understanding git. 😄

PR #546, opened 2017-08-30
Fix test failures because two previous PRs interfered with each other

MSeifert04:
Will still fail the astropy-dev test but only until https://github.com/astropy/astropy/pull/6489 is merged.


MSeifert04:
It's a trivial change and required to get the test suite passing again, so I'm merging this myself.

PR #547, opened 2017-08-30
Fix a FutureWarning from NumPy regarding non-dtype argument for issubdtype

MSeifert04:
I found the warning in a test log against NumPy master: https://travis-ci.org/astropy/ccdproc/jobs/270021313#L787

The correct form is to use `np.floating` as second argument.


MSeifert04:
It's a small and trivial change so I'm merging this myself.

Issue #395, opened 2016-09-08
ccmap


pllim:
Prompted by @saraogaz in [astropy-dev mailing list asking about ccmap equivalent](https://groups.google.com/forum/#!topic/astropy-dev/7OJSXlVWBWI), I am wondering if there is already something like that in `ccdproc`. It seems to me that this would be a logical place for such a thing to exist if it does at all.

`ccmap` documentation: http://stsdas.stsci.edu/cgi-bin/gethelp.cgi?ccmap

crawfordsm:
CCMAP itself hasn't been on our wish list for awhile but finding astrometric solutions to an image has been.  It's been long on the to-do list although some progress has been made on it in the astroquery package (although I think that PR was never completed).  

See #160  and also https://github.com/astropy/astroquery/pull/521

SaraOgaz:
Any updates on this?  Looks like the linked stuff hasn't been touched in a while.  I wanted to double check there weren't some other solutions in other places...

mwcraig:
Short answer, no. I'd *really* like astrometry.net to work in astroquery, so if I have time this week I'll work on that.

bsipocz:
Hmm, that PR in astroquery is very much abandoned. However, if anyone is keen on picking it up, it shouldn't be too far from being finished.

mwcraig:
Yeah, I've had it open in a browser tab for ~1.5 months. Looks very close to done.

Issue #160, opened 2014-08-05
 astrometric solution for the image


crawfordsm:
Calculate the astrometric solution for the images by comparing objects in the field to sources of known position

crawfordsm:
This could be taken care/integrating in with this pull request to astroquery:
https://github.com/astropy/astroquery/pull/521

crawfordsm:
Some useful other projects that should be bootstrapped to be used here:
https://github.com/spacetelescope/tweakwcs
https://github.com/astropy/astroquery/pull/1163

stscicrawford:
Another example would be https://github.com/toros-astro/astroalign, which is an astropy affiliated package.

Astrometry.net support has been added to astroquery:
https://astroquery.readthedocs.io/en/latest/astrometry_net/astrometry_net.html

PR #549, opened 2017-09-07
[WIP] Astrometry pull request

crawfordsm:
Here is some code that implements some basic astrometry for the images.  This includes some documentation for how to use it along with some further results.  I've checked it against several images and I have a set up to test it against a larger set of images, but it would be good to have someone else verify how it is working.  

I especially feel that there might be a more accurate way to calculate the WCS or a better way then what I am doing in here.  

I still need to add unit tests, but I think it might be ready for some other people to play with.  My feeling is this is not perfect right now, but I think it is useful.  

For new functionality:


crawfordsm:
Address #395 and #160.  May be of interest to @SaraOgaz @pllim.   

Also would be curious if @nden would have any comments on how to better do the WCS.

MSeifert04:
@crawfordsm I have several small suggestions, should I add them as comments or would you mind if I do a PR against your branch? That would avoid spamming comments.

MSeifert04:
Only for the smallish comments, for "not nitpicking comments" I will add comments.

crawfordsm:
Go ahead with the pull request -- it will be quicker than me fixing it.  I'm fine with that.

pllim:
Thanks, @crawfordsm ! I see some obvious typos and nitpicky things but I'll let @MSeifert04 do his PR first before commenting on those.

As for algorithm, I wonder if @mcara has anything beneficial that can be ported from `drizzlepac` here.

SaraOgaz:
Woot! I can test this out next week.

MSeifert04:
@pllim I'm too dumb to push to a PR so it's currently awaiting a merge in https://github.com/crawfordsm/ccdproc/pull/1. 😅

bsipocz:
@MSeifert04 - assuming you've added @crawfordsm's fork as remove with the name python crawfordsm and switched to this branch, and committed your changes, then:

``git push crawfordsm HEAD:astrom``

MSeifert04:
@bsipocz Thank you! That worked.

ccdproc/astrometry.py
```diff
@@ -0,0 +1,478 @@
...
+
+    Returns
+    -------
+    d : float or `numpy.ndarray`
+        Distance between x1,y1 and x2,y2
+    """
+    return ((x1-x2)**2 + (y1-y2)**2)**0.5
+
+
+def _get_index(z, n_stars):
```
MSeifert04: I haven't found a place where this function is actually used. Is there a reason to keep it?

docs/ccdproc/astrom_example.rst
```diff
@@ -0,0 +1,144 @@
...
+The function calculates a WCS based on a linear transformation of the
+coordinates.  This does not currently include any higher distortion terms
+and so for the most accurate results, any distortion should be already
+removed from the image.
+
+
+References
+----------
+
+.. [1]
```
MSeifert04: I was a bit too lazy to fill in the actual links. But we shouldn't forget to actually include them before we merge it.

ccdproc/astrometry.py
```diff
@@ -0,0 +1,478 @@
...
+        x-position of objects with duplicates removed
+
+    y : `numpy.ndarray`
+        y-position of objects with duplicates removed
+    """
+    mask = np.ones(len(x), dtype=bool)
+
+    for i in range(len(x)):
+        if mask[i]:
+            d = ((x - x[i])**2 + (y - y[i])**2)**0.5
```
MSeifert04: You could use the `distance` function here.

ccdproc/astrometry.py
```diff
@@ -0,0 +1,478 @@
...
+
+    y2 : float or `numpy.ndarray`
+        y-position of objects
+
+    Returns
+    -------
+    d : float or `numpy.ndarray`
+        Distance between x1,y1 and x2,y2
+    """
+    return ((x1-x2)**2 + (y1-y2)**2)**0.5
```
MSeifert04: You could use `np.hypot(x1-x2, y1-y2)` instead. Not sure if it's faster (probably not) but it makes the code easier to read. :)

ccdproc/astrometry.py
```diff
@@ -0,0 +1,478 @@
...
+    -------
+    ratio: float
+        Ratio between the distance from point i to j to the distance between i
+        and k
+
+    """
+    d1 = distance(x[i], y[i], x[j], y[j])
+    d2 = distance(x[i], y[i], x[k], y[k])
+    if d2 == 0:
+        ratio = np.nan
```
MSeifert04: Given that you don't return `ratio` you could just remove this branch?

ccdproc/astrometry.py
crawfordsm: will replace with np.hypot

ccdproc/astrometry.py
crawfordsm: Ah good catch, should probably just return np.nan which I think was suppose to happen.

ccdproc/astrometry.py
crawfordsm: Good point -- it is probably a bit more robust and would good to replace it.

ccdproc/astrometry.py
crawfordsm: Used in an earlier version of the code -- other than still might be useful, but if it really isn't needed in the end, it would be good to remove it.

MSeifert04: Just in case we do need it, I could provide you a solution that mathematically solves your problem instead of "by iteration". I was actually curious if it was easily possible and implemented it, but because the function wasn't actually used I put it on [StackOverflow](https://stackoverflow.com/a/46132585/5393381).

crawfordsm:
just cc'ing @cmccully in case he is interested in taking a look at this

SaraOgaz:
Okay, on further reflection I don't actually have any good data on hand for testing this.  But I did a quick comparsion with ccmap, and it looks like the create_wcs_from_fit definitely fits the bill for my needs, assuming there's a way to rewrite the CCDData.wcs back out to a FITS file when you do a `write()` on the CCDData object.  I can't quite tell from the doc page if this is the case.

nden:
@crawfordsm I've only looked at the WCS part and it looks reasonable.

crawfordsm:
@SaraOgaz -- that should be a case.  The WCS will update when the image is written out.   Thanks for checking on it

crawfordsm:
Thanks for the comments @mwcraig.   If you have a chance, can you send me the image?

mwcraig:
This link should work: https://www.dropbox.com/s/bjxpxumteaxcv4h/ey-uma-022R.fit?dl=0

mwcraig:
Also the notebook: https://gist.github.com/mwcraig/ac20da8ad3037f5ef1a0e08d9d907db1

I did have to modify astrometry.py a bit to get this to run...

ccdproc/astrometry.py
```diff
@@ -0,0 +1,489 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# This module implements the combiner class.
```
MSeifert04: Could you remove that line?

ccdproc/tests/test_astrometry.py
```diff
@@ -0,0 +1,155 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+# This module implements the base CCDData class.
```
MSeifert04: also this one.

crawfordsm:
Here's the script that I was using and seemed to work now with your data:
https://gist.github.com/crawfordsm/0dace3bec8cdae80e7aa3d2f9212956b

mwcraig:
Changed milestone to 2.0 (with target date of ~1 month) so there is more time to test this.

mwcraig:
Discussed at telecon on 4/5. Better approaches are available, may be migrated to reproject instead of ccdproc.

Keeping this open in case we decide to pick it up soon.

crawfordsm:
Thanks for all the reviews on this.  Closing this in light of other work that is being done.

PR #553, opened 2017-09-16
Fix RessourceWarnings in ImageFileCollection._generator

MSeifert04:

Supercedes #545 which I messed up (without having a backup branch 😅 )

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -820,29 +820,34 @@ class ImageFileCollection(object):
         ccd_kwargs = ccd_kwargs or {}
 
         for full_path in self._paths():
-            no_scale = do_not_scale_image_data
-            hdulist = fits.open(full_path,
-                                do_not_scale_image_data=no_scale)
+            add_kwargs = {'do_not_scale_image_data': do_not_scale_image_data}
+
+            # We need to open the file here, get the appropriate values and then
+            # close it again before it "yields" otherwise it's not garantueed
+            # that the generator actually advances and closes the file again.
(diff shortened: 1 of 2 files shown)


ccdproc/image_collection.py
```diff
@@ -820,29 +820,34 @@ def _generator(self, return_type,
...
+                return_thing = fits.getdata(full_path, self.ext, **add_kwargs)
+            elif return_type == 'ccd':
+                return_thing = fits_ccddata_reader(
+                    full_path, hdu=self.ext, **ccd_kwargs)
+            elif return_type == 'hdu':
+                with fits.open(full_path, **add_kwargs) as hdulist:
+                    ext_index = hdulist.index_of(self.ext)
+                    # Need to copy the HDU to prevent lazy loading problems
+                    # and "IO operations on closed file" errors
+                    return_thing = hdulist[ext_index].copy()
```
MSeifert04: Unfortunately this `.copy()` was the cleanest way I found to make it not throw "fileio on closed file". But if you have another suggestion that would be fine!

MSeifert04:
@mwcraig @crawfordsm Does this look okay?

ccdproc/image_collection.py
mwcraig: This is probably necessary because of lazy loading of FITS files....

MSeifert04: Not sure. I also suspect `memmap`. However after I realized that `.copy()` does the trick I stopped further investigation. I was just interested if you know a better way.

PR #566, opened 2017-10-19
Fix failures against numpy dev

MSeifert04:
Based on https://github.com/astropy/astropy/pull/6090. Let's just hope it's backported and the astropy 2.0.3 is released before NumPy 1.14 is, otherwise the `+FLOAT_CMP` won't do the "correct thing".


MSeifert04:
@mwcraig @crawfordsm Do you want to have a look. It currently makes the CI pass again and it's a very minor change.

crawfordsm:
Thanks for the fix -- at least the tests are working again

Issue #559, opened 2017-09-26
filtering ImageFileCollection with "-" in header keys


MSeifert04:
#539 allowed replacing characters with whitespaces but there are more characters that are often used in headers that can't be used for named arguments: `-` (and all other arithmetic operators but `-` is probably the most common one).

MSeifert04:
somewhat "fixed" by #580

PR #557, opened 2017-09-24
Enhance image collection.py

michaelwerger:
Allow search for keywords like OBS-TIME or EXP-TIME which are FITS compliant, examples:
print(collection.files_dictfiltered({'CCD-TEMP':'0.4'}))
or
print(collection.files_dictfiltered({'EXPTIME':'0.03', 'DATE-OBS':'27T14'}))

Please have a look at the following list and replace the "[ ]" with a "[x]" if
the answer to this question is yes.

- [x] For new contributors: Did you add yourself to the "Authors.rst" file?

For documentation changes:

- [ ] For documentation changes: Does your commit message include a "[skip ci]"?
      Note that it should not if you changed any examples!

For bugfixes:

- [ ] Did you add an entry to the "Changes.rst" file?
- [ ] Did you add a regression test?
- [ ] Does the commit message include a "Fixes #issue_number" (replace "issue_number").
- [ ] Does this PR add, rename, move or remove any existing functions or parameters?

For new functionality:

- [ ] Did you add an entry to the "Changes.rst" file?
- [x] Did you include a meaningful docstring with Parameters, Returns and Examples?
- [ ] Does the commit message include a "Fixes #issue_number" (replace "issue_number").
- [ ] Did you include tests for the new functionality?
- [ ] Does this PR add, rename, move or remove any existing functions or parameters?

Please note that the last point is not a requirement. It is meant as a check if
the pull request potentially breaks backwards-compatibility.

-----------------------------------------

PR #580, opened 2017-10-24
Remove replace_ argument for files_filtered again

MSeifert04:
The `replace_` argument was basically a good idea but as #559 and #557 show whitespaces are not the only problematic character. ``-`` is quite frequently used and sometimes even ``.``. Other invalid characters for keyword arguments are quite rare.

However it might be better to just document the workaround (passing in an unpacked dictionary) than actually inventing a functionality that tries to cover these cases in an easy to use way.

If we make `replace_` a dictionary for characters to translate that will be very awkward to use (because you could end up inventing different placeholders for different characters) and when you're passing in a dictionary already one might as well pass in the dictionary containing the key-value pairs.

Note that `replace_` hasn't been released yet so this doesn't need a deprecation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,9 +28,6 @@ New Features
 
 - Added ``sum`` option in method for ``combime``. [#500, #508]
 
-- Add ``replace_`` argument for the ``files_filtered`` and generator methods of
-  ``ImageFileCollection`` to allow filtering for whitespace seperated keys. [#539]
-
 
 Other Changes and Additions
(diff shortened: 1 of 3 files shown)


MSeifert04:
xref: #481

mwcraig:
Any thoughts on this, @bsipocz?

mwcraig:
Ideally we'd talk about this in the narrative documentation, but I think those need a much larger overhaul that makes sense as part of ccdproc 2.0, so I'm merging this as-is.

PR #581, opened 2017-10-24
Fix deprecations in tests with astropy development version

MSeifert04:
https://github.com/astropy/astropy/pull/6590 caused two deprecation warnings in our test suite but not related to actual code, just inside the test.

The reason is that composite-units like `u.electron / u.adu` create Quantities and casting them to `bool` currently raises a deprecation but it will be an Error in the future.


MSeifert04:
@mwcraig @crawfordsm This is a quite trivial change and the CI passed without deprecation warnings for astropy dev so I'll merge this this evening or tomorrow if you don't object.

PR #583, opened 2017-10-26
Catch warnings if no files are in the collection

MSeifert04:
It seems a bit odd that these are also emitted if one loads from an info-file but technically it's correct to emit them (and informational for the user) because the info-file doesn't populate the files attribute.


MSeifert04:
@mwcraig @crawfordsm This should be straight-forward given that I just catch the warnings in the tests (and the CI passed).  So if one of you wants to have a quick look over it (+merge if it looks alright) that would be great. :)

mwcraig:
thanks!

PR #584, opened 2017-10-27
Update flat_correct to allow passing in a normalization.

### LINKED ISSUE #577: flat_correct: allow passing in a normalization factor
flat_correct automatically calculates the mean of the flat field and normalizes the flat before dividing by it. This is convenient, but if you want to use two different flats on different sets of images but force them to have the same normalization, there is no way to turn off the normalization or pass in an externally determined factor.

This becomes an issue when processing multi-extension FITS files. For example, if extensions 1 and 2 are two different CCDs, or two halves of the same CCD, and one tries to process them by running flat_correct on the two extensions separately, then the flats need to have the same normalization. (Imagine that the left half of the CCD is extension 1 and it is 10% more sensitive than ext 2. Then both the science targets and the flat field will have 10% more counts in ext 1, and the target should be divided down by the brighter flat field. Renormalizing ext 1 and ext 2 separately by their own means will produce a non-flat science image.)

For an example of how I had to work around this behavior, you can see the function _renormalize_by_flat()_ in the pipeline [here](https://github.com/bjweiner/ARTN/blob/master/mont4k_pipeline/kuiper_reduce.py#L509). It would be simpler just to have a normalizing factor passed in as an optional argument to ccdproc.flat_correct. I can write this code and issue a pull request, but someone with more practice will need to read it for style and write a test.

This is a small part of making ccdproc friendly to multi-extension FITS files, but can stand on its own.

bjweiner:

Fixes #577

### DIFF
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -722,7 +722,8 @@ def gain_correct(ccd, gain, gain_unit=None):
 def flat_correct(ccd, flat, min_value=None, mean_value=None):
     """Correct the image for flat fielding.
 
-    The flat field image is normalized by its mean before flat correcting.
+    The flat field image is normalized by its mean or a user-supplied value
+    before flat correcting.
 
     Parameters
     ----------
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -719,7 +719,7 @@ def gain_correct(ccd, gain, gain_unit=None):
 
 
 @log_to_metadata
-def flat_correct(ccd, flat, min_value=None, mean_value=None):
+def flat_correct(ccd, flat, min_value=None, norm_value=None):
     """Correct the image for flat fielding.
 
     The flat field image is normalized by its mean or a user-supplied value
@@ -739,12 +739,11 @@ def flat_correct(ccd, flat, min_value=None, mean_value=None):
         will replace all values in the flat by the min_value.
         Default is ``None``.
 
-    mean_value : float or None, optional
-        Normalize flat field by this argument rather than the mean of the
-        image. This allows fixing several different flat fields to have the
-        same scale. If this value is negative or 0, a ``ValueError`` is
-        raised.
-        Default is ``None``.
+    norm_value : float or None, optional
+        If not ``None``, normalize flat field by this argument rather than the
+        mean of the image. This allows fixing several different flat fields to
+        have the same scale. If this value is negative or 0, a ``ValueError``
+        is raised. Default is ``None``.
 
     {log}
 
@@ -760,14 +759,14 @@ def flat_correct(ccd, flat, min_value=None, mean_value=None):
         flat_min.data[flat_min.data < min_value] = min_value
         use_flat = flat_min
 
-    # If a mean_value was input and is positive, use it to scale the flat
-    if mean_value is not None and mean_value > 0:
-        flat_mean_val = mean_value
-    elif mean_value is not None:
-        # mean_value was set to a bad value
-        raise ValueError('mean_value must be greater than zero.')
+    # If a norm_value was input and is positive, use it to scale the flat
+    if norm_value is not None and norm_value > 0:
+        flat_mean_val = norm_value
+    elif norm_value is not None:
+        # norm_value was set to a bad value
+        raise ValueError('norm_value must be greater than zero.')
     else:
-        # mean_value was not set, use mean of the image.
+        # norm_value was not set, use mean of the image.
         flat_mean_val = use_flat.data.mean()
 
     # Normalize the flat.
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -738,11 +738,12 @@ def flat_correct(ccd, flat, min_value=None, mean_value=None):
         minimum value is applied to the flat or specified by a float which
         will replace all values in the flat by the min_value.
         Default is ``None``.
-        
+
     mean_value : float or None, optional
-        Normalize flat field by this argument rather than the mean of the image.
-        This allows fixing several different flat fields to have the same scale.
-        If this value is negative or 0 it will be ignored.
+        Normalize flat field by this argument rather than the mean of the
+        image. This allows fixing several different flat fields to have the
+        same scale. If this value is negative or 0, a ``ValueError`` is
+        raised.
         Default is ``None``.
 
     {log}
@@ -758,13 +759,13 @@ def flat_correct(ccd, flat, min_value=None, mean_value=None):
         flat_min = flat.copy()
         flat_min.data[flat_min.data < min_value] = min_value
         use_flat = flat_min
-    
+
     # If a mean_value was input and is positive, use it to scale the flat
     if mean_value is not None and mean_value > 0:
         flat_mean_val = mean_value
     else:
         flat_mean_val = use_flat.data.mean()
-   
+
     # Normalize the flat.
     flat_mean = flat_mean_val * use_flat.unit
     flat_normed = use_flat.divide(flat_mean)
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -763,7 +763,11 @@ def flat_correct(ccd, flat, min_value=None, mean_value=None):
     # If a mean_value was input and is positive, use it to scale the flat
     if mean_value is not None and mean_value > 0:
         flat_mean_val = mean_value
+    elif mean_value is not None:
+        # mean_value was set to a bad value
+        raise ValueError('mean_value must be greater than zero.')
     else:
+        # mean_value was not set, use mean of the image.
         flat_mean_val = use_flat.data.mean()
 
     # Normalize the flat.
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -719,7 +719,7 @@ def gain_correct(ccd, gain, gain_unit=None):
 
 
 @log_to_metadata
-def flat_correct(ccd, flat, min_value=None):
+def flat_correct(ccd, flat, min_value=None, mean_value=None):
     """Correct the image for flat fielding.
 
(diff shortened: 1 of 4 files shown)


MSeifert04:
Thank you for making the pull request 👍 

I hope I have some time this weekend to look at it in more detail but regarding the test failure: There are some whitespaces in the "empty lines". Also I would prefer an Exception (ValueError) in case it's <= 0, silently ignoring invalid arguments.

mwcraig:
@bjweiner -- we are hoping to get a release of version 1.3 done otday or tomorrow and I really want to get this change in so I'm adding a test and addressing @MSeifert04's comments. I'll push those to your patch branch in ~15 minutes (and add you to the AUTHORS list). No need for further action on your part.

Normally we'd walk you through revising the PR but we're already ~3 weeks behind on the 1.3 release...

Thanks again for doing this!

mwcraig:
@MSeifert04 -- could you please give this a quick look?

ccdproc/core.py
```diff
@@ -736,6 +737,13 @@ def flat_correct(ccd, flat, min_value=None):
         will replace all values in the flat by the min_value.
         Default is ``None``.
 
+    mean_value : float or None, optional
+        Normalize flat field by this argument rather than the mean of the
```
MSeifert04: Probably should start with an `If not None, normalize flat ....`

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,47 @@ def test_flat_correct_min_value(ccd_data):
...
+                               flat.data / flat_mean)
+
+    # check that metadata is unchanged (since logging is turned off)
+    assert flat_data.header == ccd_data.header
+
+
+def test_flat_correct_mean_value_bad_value(ccd_data):
+    # Test that flat_correct raises the appropriate error if
+    # it is given a bad mean_value. Bad means <=0.
+    ccd_data.header['my_key'] = 42
```
MSeifert04: The key-value pair is not really essential for this test, right?

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,47 @@ def test_flat_correct_min_value(ccd_data):
...
+def test_flat_correct_mean_value(ccd_data):
+    # add metadata to header for a test below...
+    ccd_data.header['my_key'] = 42
+    size = ccd_data.shape[0]
+    # create the flat, with some scatter
+    # Note that mean value of flat is set below and is different than
+    # the mean of the flat data.
+    flat_mean = 5.0
+    data = np.random.normal(loc=1.0, scale=0.05, size=(size, size))
+    flat = CCDData(data, meta=fits.header.Header(), unit=ccd_data.unit)
```
MSeifert04: you can omit the `.header`, just using `fits.Header` should work without problems.

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,47 @@ def test_flat_correct_min_value(ccd_data):
...
+def test_flat_correct_mean_value_bad_value(ccd_data):
+    # Test that flat_correct raises the appropriate error if
+    # it is given a bad mean_value. Bad means <=0.
+    ccd_data.header['my_key'] = 42
+    size = ccd_data.shape[0]
+    # create the flat, with some scatter
+    data = np.random.normal(loc=1.0, scale=0.05, size=(size, size))
+    flat = CCDData(data, meta=fits.header.Header(), unit=ccd_data.unit)
+    with pytest.raises(ValueError) as e:
+        _ = flat_correct(ccd_data, flat, add_keyword=None,
```
MSeifert04: The `_ =` can be omitted.

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,47 @@ def test_flat_correct_min_value(ccd_data):
...
+
+
+def test_flat_correct_mean_value_bad_value(ccd_data):
+    # Test that flat_correct raises the appropriate error if
+    # it is given a bad mean_value. Bad means <=0.
+    ccd_data.header['my_key'] = 42
+    size = ccd_data.shape[0]
+    # create the flat, with some scatter
+    data = np.random.normal(loc=1.0, scale=0.05, size=(size, size))
+    flat = CCDData(data, meta=fits.header.Header(), unit=ccd_data.unit)
```
MSeifert04: Same here, it doesn't need the `.header`

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,47 @@ def test_flat_correct_min_value(ccd_data):
...
+    # create the flat, with some scatter
+    # Note that mean value of flat is set below and is different than
+    # the mean of the flat data.
+    flat_mean = 5.0
+    data = np.random.normal(loc=1.0, scale=0.05, size=(size, size))
+    flat = CCDData(data, meta=fits.header.Header(), unit=ccd_data.unit)
+    flat_data = flat_correct(ccd_data, flat, add_keyword=None,
+                             mean_value=flat_mean)
+
+    #check that the flat was normalized
```
MSeifert04: Should have a whitespace between `#` and `check ...`

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,47 @@ def test_flat_correct_min_value(ccd_data):
...
+@pytest.mark.data_scale(10)
+def test_flat_correct_mean_value(ccd_data):
+    # add metadata to header for a test below...
+    ccd_data.header['my_key'] = 42
+    size = ccd_data.shape[0]
+    # create the flat, with some scatter
+    # Note that mean value of flat is set below and is different than
+    # the mean of the flat data.
+    flat_mean = 5.0
+    data = np.random.normal(loc=1.0, scale=0.05, size=(size, size))
```
MSeifert04: Probably more robust if we use `size=ccd_data.shape` here (and in the other new test).

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,47 @@ def test_flat_correct_min_value(ccd_data):
     assert flat_orig_data is not flat.data
 
 
+# test for flat correction
```
MSeifert04: I guess this was copied from the test above. I don't think this accurately describes the test case, it probably needs a `with mean_value` appended or we could remove it here altogether.

MSeifert04:
I added some "nitpick-comments", feel free to ignore them :) A lot of these are also present in other tests, so that's no merge-blocker.

However I've been thinking about the argument name and `mean_value` seems a bit too specific. Wouldn't `norm_value` describe the purpose more accurately?

Also if you have the time a changelog entry would be good to have. But we can also add that when we do the next release. :)

mwcraig:
Addressed the review comments and rebased to resolve the changelog conflict.

@bjweiner -- I've changed the name of the argument from `mean_value` to `norm_value` but if you feel strongly that it should be changed back please let us know. We are hoping to do a release of 1.3 some time on Wednesday.

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,42 @@ def test_flat_correct_min_value(ccd_data):
...
+    # the mean of the flat frame.
+
+    # create the flat, with some scatter
+    # Note that mean value of flat is set below and is different than
+    # the mean of the flat data.
+    flat_mean = 5.0
+    data = np.random.normal(loc=1.0, scale=0.05, size=ccd_data.shape)
+    flat = CCDData(data, meta=fits.header, unit=ccd_data.unit)
+    flat_data = flat_correct(ccd_data, flat, add_keyword=None,
+                             mean_value=flat_mean)
```
MSeifert04: should be `norm_value`

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,42 @@ def test_flat_correct_min_value(ccd_data):
...
+def test_flat_correct_mean_value(ccd_data):
+    # Test flat correction with mean value that is different than
+    # the mean of the flat frame.
+
+    # create the flat, with some scatter
+    # Note that mean value of flat is set below and is different than
+    # the mean of the flat data.
+    flat_mean = 5.0
+    data = np.random.normal(loc=1.0, scale=0.05, size=ccd_data.shape)
+    flat = CCDData(data, meta=fits.header, unit=ccd_data.unit)
```
MSeifert04: sorry, it should be `fits.Header()`, not `fits.header`.

ccdproc/tests/test_ccdproc.py
```diff
@@ -481,6 +481,42 @@ def test_flat_correct_min_value(ccd_data):
...
+def test_flat_correct_mean_value_bad_value(ccd_data):
+    # Test that flat_correct raises the appropriate error if
+    # it is given a bad mean_value. Bad means <=0.
+
+    # create the flat, with some scatter
+    data = np.random.normal(loc=1.0, scale=0.05, size=ccd_data.shape)
+    flat = CCDData(data, meta=fits.header.Header(), unit=ccd_data.unit)
+    with pytest.raises(ValueError) as e:
+        flat_correct(ccd_data, flat, add_keyword=None,
+                     mean_value=-7)
```
MSeifert04: Should be `norm_value` now

mwcraig:
@MSeifert04 -- just a heads up that it will be another 4-5 hours before I can update this...feel free to push changes before then if you want.

MSeifert04:
@mwcraig The tests pass again and I think I replaced all remaining mean_value references with norm_value. I also added this to the 1.3 milestone and will merge it later if you or @crawfordsm don't object. :)

MSeifert04:
Thanks @bjweiner and @mwcraig!

PR #588, opened 2017-11-02
Remove pytest.ini in favor of setup.cfg

MSeifert04:
We currently use `pytest.ini` **and** the pytest section in `setup.cfg`. This PR removes the pytest.ini and puts the stuff in the setup.cfg.

Not sure if that makes sense though.


mwcraig:
This seems like a good move -- may as well keep all of the configuration in one place!

PR #598, opened 2017-12-01
Wrong parameter name in docstring

MSeifert04:
The parameter name is `extensions` not `extension`.


bsipocz:
docstring changes like this are not really tested atm (unlike intersphinx links, and headings), so you can confidently use [skip ci] in the future if you like.

MSeifert04:
You're right but I do feel better if I let the doc-build run whenever I change some of the docs. :)

PR #599, opened 2017-12-01
Speed up ImageFileCollection by using tolist more often

MSeifert04:
This speeds up several places where a Python iteration over a numpy.ndarray or astropy.table.Column is made. It will require slightly more memory but that is probably not noticeable for ImageFileCollections with less than several thousand files and my assumption is that ImageFileCollection will be very slow (probably unusable slow) for that many files anyway.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ New Features
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- Improved the performance of several ``ImageFileCollection`` methods. [#599]
+
 Bug Fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


MSeifert04:
I've used a data-set of 130 FITS files with a total of 321 different keywords (one of my recent datasets):

```
import ccdproc
%timeit ccdproc.ImageFileCollection(folder)
# 11.6 s ± 1.12 s per loop (mean ± std. dev. of 7 runs, 1 loop each)  NEW
# 21.3 s ± 135 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)  OLD
# Speedup: ~2

ifc = ccdproc.ImageFileCollection(folder)
%timeit ifc.values('naxis')
# 604 µs ± 25.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)  NEW
# 32.6 ms ± 735 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)  OLD
# Speedup: ~54
```

For the other functions the speedup isn't that much, for example `files_filtered` and `_paths` are only 1.3-1.5 times faster.

MSeifert04:
i'll add a small changelog later.

MSeifert04:
Added the changelog entry :)

PR #600, opened 2017-12-02
Pin Pytest version for astropy 1.3 travis build

MSeifert04:
This should fix the remaining test that errored out.


MSeifert04:
Merged this myself so the CI can run correctly for other PRs

PR #615, opened 2018-04-05
Update .travis.yml

### LINKED ISSUE #613: Drop one test case
Drop test case astropy1.0.x (LTS) and python 3.6, see #612

crawfordsm:
remove pre-astropy-2.0 from our test suite

closes #613

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,15 +61,6 @@ matrix:
           env: ASTROPY_VERSION=LTS
                SETUP_CMD='test --coverage'
 
-        # Check compatibility with the pre-LTS astropys
-        - python: 3.6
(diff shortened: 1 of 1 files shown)


bsipocz:
@crawfordsm - if you have commit rights, you can't use the github interface to edit and open a PR as it won't create the branch on your fork, but directly in the repo. This is wrong for two reasons, CI is running double and also the branch structure will be messier.

crawfordsm:
Thanks @bsipocz  -- I'll follow-up with a proper PR later to keep things clean.  But as a proof of concept, at least this worked expect the known current failure

bsipocz:
Sure, it's actually a bug in github I feel there should be an option to choose where that `patch` branch should go

PR #618, opened 2018-04-16
Astropy > 2.0 for rtd pip-requirements

MSeifert04:
I missed that in my earlier PR.


MSeifert04:
Should be uncontroversial so I'm merging this myself.

Issue #624, opened 2018-06-04
Sigma clipping is painfully slow


mwcraig:
I don't know what the underlying performance issue is, but wow it is slow (clipping a stack of 25 4k x 4k images takes minutes).

MSeifert04:
I also noticed that it's very slow. Thanks for bringing this up.

Could you provide a random-dataset for profiling? I could be interesting depending on the data-types and the strides of the dataset (the actual data is probably not important except it contains lots of unusual values like `NaN`s and so forth.

Also because we're re-using astropys sigma-clipping it would be handy to know what astropy version you were using.

One additional thing that could be a problem is RAM. Depending on the RAM and size of the arrays there may be lots of swapping going on.

Sorry for asking so much information from you but without having the specifics it's probably hard to profile it (correctly).

bsipocz:
This really brings up the issue that we need to add more benchmarks to `astropy-benchmarks`, currently there is nothing for the stats package. While the 4.5mins long test is not feasible, if it's not a local RAM issue I suspect it can be scaled back to be useful but still slow?

mwcraig:
Sure; I can also provide the data set I was working with if you want. It is just a bunch of dark frames. It is conceivable the bottle neck is in the combining, not the clipping, I suppose.

SaOgaz:
Probably related, @larrybradley is working on speeding up the astropy one: #7478

mwcraig:
Very related, thanks for mentioning that, @SaOgaz. And, of course, thanks to @larrybradley  for working on it!

mwcraig:
I haven't had a time to do much profiling, but here is an example of a set of calls that takes several minutes to run (for the combine part). Adding in some outlying pixels seems to help drive up the run time ( about 30 secs faster without that).

There are a number of things here that can lead to a speedup, I think. Even without sigma clipping, it takes close to a minute to combine the images.

Away from keyboard for a week, but will take another look when I'm back...

```python
import numpy as np
from astropy.io import fits
from ccdproc import combine
from astropy.stats import median_absolute_deviation

size = [4096, 4109]
n_images = 25

base_name = 'test-combine-{num:03d}.fits'

for num in range(n_images):
    data = np.random.normal(size=size)
    # Now add some outlying pixels so there is something to clip
    n_bad = 50000
    bad_x = np.random.randint(0, high=size[0] - 1, size=n_bad)
    bad_y = np.random.randint(0, high=size[1] - 1, size=n_bad)
[...]

PR #630, opened 2018-06-23
Refactor combiner.combine to minimize the number of files open at one time

### LINKED ISSUE #629: too many files open
This is the template for bug reports, if you have a feature request or question
you can safely ignore and delete this prefilled text.

Include a description of the problem: What are you trying to do (include your
code and the **full** traceback)? What did you expect?
It's openning images multiple times instead of using the same image that is already open, so it's openning the same set of images 14 times instead of using the same image each time.

```python traceback
INFO:astropy:splitting each image into 14 chunks to limit memory usage to 1000000000.0 bytes.
INFO: splitting each image into 14 chunks to limit memory usage to 1000000000.0 bytes. [ccdproc.combiner]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/miniconda3/envs/ast366/lib/python3.6/site-packages/astropy/utils/decorators.py in __get__(self, obj, owner)
    705         try:
--> 706             return obj.__dict__[self._key]
    707         except KeyError:

KeyError: 'data'

During handling of the above exception, another exception occurred:

OSError                                   Traceback (most recent call last)
~/miniconda3/envs/ast366/lib/python3.6/site-packages/reducer/gui.py in handler(b)
    563             self._go_button.disabled = True
    564             print(self)
--> 565             self.action()
    566 
    567             # change button should really only appear after the work is done.

~/miniconda3/envs/ast366/lib/python3.6/site-packages/reducer/astro_gui.py in action(self)
    471                 ("Processing {} of {} "
    472                  "(may take several minutes)".format(idx + 1, n_groups))
--> 473             combined = self._action_for_one_group(combo_group)
    474             name_addons = ['_'.join([str(k), str(v)])
    475                            for k, v in six.iteritems(combo_group)]

~/miniconda3/envs/ast366/lib/python3.6/site-packages/reducer/astro_gui.py in _action_for_one_group(self, filter_dict)
    521         combined = ccdproc.combine(file_list,
    522                                    mem_limit=DEFAULT_MEMORY_LIMIT,
--> 523                                    **combine_keyword_args)
    524 
    525         sample_image = ccdproc.CCDData.read(file_list[0])

~/miniconda3/envs/ast366/lib/python3.6/site-packages/ccdproc/combiner.py in combine(img_list, output_file, method, weights, scale, mem_limit, clip_extrema, nlow, nhigh, minmax_clip, minmax_clip_min, minmax_clip_max, sigma_clip, sigma_clip_low_thresh, sigma_clip_high_thresh, sigma_clip_func, sigma_clip_dev_func, dtype, combine_uncertainty_function, **ccdkwargs)
    758                     imgccd = image
    759                 else:
--> 760                     imgccd = CCDData.read(image, **ccdkwargs)
    761 
    762                 # Trim image

~/miniconda3/envs/ast366/lib/python3.6/site-packages/astropy/nddata/mixins/ndio.py in read(cls, *args, **kwargs)
     25         formats.
     26         """
---> 27         return io_registry.read(cls, *args, **kwargs)
     28 
     29     def write(self, *args, **kwargs):

~/miniconda3/envs/ast366/lib/python3.6/site-packages/astropy/io/registry.py in read(cls, format, *args, **kwargs)
    515 
    516         reader = get_reader(format, cls)
--> 517         data = reader(*args, **kwargs)
    518 
    519         if not isinstance(data, cls):

~/miniconda3/envs/ast366/lib/python3.6/site-packages/astropy/nddata/ccddata.py in fits_ccddata_reader(filename, hdu, unit, hdu_uncertainty, hdu_mask, hdu_flags, **kwd)
    499         # search for the first instance with data if
    500         # the primary header is empty.
--> 501         if hdu == 0 and hdus[hdu].data is None:
    502             for i in range(len(hdus)):
    503                 if hdus.fileinfo(i)['datSpan'] > 0:

~/miniconda3/envs/ast366/lib/python3.6/site-packages/astropy/utils/decorators.py in __get__(self, obj, owner)
    706             return obj.__dict__[self._key]
    707         except KeyError:
--> 708             val = self.fget(obj)
    709             obj.__dict__[self._key] = val
    710             return val

~/miniconda3/envs/ast366/lib/python3.6/site-packages/astropy/io/fits/hdu/image.py in data(self)
    241             return
    242 
--> 243         data = self._get_scaled_image_data(self._data_offset, self.shape)
    244         self._update_header_scale_info(data.dtype)
    245 

~/miniconda3/envs/ast366/lib/python3.6/site-packages/astropy/io/fits/hdu/image.py in _get_scaled_image_data(self, offset, shape)
    707         code = BITPIX2DTYPE[self._orig_bitpix]
    708 
--> 709         raw_data = self._get_raw_data(shape, code, offset)
    710         raw_data.dtype = raw_data.dtype.newbyteorder('>')
    711 

~/miniconda3/envs/ast366/lib/python3.6/site-packages/astropy/io/fits/hdu/base.py in _get_raw_data(self, shape, code, offset)
    473                               offset=offset)
    474         elif self._file:
--> 475             return self._file.readarray(offset=offset, dtype=code, shape=shape)
    476         else:
    477             return None

~/miniconda3/envs/ast366/lib/python3.6/site-packages/astropy/io/fits/file.py in readarray(self, size, offset, dtype, shape)
    293                     # file)
    294                     memmap = Memmap(self._file, mode=MEMMAP_MODES[self.mode],
--> 295                                     dtype=np.uint8)
    296 
    297                     # Now we immediately discard the memmap array; we are

~/miniconda3/envs/ast366/lib/python3.6/site-packages/numpy/core/memmap.py in __new__(subtype, filename, dtype, mode, offset, shape, order)
    277         elif (hasattr(filename, "name") and
    278               isinstance(filename.name, basestring)):
--> 279             self.filename = os.path.abspath(filename.name)
    280         # same as memmap copies (e.g. memmap + 1)
    281         else:

~/miniconda3/envs/ast366/lib/python3.6/posixpath.py in abspath(path)
    374             cwd = os.getcwdb()
    375         else:
--> 376             cwd = os.getcwd()
    377         path = join(cwd, path)
    378     return normpath(path)

OSError: [Errno 24] Too many open files
```

Don't forget to include the version of astropy, ccdproc and numpy, just copy
this into your Python interpreter (without the backticks):

```
import astropy
print(astropy.__version__)
import ccdproc
print(ccdproc.__version__)
import numpy
print(numpy.__version__)
```
astropy verson - 3.0.1
ccdproc version - 1.3.0
numpy version - 1.14.2

mwcraig:
This will fix #629; the initial commit is two new tests, one of which will fail before the issue is fixed. Once the failures are verified on CI, I'll add a fix.

For bugfixes:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,8 @@ Bug Fixes
 - Function ``median_combine`` now correctly calculates the uncertainty for
   masked ``CCDData``. [#608]
 
+- Function ``combine`` now avoids opening files unnecessarily. [#629, #630]
+
 1.3.0 (2017-11-1)
 -----------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,7 +24,7 @@ Bug Fixes
 - Function ``median_combine`` now correctly calculates the uncertainty for
   masked ``CCDData``. [#608]
 
-- Function ``combine`` now avoids opening files unnecessarily. [#629, #630]
+- Function ``combine`` avoids keeping files open unnecessarily. [#629, #630]
 
 1.3.0 (2017-11-1)
 -----------------
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -804,7 +804,9 @@ def combine(img_list, output_file=None,
             if ccd.uncertainty is not None:
                 ccd.uncertainty.array[x:xend, y:yend] = comb_tile.uncertainty.array
             # Clean up any open files
-            del comb_tile, tile_combiner, ccd_list
+            del comb_tile
+            del tile_combiner
+            del ccd_list
 
     # Write fits file if filename was provided
     if output_file is not None:
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -740,46 +740,24 @@ def combine(img_list, output_file=None,
             'func': sigma_clip_func,
             'dev_func': sigma_clip_dev_func}
 
-    # These set up for the case that all of the images we are opening
-    # can be memmapped. In that case we only need to open the images
-    # ONCE.
-    memmap_failed = False
-    memmapped_images = []
-
-    # Finally Run the input method on all the subsections of the image
+   # Finally Run the input method on all the subsections of the image
     # and write final stitched image to ccd
     for x in range(0, xs, xstep):
         for y in range(0, ys, ystep):
             xend, yend = min(xs, x + xstep), min(ys, y + ystep)
             ccd_list = []
-            if not memmapped_images:
-                # Either we haven't checked whether we can memmap
-                # or we tried and it failed.
-                for image in img_list:
-                    if isinstance(image, CCDData):
-                        imgccd = image
-                        # One of the images is already a CCDData object, no
-                        # point in trying to memmap the rest
-                        memmap_failed = True
-                    else:
-                        imgccd = CCDData.read(image, **ccdkwargs)
-                        if not memmap_failed:
-                            # Apparently we haven't check whether we  can
-                            # memmap.
-                            if isinstance(imgccd.data.base, mmap):
-                                memmapped_images.append(imgccd)
-                            else:
-                                # This image isn't a memmap, so we can't
-                                # take this shortcut. We'll just open files
-                                # each time.
-                                memmap_failed = True
-                                memmapped_images = []
-                    # Trim image
-                    ccd_list.append(imgccd[x:xend, y:yend])
-            else:
-                # We've already opened the images and can use the memmap
-                for image in memmapped_images:
-                    ccd_list.append(image[x:xend, y:yend])
+            for image in img_list:
+                if isinstance(image, CCDData):
+                    imgccd = image
+                else:
+                    imgccd = CCDData.read(image, **ccdkwargs)
+
+                # Trim image and copy
+                # The copy is *essential* to avoid having a bunch
+                # of unused file references around if the files
+                # are memory-mapped. See this PR for details
+                # https://github.com/astropy/ccdproc/pull/630
+                ccd_list.append(imgccd[x:xend, y:yend].copy())
 
             # Create Combiner for tile
             tile_combiner = Combiner(ccd_list, dtype=dtype)
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 """This module implements the combiner class."""
+from mmap import mmap
 
 import numpy as np
 from numpy import ma
@@ -739,23 +740,50 @@ def combine(img_list, output_file=None,
             'func': sigma_clip_func,
             'dev_func': sigma_clip_dev_func}
 
+    # These set up for the case that all of the images we are opening
+    # can be memmapped. In that case we only need to open the images
+    # ONCE.
+    memmap_failed = False
+    memmapped_images = []
+
     # Finally Run the input method on all the subsections of the image
     # and write final stitched image to ccd
     for x in range(0, xs, xstep):
         for y in range(0, ys, ystep):
             xend, yend = min(xs, x + xstep), min(ys, y + ystep)
             ccd_list = []
-            for image in img_list:
-                if isinstance(image, CCDData):
-                    imgccd = image
-                else:
-                    imgccd = CCDData.read(image, **ccdkwargs)
-
-                # Trim image
-                ccd_list.append(imgccd[x:xend, y:yend])
+            if not memmapped_images:
+                # Either we haven't checked whether we can memmap
+                # or we tried and it failed.
+                for image in img_list:
+                    if isinstance(image, CCDData):
+                        imgccd = image
+                        # One of the images is already a CCDData object, no
+                        # point in trying to memmap the rest
+                        memmap_failed = True
+                    else:
+                        imgccd = CCDData.read(image, **ccdkwargs)
+                        if not memmap_failed:
+                            # Apparently we haven't check whether we  can
+                            # memmap.
+                            if isinstance(imgccd.data.base, mmap):
+                                memmapped_images.append(imgccd)
+                            else:
+                                # This image isn't a memmap, so we can't
+                                # take this shortcut. We'll just open files
+                                # each time.
+                                memmap_failed = True
+                                memmapped_images = []
+                    # Trim image
+                    ccd_list.append(imgccd[x:xend, y:yend])
+            else:
+                # We've already opened the images and can use the memmap
+                for image in memmapped_images:
+                    ccd_list.append(image[x:xend, y:yend])
 
             # Create Combiner for tile
             tile_combiner = Combiner(ccd_list, dtype=dtype)
+
             # Set all properties and call all methods
             for to_set in to_set_in_combiner:
                 setattr(tile_combiner, to_set, to_set_in_combiner[to_set])
@@ -775,6 +803,8 @@ def combine(img_list, output_file=None,
                 ccd.mask[x:xend, y:yend] = comb_tile.mask
             if ccd.uncertainty is not None:
                 ccd.uncertainty.array[x:xend, y:yend] = comb_tile.uncertainty.array
+            # Clean up any open files
+            del comb_tile, tile_combiner, ccd_list
 
     # Write fits file if filename was provided
     if output_file is not None:
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 """This module implements the combiner class."""
+
 import numpy as np
 from numpy import ma
 from .core import sigma_func
@@ -738,7 +739,7 @@ def combine(img_list, output_file=None,
             'func': sigma_clip_func,
             'dev_func': sigma_clip_dev_func}
 
-   # Finally Run the input method on all the subsections of the image
+    # Finally Run the input method on all the subsections of the image
     # and write final stitched image to ccd
     for x in range(0, xs, xstep):
         for y in range(0, ys, ystep):
@@ -779,7 +780,7 @@ def combine(img_list, output_file=None,
                 ccd.mask[x:xend, y:yend] = comb_tile.mask
             if ccd.uncertainty is not None:
                 ccd.uncertainty.array[x:xend, y:yend] = comb_tile.uncertainty.array
-            # Clean up any open files
+            # Free up memory to try to stay under user's limit
             del comb_tile
             del tile_combiner
             del ccd_list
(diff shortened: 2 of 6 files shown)


mwcraig:
Tests are now (mostly) failing as intended; removed an f-string to fix fails on python 3.4 and 3.5.

mwcraig:
Not sure why coverage went down; the lines marked as not covered are actually covered by the new test. Perhaps because the test is run in a subprocess.

mwcraig:
@MSeifert04 @crawfordsm -- this is ready for a look.

MSeifert04:
Interesting. Reading through the code we really have some more problems there. In either case passing `memmap=False` will lead to trouble because we (1) had slices that prevented the original from "deallocating" or with the new approach we (2) explicitly keep them alive. That makes the "memory-saving" completely wrong if one chose to do `memmap=False` (or one overwrote the astropy fits memmap config entry). 

I like the fact that images are read only once with the new approach (at least if all of them are actually files) even though the time reading the files will probably be negligible compared to the combining time - I mean the "tiling" only happens when we expect that each "tiled combine" actually exceeds a few GB. So I don't think it's actually necessary to "keep them around". The best idea here would be to read the file, copy the slice and then completely dispose of the CCDData (which means the memory is freed and the file handle is disposed). That way we (probably) keep only one file-handle open at each point (in case of memory mapping) and don't keep the original alive because we sliced a real array (in case of non-memory-mapping). The copy is actually a bit "annoying", however we have to assume that the copy operation will also be negligible compared to actually combining the images. I mean we already copy the arrays inside the combiner - if we could do that step in `combine` instead of `Combiner.__init__` in that case we would actually don't have any additional overhead compared to the status-quo.

Long story short: I'm not really sure how we should proceed. In fact we could simply use your approach for now (looks good - except that the tests are not picked up, which should be corrected) and think about a more general refactor later on - I mean most of the things I've just written are pure speculation (especially those statements that contained the word "negligible") and have to be verified before we actually do that.

mwcraig:
>  except that the tests are not picked up, which should be corrected

How should that be fixed?

mwcraig:
Thanks for the comments, @MSeifert04. 

I think that as currently written there are a few different outcomes:

### Input list is `CCDData` objects

No change from prior behavior, but also no real memory savings, since all of the CCDData objects are already in memory. This might not strictly be the case if all of the CCDData objects are actually memory mapped, I suppose.

### Input list is filenames and all files are opened memmap=True and all files are memory-mappable (e.g. no unsigned int or other bzero/bscale)

Files are opened only once, memmap references are preserved for later use. This is probably the optimal way to go in this case.

### Input list is filenames and one more more files is opened with memmap=False or cannot be memory-mapped

As soon as one of those files is hit, `memmap_failed` is set to `True`. After that point, each file is opened, which means it is read into memory, a slice is made. As far as I can tell, keeping the slice around actually keeps around a copy of the full array (code snippet below)...so unless we copy, this case might read everything into memory. I think this would have been true before the PR too...

code snippet:

```python
foo = np.zeros([10, 100])
foo.size  # This is 1000
bar = foo[3:5, 30:40]
del foo
bar.size   # 20, as expected
bar.base.size   # 1000, so it looks like we still have a reference to foo...
bar.base[7:10, 80:90]  # Yep, we can index this no problem 
```
#### So we need to handle this case better...

I think that in this case what you proposed (slice the CCDData and force a copy of the data) is necessary to avoid having all of the data read into memory. I might try writing a test that fails in this case (i.e. uses way too much memory) to confirm this is the case, then push a fix.

MSeifert04:
>>  except that the tests are not picked up, which should be corrected
> How should that be fixed? By which I mean "I don't know how to force the coverage to pick up the tests that are run in a subprocess but would be happy to make that happen if I knew how" 😀

Ah, maybe by avoiding "subprocess" here? Would that be feasible?

MSeifert04:
> ### Input list is filenames and all files are opened memmap=True and all files are memory-mappable (e.g. no unsigned int or other bzero/bscale)
> Files are opened only once, memmap references are preserved for later use. This is probably the optimal way to go in this case.

Not sure if it's ideal. I mean you keep a lot of open file-handles that could still conflict with the open-files-limit just to avoid opening the same file multiple times. I haven't done any timings but my feeling is that opening the files multiple times isn't really significant in the runtime of the function.

MSeifert04:
> As far as I can tell, keeping the slice around actually keeps around a copy of the full array (code snippet below)...so unless we copy, this case might read everything into memory.

Yes, that's the problem. We either keep the memmap alive (if it's memmappable) or the full array (in case it's not). Both isn't great.

mwcraig:
>>> except that the tests are not picked up, which should be corrected

>> How should that be fixed? By which I mean "I don't know how to force the coverage to pick up the tests that are run in a subprocess but would be happy to make that happen if I knew how" 😀

> Ah, maybe by avoiding "subprocess" here? Would that be feasible?

No, because the tests lower the limit (in Python) on the number of simultaneously open files, and once the limit is lowered, you cannot raise it.

mwcraig:
Going to run some profiles now to inform the discussion here. IIRC, the bulk of the time is spent reading data from disk (at least in the case that 100 files are being combined)...

MSeifert04:
> Going to run some profiles now to inform the discussion here. IIRC, the bulk of the time is spent reading data from disk (at least in the case that 100 files are being combined)...

That would be great. Could you share the results? :)

mwcraig:
@MSeifert04 -- see https://github.com/astropy/ccdproc/issues/624#issuecomment-411625196 for one graph. I think the profiling/optimization is separate from this file issue, though. I was being thrown by  `__getitem__` taking so much time; I assumed that was from io.fits but it is from numpy.ma.

I'll add more comments on that issue tomorrow, but it looks like the easy speedups are going to be using the nan functions instead of masked, and maybe allowing for bottleneck.

mwcraig:
A number of comments:

+ My original approach here (making a list of memory-mapped files) is terrible. It has the effect of breaking the memory limit that `combine` is trying to impose. The reason is that as successive slices are read in the memory-mapped array is "filled in" with the values read until each is taking up its full size in memory.
    - For 14 files, 2k x 2k, with memory limit of 32MB, the maximum memory use in my keep-the-list-of-memmaps approach used 450MB by the end.
    - With the latest version the maximum memory use is about 128MB. That is still too high but I believe it is unrelated to the changes here. Will open a separate issue for it.
+ The amount of time spent (re)opening files and copying in the latest approach is small. There was less than a 1 second difference between the latest approach and the earlier one of keeping a list of the memmaped files. In any event, the time spent copying is *tiny* (about 1 second) compared to the time spent in `sigma_clipping` (about 26 seconds) and `average_combine`ing (about 6 seconds).

I have one more minor change to push then I think this is ready to merge. Will add a cross-reference to the memory issue shortly.

mwcraig:
Incidentally, to do the memory profiling, I installed [memory_profiler](https://pypi.org/project/memory_profiler/) and ran it in `ccdproc/tests` with:

```
mprof run run_with_file_number_limit.py --overhead 6 --kind fits --open-by combine-chunk --size 2000 --frequent-gc 50
```

mwcraig:
Memory use issue is #638

CHANGES.rst
```diff
@@ -24,6 +24,8 @@ Bug Fixes
 - Function ``median_combine`` now correctly calculates the uncertainty for
   masked ``CCDData``. [#608]
 
+- Function ``combine`` now avoids opening files unnecessarily. [#629, #630]
```
MSeifert04: The new approach actually doesn't avoid opening the files unnecessarily.

ccdproc/combiner.py
```diff
@@ -1,7 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 """This module implements the combiner class."""
-
```
MSeifert04: Could be kept. I like blank lines between module docs and imports.

ccdproc/combiner.py
```diff
@@ -739,7 +738,7 @@ def combine(img_list, output_file=None,
             'func': sigma_clip_func,
             'dev_func': sigma_clip_dev_func}
 
-    # Finally Run the input method on all the subsections of the image
+   # Finally Run the input method on all the subsections of the image
```
MSeifert04: That seems wrong. I would've expected this to throw an IndentationError... But I think we should stick to 4 space indentation even for comments.

ccdproc/combiner.py
```diff
@@ -775,6 +779,10 @@ def combine(img_list, output_file=None,
                 ccd.mask[x:xend, y:yend] = comb_tile.mask
             if ccd.uncertainty is not None:
                 ccd.uncertainty.array[x:xend, y:yend] = comb_tile.uncertainty.array
+            # Clean up any open files
```
MSeifert04: The comment is now wrong. At this point there should only be no open file anymore

ccdproc/combiner.py
```diff
@@ -775,6 +779,10 @@ def combine(img_list, output_file=None,
...
+            del comb_tile
```
MSeifert04: Not sure if I really like those `del`s. Are these still necessary for reducing the memory?

MSeifert04 (commented):
I haven't looked through the tests. It looks very good overall, just a few comments that need to be addressed (and some that are more optional).

CHANGES.rst
mwcraig: I think it still does; without the `copy`, `ccd_list` held a file reference to each file in the list if the files were mem-mapped. The `copy` in the update avoids that, so that at most 1 file is open at a time.

mwcraig: I suppose it would be more accurate to say that it doesn't keep files open unnecessarily....

ccdproc/combiner.py
mwcraig: Oops, that was unintentional...

ccdproc/combiner.py
mwcraig: Good catch! Haven't figured why my editor sometimes misses things like this....

mwcraig: [Turns out my *editor* did catch it...my eyes missed it]

ccdproc/combiner.py
mwcraig: Updating....

ccdproc/combiner.py
mwcraig: I think so; can run a profile in a moment to check. The only *essential* one, I think is to delete `tile_combiner`. That will, by design, use roughly the amount of memory the user has set as the cap. If we don't delete it, then on the second pass through the loop we'll (briefly) consume twice that memory on line 761 (or 2 depending on commit) as we construct the new combiner.

mwcraig:
Thanks for the comments @MSeifert04 -- I have addressed them in code or in the comments above.

MSeifert04 (approved):
Still haven't got the time to review the tests but the actual code changes look good to me 👍

CHANGES.rst
MSeifert04: > I suppose it would be more accurate to say that it doesn't keep files open unnecessarily....

I like that.

PR #632, opened 2018-07-07
Test: Use a numba-powered function for median sigma clipping

MSeifert04:
Just a test (see also https://github.com/astropy/ccdproc/issues/624#issuecomment-403227184), it's not meant to be merged 😄 

Related to #624

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
         # to repeat them for all configurations.
         - NUMPY_VERSION=stable
         - ASTROPY_VERSION=stable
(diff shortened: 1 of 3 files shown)


MSeifert04:
For the example in https://github.com/astropy/ccdproc/issues/624#issuecomment-396072627 the new version takes 1min 24s while the old version takes 4min 25s. Not sure if the difference justifies the special-casing and the numba dependency 🤷‍♂️

MSeifert04:
@mwcraig @crawfordsm Independent of this PR, what's your general opinion regarding numba? Should we avoid it and rather use Cython/C extensions or would numba be okay (as optional dependency)?

PR #633, opened 2018-07-08
Unify scale handling in the combine methods and avoid multiplying by 1 if none is given

MSeifert04:
This unifys the `scale` handling in the `*combine` methods by putting it in a separate method. It also simply returns the `data_arr` in case there is no scaling. This has a potentially huge advantage because in the common case that there is no scaling it doesn't do a copy. That copy could be several GB big (because it's a copy of the stacked ccds and the mask) so avoiding it when possible at least saves a bit of memory.

However I noticed that we don't use the scale for the uncertainty calculation, that probably warrants an own issue (or pull request). I tried to keep it to a simple refactor for this PR.

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -302,6 +302,13 @@ class Combiner(object):
             mask = (self.data_arr - baseline > high_thresh * dev)
             self.data_arr.mask[mask] = True
 
+    def _get_scaled_data(self, scale_arg):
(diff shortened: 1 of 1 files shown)


MSeifert04:
@mwcraig @crawfordsm Does this look good?

crawfordsm:
I’ll try to take a look at this later this week.

PR #640, opened 2018-08-12
Fix calculation of `ystep` in `combine`

mwcraig:
The first commit here adds a regression test; I haven't pushed the fix yet so I expect it to fail.

I moved calculation of the steps into a function of its own because it made testing much simpler.

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -496,6 +496,23 @@ class Combiner(object):
         return combined_image
 
 
+def _calculate_step_sizes(x_size, y_size, num_chunks):
+    """
+    Calculate the strides in x and y to achieve at least
+    the ``num_chunks`` pieces.
+
+    Parameters
+    ----------
+    """
+    # First we try to split only along fast x axis
+    xstep = max(1, int(x_size / num_chunks))
+
+    # If more chunks need to be created we split in Y axis for remaining number
+    # of chunks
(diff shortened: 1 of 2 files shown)


mwcraig:
This is ready for a look, assuming tests pass

MSeifert04:
It could make sense to **only** slice along the non-contiguous dimension to only get contigous "chunks". That would avoid memory fragmentation and (somewhat) unnecessary mmap-calls (at least if memory-mapping is used). That would also simplify the logic because we only need to check which axis should be sliced and how often, that way it would be (relatively) easy to also support ND images, not just 2D. I guess the case where slicing along one axis in 1-steps (the lower limit) would exceed the memory limit is probably very rare?!

mwcraig:
I'm guessing the cases where slicing along more than 1 axis needs to happen are exceedingly rare (unless you do all of your data reduction on a raspberry pi or some maybe?).

it is tempting to have numpy reshape the data into a 1d array always; then the chunking is very similar. The result could be built up as a 1d array then reshaped to the correct shape at the end.

I ran out of time/enthusiasm though....

MSeifert04:
> I ran out of time/enthusiasm though....

If you want - I can give it a try. :)

mwcraig:
Feel free!

crawfordsm:
Let's merge as is -- we can revisit this as part of the combine re-factor later if needed.

PR #644, opened 2018-08-20
Change in image_collection.py to support for .fz files

### LINKED ISSUE #643: Support for fpack (.fz) files in ImageFileCollection
When trying to use `ImageFileCollection` on `fpack` files (i.e. extension `.fz`), e.g. :

```files_list = ImageFileCollection(location=directory, keywords=['EXPTIME']).files_filtered(EXPTIME=1.0)```

it does not read the headers and instead returns an empty table/list,  unless you manually specify the names with the `filenames` parameter.  However, I did a little change and added `'.fz'` to the list of compressed file extensions on line 735 of `image_collection.py` (i.e. within the method `_fits_files_in_directory`), and additionally used `ext=1` in `ImageFileCollection`, and it worked perfectly filtering the files.

I did notice your comment about not supporting `.fz`, but it was not totally clear the reason why it did not. Thus, I can put in a PR for the change, since `ImageFileCollection` seems to work fine with it,  and this could be a potential enhancement to the functionality of `ccdproc`.

seap-jaime:
This PR is a change for the support of fpack files, and fixes #643 . That issue has all the descriptions for this little change in `image_collection.py`

- [yes] Does the commit message include a "Fixes #643 " (replace "issue_number").
- [] Does this PR add, rename, move or remove any existing functions or parameters?
---------------------------------------

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -732,7 +732,7 @@ class ImageFileCollection(object):
 
         # The common compressed fits image .fz not supported.
         if compressed:
-            for comp in ['.gz', '.bz2', '.Z', '.zip']:
+            for comp in ['.gz', '.bz2', '.Z', '.zip', '.fz']:
                 with_comp = [extension + comp for extension in full_extensions]
                 full_extensions.extend(with_comp)
 


mwcraig:
@seap-jaime -- thanks so much for opening the issue and the PR! I'd like to add a test of this functionality before merging. I could add one myself to your PR in the next day or two, or, if you would prefer, you could work on a test. 

I don't know that we had a good reason for *not* supporting tile compressed files...I just think we never got around to it. 

Thanks again for opening the PR and let me know how you would like to proceed!

mwcraig:
@seap-jaime -- excellent! let me know if you need any help.

AnthonyHorton:
The second link isn't actually making an fpack file, it's a test to seeing if a gzipped FITS file with the wrong extension (`.fz`) still gets properly identified as a gzip file.

Real fpack `.fz` files differ from gzipped FITS files in an important way: only the data is compressed, not the whole file.  This makes it much faster to filter them by their header contents because it's not necessary to uncompress the file order to read the header, as it is with gzipped FITS files. That's why full fpack support is really good for `ImageFileCollection`, if you've got lots of FITS files to sort and filter you really want them to be fpacked instead of gzipped.

It really does look like the only thing preventing full fpack support is that `.fz` isn't included as a recognised compressed file extension to use when search a directory for FITS files. If you work around that by passing the actual filenames to `ImageFileCollection()` everything works as intended. The only gotcha is that for `.fz` files the actual header and data ends up one extension to the right, i.e. you have to specify `ext=1` to work with the primary extension. That could cause trouble when working with a mixture of fpacked and non-fpacked FITS files.

MSeifert04:
@AnthonyHorton Thank you for the explanation!

> It really does look like the only thing preventing full fpack support is that .fz isn't included as a recognised compressed file extension to use when searching a directory for FITS files.

In that case we might just add it now and see if it breaks something :)

seap-jaime:
Hi @mwcraig and @MSeifert04,

I've successfully added the test to support for fpack (`.fz`) files. I created the `.fz` file in the `test_dir` of `pytest_fixtures.py`, i.e. inside the `triage_setup` fixture, so it is just temporary not to add any additional large file. When you run `pytest test_image_collection.py` all its 70 tests are passed (including the one I added), filtering the file by exposure time, so now `ImageFileCollection` is reading the headers of this type of files :).

I modified: `pytest_fixtures.py` , `test_image_collection` and `setup.cfg`. In the latter was just a little change, in the line L10: [pytest] -----> [tool:pytest] , since the former was deprecated when running the tests. All the other tests also work if you do `python setup.py test` from the top level directory `ccdproc`.

Hope everything is fine, please let me know any suggestion or change I should do.

seap-jaime:
Finally after two more commits (white spaces fixed) most checks were completed! :The only one that is not working is the Travis CI Build, it is something about NUMPY_VERSION=1.13, but I haven't changed anything of it in the code. Not sure how to make this check works though

AnthonyHorton:
The failed test seems to be a Travis glitch (it hung during startup). Need someone with admin to force it to try again.

mwcraig:
@seap-jaime @AnthonyHorton -- Just restarted the broken build. Thanks again for your work on this!

One last task -- can you please add @seap-jaime to the list of contributors at `AUTHORS.rst`? 

@MSeifert04 -- thanks for helping out with the review of this!

seap-jaime:
@mwcraig -- Just added @seap-jaime to `AUTHORS.rst` 

It was a pleasure helping!

PR #653, opened 2018-12-07
Update README.rst

### LINKED ISSUE #556: Documentation notes
Here are some documentation  notes/areas that could use improvement:

- the reduction tool box is given as a series of tasks but it might be useful to organize it so it is more of a tutorial
- our users may not be coming from a numpy/python background so expanding on ideas or proving better links to concepts like indices, pass by referencing, and masking may be helpful
- in README add links to indicate other places that might be useful to ask questions like stackoverflow, python and astronomy facebook group, and astropy mailing list.
- Add link to numfocus for donations

crawfordsm:
Fixes #556

This does not address the issue of indexing but there is another issue with that open.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,8 @@ in this source distribution.
 Documentation is at `ccdproc.readthedocs.io
 <http://ccdproc.readthedocs.io/en/latest/>`_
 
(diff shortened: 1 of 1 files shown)


crawfordsm:
Kill tests as I should have added [skip CI]

PR #670, opened 2019-06-19
Fixes #634

### LINKED ISSUE #634: Request: Passing 1D array of weights to average_combine
The default scaling function for average_combine (numpy.ma.average) can take a list of weights, but average_combine requires the weights to be 2D arrays. Can average_combined be changed to allow 1D arrays of weights?

Here's a link to the numpy.ma.average doc for convenience: https://docs.scipy.org/doc/numpy/reference/generated/numpy.ma.average.html

-Matt

hlabathems:

For new functionality:


ccdproc/tests/test_combiner.py
```diff
@@ -100,6 +100,16 @@ def test_weights_shape(ccd_data):
...
         c.weights = ccd_data.data
 
+def test_1Dweights():
+    ccd_list = [CCDData(np.zeros((10, 10)), unit=u.adu),
+                CCDData(np.zeros((10, 10)) - 1000, unit=u.adu),
+                CCDData(np.zeros((10, 10)) + 1000, unit=u.adu)]
+        
+    c = Combiner(ccd_list)
+    c.weights = np.array([1, 5, 10])
+    ccd = c.average_combine()
```
mwcraig: So far this ensures that the new weights succeed but we should also check the result. It looks like the result should be 312.5 for the data and weights given, so something like this should do the trick:

```python
assert all(ccd.data == 312.5)
```

If that fails it is probably because of rounding, so you may need to do this instead:

```python
np.testing.assert_almost_equal(ccd.data, 312.5)
```

ccdproc/tests/test_combiner.py
```diff
@@ -100,6 +100,16 @@ def test_weights_shape(ccd_data):
...
+
+
```
MSeifert04: Sorry for the nitpicky comment: We're mostly using PEP8 with 2 empty lines between functions. That means it's one empty line too few before the function but one too much after. Could you fix that?

MSeifert04:
Hey @hlabathems, thank you for the pull request.

This definetely needs and entry in the changelog and since this is your first contribution you can also add yourself to the contributor page!

There is one problem with the `axis` parameter in the setter (I left an inline comment) that needs to be resolved (somehow) before we can think about merging this. I'm not exactly sure what the best API would be in this case. A code sample how one could use the `axis` for the weights would be very helpful (for my understanding and as test case).

ccdproc/combiner.py
```diff
@@ -124,14 +124,15 @@ def weights(self):
         return self._weights
 
     @weights.setter
-    def weights(self, value):
+    def weights(self, value, axis=0):
```
MSeifert04: This probably won't work as expected. The `property.setter` doesn't allow passing in multiple arguments. How should this be "invoked"?

We also need a test that covers non-default axis, just to make sure it will continue to work in the future.

hlabathems: @MSeifert04 I will give this some thought

ccdproc/tests/test_combiner.py
MSeifert04: Please don't "resolve conversation" when they are in fact not resolved (either in code or by comment).

ccdproc/combiner.py
```diff
@@ -124,14 +124,15 @@ def weights(self):
...
             if isinstance(value, np.ndarray):
-                if value.shape == self.data_arr.data.shape:
-                    self._weights = value
-                else:
-                    raise ValueError(
-                        "dimensions of weights do not match data.")
+                if value.shape != self.data_arr.data.shape:
+                    if value.ndim != 1:
+                        raise ValueError("1D weights expected when shapes of the data and weights differ.")
+                    if value.shape[0] != self.data_arr.data.shape[axis]:
```
crawfordsm: I think axis can be set to 0 here and removed as a parameter.  

There will need to be a test case for this added as well and then this should be ready

PR #678, opened 2019-07-26
Rewrite pytest fixture for generating ImageFileCollection test directory

### LINKED ISSUE #468: Test suite of ImageFileCollection uses non-existing header keywords
While fixing #461 I noticed that the exceptions came because the keywords passed to the file collection and the ones present in the fits-file-headers differ. That lead to a number of subtle bugs (and a lot of warnings when manually executing them).

Maybe the tests (including those in the narrative documentation) should be updated to use actual present header pairs. It wouldn't hurt to have some testing what happens when non-existing keywords are requested but the majoritiy of the test suite shouldn't depend on it.

mwcraig:
This PR makes the set up of the `ImageFileCollection` tests clearer, and I think helps resolve #468 by vastly limiting the number of cases in which there are missing entries in the summary table.


mwcraig:
Self merging this one so that I can use it as the base for the rest of the `ImageFileCollection` PRs coming in later today.

PR #680, opened 2019-07-26
WIP lots of fixes for ImageFileCollection

mwcraig:
This PR still has a few commits incoming to close a couple more issues, but we'll see if the tests run.

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -87,11 +87,14 @@ class ImageFileCollection(object):
             glob_include = str(glob_include)
         self._glob_include = glob_include
 
-        self._location = location
+        if location is not None:
+            self._location = location
+        else:
+            self._location = ''
+
         self._filenames = filenames
         self._files = []
-        if location:
-            self._files = self._get_files()
+        self._files = self._get_files()
 
         if self._files == []:
             warnings.warn("no FITS files in the collection.",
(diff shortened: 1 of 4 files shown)


ccdproc/image_collection.py
```diff
@@ -738,8 +718,16 @@ def _fits_files_in_directory(self, extensions=None,
 
         all_files = listdir(self.location)
         files = []
-        for extension in full_extensions:
-            files.extend(fnmatch.filter(all_files, '*' + extension))
+        if not self._find_fits_by_reading:
+            for extension in full_extensions:
+                files.extend(fnmatch.filter(all_files, '*' + extension))
+        else:
+            for file in all_files:
```
crawfordsm: can you change to infile as file is protected

ccdproc/image_collection.py
```diff
@@ -74,23 +72,20 @@ class ImageFileCollection(object):
         easily select subsets of files in the target directory.
         Default is ``None``.
 
-     ext: str or int, optional
-         The extension from which the header and data will be read in all files.
-         Default is ``0``.
+    ext: str or int, optional
```
crawfordsm: nitpick:  since there is a default value, is it really optional?

docs/ccdproc/image_management.rst
```diff
@@ -31,13 +31,26 @@ collection to use all keywords in the headers::
...
 expressions. For example, to include all filenames that begin with ``1d_`` but
 not ones that include the word ``bad``, you could do::
 
     >>> ic_all = ImageFileCollection('.', glob_include='1d_*',
     ...                              glob_exclude='*bad*')
 
+Alternatively, you can create the collection with an explicit list of file names::
+
+    >>> ic_names = ImageFileCollection(filenames=['a.fits', 'b.fits.gz'])
+
```
crawfordsm: Do you want to include paths in here to show they are possible?  Or just a sentence indicating that?

ccdproc/image_collection.py
mwcraig: I think it is optional in the sense that you do not need to include it in the initializer because the default value will work well for most cases.

crawfordsm (approved):

Issue #561, opened 2017-10-02
transform_image wcs update


SaraOgaz:
Sorry if this is a silly question, but where in transform_image does the wcs get updated after the transformation is done?  I don't see anything in the code that looks like it's doing this, but maybe that's somewhere else in the data class? This isn't really high priority or anything, I was just curious.

http://ccdproc.readthedocs.io/en/latest/_modules/ccdproc/core.html#transform_image

Thanks!

mwcraig:
> Sorry if this is a silly question

I sort of wish it was, because the answer is that it does not transform the WCS at all. I'm not quite sure how to properly fix that, but at the very least we should warn/strip out the WCS if we cannot transform it.

SaraOgaz:
After conversing with @pllim, it *might* be a straightforward update to the CRPIX... keywords if there is no extra distortion keywords in the WCS solution.  Does that sound right @nden?

MSeifert04:
That reminds me: We probably should document which attributes are processed in the functions. There are several functions that simply copy or skip over certain attributes. For example `median_filter` ignores everything except `data` and only a handful functions respect the `wcs`.

For `transform_image` it's probably impossible to update the WCS because we cannot know what `transform_func` is used. But maybe we could add more specialized functions for "shift" or "rotate". Could be useful to know in what context you used `transform_image`. :)

SaraOgaz:
@MSeifert04 I was maybe going to use it as an example replacement for the IRAF shift functions. I don't remember if I've gotten any specific requests for it outside of some questions about it on the astropy mailing list about... a year ago I think?

mwcraig:
I think we should throw a warning if there is a WCS....

crawfordsm:
a warning will be added but we should come back to this after modeling and gwcs is updated to see if this can be made to work as a user would expect

PR #684, opened 2019-07-28
added warning to transform_image that it does not apply to WCS

crawfordsm:
This adds a warning to `transform_images` that if a WCS is present, the transform is not applied to that WCS.

### DIFF
--- a/ccdproc/tests/test_ccdproc.py
+++ b/ccdproc/tests/test_ccdproc.py
@@ -580,7 +580,7 @@ def test_catch_transform_wcs_warning(ccd_data):
         return 10 * arr
 
     with catch_warnings() as w:
-         tran = transform_image(ccd_data, tran)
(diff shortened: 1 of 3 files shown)


crawfordsm:
Partially addresses #561

ccdproc/tests/test_ccdproc.py
```diff
@@ -573,6 +573,15 @@ def test_transform_isfunc(ccd_data):
...
         transform_image(ccd_data, 1)
 
+#test warning is issue if WCS information is available
+def test_catch_transform_wcs_warning(ccd_data):
+
+    def tran(arr):
+        return 10 * arr
+
+    with catch_warnings() as w:
+        tran = transform_image(ccd_data, tran)
```
mwcraig: Should you check the content of the warning here too?

mwcraig:
This looks good otherwise 👍

PR #688, opened 2019-07-28
added better negative number handling to create_deviation

### LINKED ISSUE #573: How should create_deviation handle negative values
While investigating #572 I've seen that the create_deviation function creates `nan`s where the data*gain + readnoise**2 is negative. That makes sense but currently the warning (from NumPy) just mentions "invalid value encountered in sqrt". I'm not sure if we shouldn't error out or provide a more informative warning in case (some of) the data is negative. Or if we should handle the pixels where the data is negative differently.

crawfordsm:
Added an option to disregard negative values passed to ``create_deviation``and assume the error is represented by the read noise.  If not and there are negative values are in the array, a more informative warning is created.

Closes #573

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,9 @@ New Features
 - Add option to use regular expression matching when filtering items in
   ``ImageFileCollection``. [#480, #595, #682]
 
+- Added an option to disregard negative values passed to ``create_deviation``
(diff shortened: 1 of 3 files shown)


mwcraig:
Looks like a couple minor spacing issues need to be fixed...

PR #690, opened 2019-07-31
Add filter method to ImageFileCollection

### LINKED ISSUE #596: files_filtered question (ImageFileCollection)
Tangentially related to #595, the way I see most people use hselect is to return the `table` style output after filtering by some keyword values.  I can't see from the documentation, is there an easy way to take the return list of filenames from `files_filtered` and return an updated table based on that filter? It seems like if I were going to use built in `Table` methods to do this I would have the user skip straight to filtering the first table return, and not use `files_filtered` at all.  But maybe this was the intention all along?

I guess what I'm really asking is, are there plans to write a table_filtered method, and if not, what's the reasoning to having a 'files_filtered' and not a `table_filtered'?  Just thinking out loud here, apologies if these seems overly critical.

mwcraig:
This fixes #596 by adding a `filter` method.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@ New Features
 - Added an option to disregard negative values passed to ``create_deviation``
   and assume the error is represented by the read noise [#688]
 
+- Add ``filter`` method to ``ImageFileCollection`` that creates a new
+  collection by filtering based on header keywords. [#596, #690]
+
 Other Changes and Additions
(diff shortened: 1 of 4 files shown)


mwcraig:
Anyone object to me merging this? The one failure is due to astropy/astropy#9075

mwcraig:
forgot to mention @crawfordsm @MSeifert04

PR #694, opened 2019-08-03
Add example of processing a MEF file and a little tinkering with section headings

### LINKED ISSUE #673: Preserving WCS with CCDProc and MEF files
I'm trying to preserve the WCS on some MEF files from KPNO using CCDProc and CCDData objects. (Trying to preserve it so after I flat correct I can combine the data into a single extension for photometry)

```python
file = 'flat2.fits'
flat = fits.getdata(file)
flat =CCDData(flat, unit=u.electron)
im = fits.getdata('n209767.fits')
im = CCDData(im, unit=u.electron)
reduced = ccdproc.flat_correct(im,flat)
reduced.write('data.fits')
```

I'm using Python 3.7.0, Astropy 3.1.2, CCDProc 1.3.0.post1, and Numpy 1.16.2

mwcraig:
This fixes #673 by adding an example in the documentation of processing a multi-extension fits file. 

While I was looking at the documentation again I fleshed out some of the section headings and rearranged a couple to make it easier for users to find what they are looking for.

@katherinebruce529, can you take a quick look? The rendered documentation is at http://physics.mnstate.edu/craig/apy20/ 

@MSeifert04 or @crawfordsm, if you could also take a look that would be great.


docs/reduction_toolbox.rst
```diff
@@ -428,6 +432,112 @@ which should be removed rather than propagated:
...
+    ...     """
+    ...     with NumpyRNGContext(1234):
+    ...         number_of_image_extensions = 3
+    ...         science_image = [fits.PrimaryHDU()]
+    ...         flat_image = [fits.PrimaryHDU()]
+    ...         for _ in range(number_of_image_extensions):
+    ...             # Simulate a cloudy night, average pixel
+    ...             # value of 100 with a read_noise of 1 electron.
+    ...             # Image size is 150 × 150.
+    ...             data = np.random.normal(100., 1.0, [150, 150])
```
MSeifert04: Is 150 really needed? In the interest of the test-runs and so on we could also use 10 x 10, right?

That saves a bit of memory and time.

docs/reduction_toolbox.rst
```diff
@@ -428,6 +432,112 @@ which should be removed rather than propagated:
...
+    >>> science_mef = 'science.fits'
+    >>> flat_mef = 'flat.fits'
+    >>>
+    >>> make_sample_mef(science_mef, flat_mef)
+    >>>
+    >>> # CALIBRATION EXAMPLE STARTS HERE
+    >>> # (everything above this was to make sample files)
+    >>>
+    >>> # Read our sample images
+    >>> science = fits.open(science_mef)
```
MSeifert04: This will leave the files open and should really be avoided in official "examples". One could use the double `with` here:

```
with fits.open(science_mef) as science, fits.open(flat_mef) as flat:
```

MSeifert04: Maybe even use `memmap=False` because we're in the global scope so the file-handle will be kept alive even if the context closes (in a function the dropping of the reference would finally close the file).

MSeifert04:
Thanks Matt, I think the example (and cleanup is very impressive).

However I wonder if we shouldn't drop the "creation"-part of the example and just provide the two files via the repo because it somewhat distracts (especially because of the length) from the actual CCDData and ccdproc related part.

docs/reduction_toolbox.rst
```diff
@@ -428,6 +432,112 @@ which should be removed rather than propagated:
...
 
 
+Working with multi-extension FITS image files
+---------------------------------------------
+
+Multi-extension FITS (MEF) image files cannot be processed natively in ``ccdproc``. The example below illustrates how to `~ccdproc.flat_correct` all of the extensions in a MEF and write out the calibrated file as a MEF. Applying other reduction steps would be similar.
+
+The example begins by creating small MEF "science" image and a sample flat:
+
+    >>> import numpy as np
```
MSeifert04: Since we're not actually showing output of any operation can't we drop the `>>>` and `...`?

docs/reduction_toolbox.rst
```diff
@@ -428,6 +432,112 @@ which should be removed rather than propagated:
     plt.show()
 
 
+Working with multi-extension FITS image files
+---------------------------------------------
+
+Multi-extension FITS (MEF) image files cannot be processed natively in ``ccdproc``. The example below illustrates how to `~ccdproc.flat_correct` all of the extensions in a MEF and write out the calibrated file as a MEF. Applying other reduction steps would be similar.
```
MSeifert04: This is a very long line, could you add some newlines? I'm not sure about the complete doc but generally 80 or 100 chars per line is a good (but soft) maximum.

mwcraig:
> However I wonder if we shouldn't drop the "creation"-part of the example and just provide the two files via the repo because it somewhat distracts (especially because of the length) from the actual CCDData and ccdproc related part.

Good idea; I'll move the MEF-generating code into a function in the `tests` directory in case we want it in the future.

Addressing the other comments on the next flight 😀

docs/reduction_toolbox.rst
mwcraig: I'd like to keep it as a doctest, though, to ensure that the code actually runs.

docs/reduction_toolbox.rst
mwcraig: I added explicit `.close` calls at the end of the example. I agree it is bad to leave them open. I think in this case `memmap=False` is a bad idea because may well copy/paste the code. `with` is probably preferable but added another level of indentation.

docs/reduction_toolbox.rst
mwcraig: Changed

docs/reduction_toolbox.rst
MSeifert04: > I think in this case `memmap=False` is a bad idea because may well copy/paste the code.

The memmap is actually *because* people may copy/paste the code. Closing the memmap'd file just closes the file and doesn't close the memmap if the variables referencing the memmap aren't disposed of.

> `with` is probably preferable but added another level of indentation.

I agree but I think that's a worth-while level of indentation :)

docs/reduction_toolbox.rst
MSeifert04: Ah, the code isn't doc-tested if it doesn't have the `>>>`?

MSeifert04 (approved):
This looks good now.

The other comments can also be changed if we wanted later.

PR #702, opened 2019-09-26
update docs for lacosmic sepmed parameter [docs only]

### LINKED ISSUE #631: cosmicray_lacosmic sepmed defaults to True
Though it is not very clear, `cosmicray_lacosmic`'s [doc](http://ccdproc.readthedocs.io/en/latest/api/ccdproc.cosmicray_lacosmic.html#ccdproc.cosmicray_lacosmic) seems to say that it's default parameter values are set such that the default values for L.A. Cosmic by van Dokkum.

But to be so, the parameter `sepmed` should be `False`, since the original L.A. Cosmic does not have seperate median method.

jradavenport:

For documentation changes:

   **My commit message used: [docs only], which I assume also behaves like [skip ci]**

For bugfixes:

For new functionality:

Fixes #631 by clarifying the `sepmed` parameter is for performance only, and not part of the L.A. Cosmic algorithm natively. 

Also corrected typo of "lacosmic" to "L.A. Cosmic", per the original paper.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,7 @@ New Features
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+- Clarified ``sepmed`` default in ``cosmicray_lacosmic`` docs. [#631]
 
 Bug Fixes
 ^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,7 @@ New Features
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
-- Clarified ``sepmed`` default in ``cosmicray_lacosmic`` docs. [#631]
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -18,6 +18,7 @@ New Features
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+- Clarified ``sepmed`` default in ``cosmicray_lacosmic`` docs. [#702]
```
bsipocz: no need for changelog for docs only changes.

bsipocz:
Welcome @jradavenport!

bsipocz:
> My commit message used: [docs only], which I assume also behaves like [skip ci]

Indeed, it's similar. It's a "home grown" option that only works with travis, so here it works as intended.

MSeifert04:
@jradavenport Thank you for the PR 👍 

Could you remove the changelog entry - Then I think this can be merged.

mwcraig:
Thanks @jradavenpor, merging!

PR #706, opened 2019-11-19
Fixing header plugin

bsipocz:
This causes an error that overshadows failures with v4.0rc1 testing

Issue #707, opened 2019-11-20
Failures with astropy v4.0rc1


bsipocz:
These are relevant looking failures. Please report back during the RC testing period whether they are issues  here only, or upstream in astropy. 

```
================================================================ test session starts ================================================================
platform darwin -- Python 3.7.4, pytest-5.2.1, py-1.8.0, pluggy-0.13.0

Running tests with ccdproc version 2.1.dev1562.
Running tests in ccdproc docs.

Date: 2019-11-20T15:40:39

Platform: Darwin-17.7.0-x86_64-i386-64bit

Executable: /usr/local/opt/python/bin/python3.7

Full Python Version: 
3.7.4 (default, Jul  9 2019, 18:15:00) 
[Clang 10.0.0 (clang-1000.11.45.5)]

encodings: sys: utf-8, locale: UTF-8, filesystem: utf-8
byteorder: little
float info: dig: 15, mant_dig: 15

Package versions: 
Numpy: 1.17.3
Scipy: 1.3.1
Matplotlib: 3.1.1.post2570+g70fb1a6b9
Pandas: 0.25.3
Astropy: 4.0rc1
astroscrappy: 1.0.8
reproject: 0.5.1
astropy-helpers: 4.0.rc1

Using Astropy options: remote_data: any.

rootdir: /private/var/folders/dc/hsm7tqpx2d57n7vb3k1l81xw0000gq/T/ccdproc-test-5_dbbnua/lib/python3.7/site-packages, inifile: setup.cfg
plugins: astropy-header-0.1.1, forked-1.0.2, arraydiff-0.3, dependency-0.4.0, xonsh-0.9.13.dev1, remotedata-0.3.2, openfiles-0.4.0, asdf-2.4.2, doctestplus-0.5.0, cov-2.8.1
collected 313 items                                                                                                                                 

ccdproc/combiner.py .                                                                                                                         [  0%]
ccdproc/core.py ....ss...                                                                                                                     [  3%]
ccdproc/extern/bitfield.py ..                                                                                                                 [  3%]
ccdproc/tests/test_bitfield.py ............                                                                                                   [  7%]
ccdproc/tests/test_ccdmask.py ...                                                                                                             [  8%]
ccdproc/tests/test_ccdproc.py ......................9
F.........................FFF....................                                         [ 31%]
ccdproc/tests/test_ccdproc_logging.py .......                                                                                                 [ 33%]
[...]

PR #710, opened 2019-11-24
Fix CI problems

MSeifert04:
It seems like we shouldn't use `python:` but instead use the `PYTHON_VERSION` variable.

I also incremented the NumPy versions (no particular reason except that we were missing NumPy 1.15 and I think that would be more appropriate than NumPy 1.13).

Closes #706 

Ref: #707

### DIFF
--- a/ccdproc/conftest.py
+++ b/ccdproc/conftest.py
@@ -6,13 +6,30 @@
 
 import os
 
-from pytest_astropy_header.display import (PYTEST_HEADER_MODULES,
-                                           TESTED_VERSIONS)
+try:
+    # When the pytest_astropy_header package is installed
+    from pytest_astropy_header.display import (PYTEST_HEADER_MODULES,
+                                               TESTED_VERSIONS)
 
-def pytest_configure(config):
-    config.option.astropy_header = True
+    def pytest_configure(config):
+        config.option.astropy_header = True
+except ImportError:
+    # TODO: Remove this when astropy 2.x and 3.x support is dropped.
+    # Probably an old pytest-astropy package where the pytest_astropy_header
+    # is not a dependency.
+    try:
+        from astropy.tests.plugins.display import (pytest_report_header,
+                                                   PYTEST_HEADER_MODULES,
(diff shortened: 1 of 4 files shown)


MSeifert04:
There's one problem that needs fixing upstream (https://github.com/astropy/astropy/issues/9664).

mwcraig:
This looks good except for the one we can't fix yet...

ccdproc/conftest.py
```diff
@@ -7,16 +7,29 @@
...
 try:
-    from astropy.tests.plugins.display import (pytest_report_header,
-                                               PYTEST_HEADER_MODULES,
+    # When the pytest_astropy_header package is installed
+    from pytest_astropy_header.display import (PYTEST_HEADER_MODULES,
                                                TESTED_VERSIONS)
+
+    def pytest_configure(config):
+        config.option.astropy_header = True
 except ImportError:
```
bsipocz: I think we should just add `pytest-astropy-header` as a hard dependency  rather than doing this nested try except (it should ideally be picked up from pytest-astropy, but clearly something is not working as it should).

MSeifert04: > I think we should just add pytest-astropy-header as a hard dependency

Yeah, I also thought about this. But the test-log shows that it's in fact installed ([corresponding log line from your PR](https://travis-ci.org/astropy/ccdproc/jobs/614284643#L964)) but not picked up makes it probably a lot harder to solve than just adding a hard-dependency.

I would prefer the nested try-catch for now, it's only astropy-2.x after all and we may even drop support for it once the new astropy LTS is out.

MSeifert04:
Should this stay open until the upstream is fixed? 

Otherwise I would temporarily set the astropy-dev test to allow-fail and put this in.

bsipocz:
I would vote for merging this (CI on master is already broken, this at least fixes some issues).

MSeifert04:
@mwcraig I'm going to merge this tomorrow if you won't object. The astropy-dev job is then (temporarily) allowed to fail.

Issue #705, opened 2019-11-05
ccdproc.cosmicray_lacosmic has undocumented side effect of gain correcting CCDData


AnthonyHorton:
Calling `ccdproc.cosmicray_lacosmic()` on a `CCDData` object that has not been gain corrected (i.e. units of ADU) has the side effect of scaling the image data by the `gain` value passed to `ccdproc.cosmicray_lacosmic`. The units of the returned `CCDData` do not get changed, however, and remain as ADU even though the data is now in electrons.

The fact that the data has been gain corrected during the cosmic ray removal does not get logged in the header either, but then neither does the cosmic ray removal itself.

This side effect is undocumented and probably unintended? My expectation was that when calling `cosmicray_lacosmic` on a non-gain corrected `CCDData` and passing in a `gain` value the data would be converted to electrons for the cosmic ray removal (which happens), and then be converted back to ADU for the returned `CCDData` (which doesn't happen). 

Minimal example:
```
In [1]: import numpy as np                                                                                                             

In [2]: import ccdproc                                                                                                                 

In [3]: c1 = ccdproc.CCDData(np.arange(1,10).reshape((3, 3)), unit='adu')                                                              

In [4]: c1                                                                                                                             
Out[4]: 
CCDData([[1, 2, 3],
         [4, 5, 6],
         [7, 8, 9]])

In [5]: c1.unit                                                                                                                        
Out[5]: Unit("adu")

In [6]: c2 = ccdproc.cosmicray_lacosmic(c1, gain=2.0)                                                                                  

In [7]: c2                                                                                                                             
Out[7]: 
CCDData([[ 2.,  4.,  6.],
         [ 8., 10., 12.],
         [14., 16., 18.]])

In [8]: c2.unit                                                                                                                        
Out[8]: Unit("adu")
```

If the gain correction is intended behaviour then it needs to also change the units, needs to log to the header, and needs to be documented.

```
astropy 3.2.3
ccdproc 2.0.1
numpy 1.17.3
```

mwcraig:
Thanks for reporting this with the short example; we should do a bug fix release for this fairly quickly. I will try to get to it this weekend. 

I think the correct behavior here should be to not modify the data.

mwcraig:
Closed by #716

PR #712, opened 2019-12-10
Do not gain correct data when identifying cosmic rays

mwcraig:
Not sure this is the best fix, but it will get the job done.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,4 @@
-2.1.0 (Unreleased)
+2.0.2 (Unreleased)
 ------------------
 
 New Features
@@ -9,6 +9,8 @@ Other Changes and Additions
 
 Bug Fixes
 ^^^^^^^^^
+
+- Do not apply gain when using LACOSMIC to detect cosmic rays. [#712, #705]
 
 2.0.1 (2019-09-05)
 ------------------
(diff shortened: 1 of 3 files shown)


MSeifert04:
Do we really want to change the output of a function so drastically in a bug-fix release?

Couldn't we add an "apply_gain" parameter to the function which defaults to True (previous behavior) which can be set to False if the uncorrected image is needed (with your patch)? That way we don't break any code that uses this function and we could (gradually) change the default in the future?

MSeifert04:
Otherwise the change/test look great. Thanks for tackling this Matt.

mwcraig:
> Do we really want to change the output of a function so drastically in a bug-fix release?

I guess it depends on whether we think it is a bug (in which case a bug fix makes sense) or an undocumented feature (in which case we'd be breaking the API).

I'll discuss with @crawfordsm today.

mwcraig:
A couple of comments while I’m thinking about them:

+ There is a unit issue in the current behavior — as pointed out in #705, the gain is applied but the units of the data are not updated. 
+ The gain cannot have have units in our current implementation because astroscrappy doesn’t handle them. We could handle units in our wrapper. 

I’m increasingly leaning towards a feature release to fix these things because the current API has issues.

MSeifert04:
> I’m increasingly leaning towards a feature release to fix these things because the current API has issues.

I like that idea. 

However there are some additional issues:

- How do we handle gain if an array is passed in instead of a `CCDData`.
- Do we make gain-correction a documented feature of the function, or an optional feature, or do we remove it completely?
- How do we handle units? Normally plain data is in units of ADU and gain is in e-/ADU (or ADU/e-). However what do we do if one of these assumptions is wrong?

mwcraig:
@crawfordsm and I talked about this extensively a little while ago. I'm going to close this and open a different PR shortly that outlines a more complete approach.

PR #716, opened 2019-12-12
Update interface to cosmic ray functions

mwcraig:
This will end up being a much longer PR, but starts by fixing #705

### DIFF
--- a/docs/reduction_toolbox.rst
+++ b/docs/reduction_toolbox.rst
@@ -101,10 +101,9 @@ Use this technique with `~ccdproc.cosmicray_lacosmic`:
 
 .. note::
 
-    A significant error in `~ccdproc.cosmicray_lacosmic` in version 2.1 of
-    ccdproc. By default, `~ccdproc.cosmicray_lacosmic` multiplies the image by
+    By default, `~ccdproc.cosmicray_lacosmic` multiplies the image by
     the gain; prior to version 2.1 it did so without changing the units of
-    the image.
+    the image which could result in incorrect results.
 
     There are two ways to correctly invoke `~ccdproc.cosmicray_lacosmic`:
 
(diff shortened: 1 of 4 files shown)


mwcraig:
This is the proposal; please let us know what you think @MSeifert04 

There would be two possible sets of inputs:

1. Supply a numpy array as the first argument, then none of gain, etc., should have units, though this won't be enforced. 
1. Supply a `CCDData` object, then consistency of units between `gain`, `ccd`, and `readnoise` will be checked.

Furthermore: 

3. A new boolean parameter, `gain_apply`, is added, with a default value of `True`. With that value the current behavior continues *BUT* the final unit will be correct if a `CCDData` object is provided on input. If `False`, then gain correction is not done, though astroscrappy will will still do the conversion internally.
1. Issue a `Warning` if the `CCDData` has units of `electron` and the gain is not 1.0.
1. Update documentation to make clearer that `lacosmic` does gain correction by default.

mwcraig:
@AnthonyHorton -- can you take a quick look at the outline above? The most salient points are:

+ A new argument, `gain_apply`, that controls whether the data is returned gain corrected. Default is `True` to preserve current behavior, BUT
+ If the input is a `CCDData` and `gain_apply` is `True`, then the returned `CCDData` will have the correct units.
+ There is also a check now for unit consistency.

mwcraig:
The failure on RTD will be fixed once #717 is merged; it contains a new configuration file that installs with `pip` instead of `setup.py`

docs/reduction_toolbox.rst
```diff
@@ -99,6 +99,22 @@ Use this technique with `~ccdproc.cosmicray_lacosmic`:
 
     >>> cr_cleaned = ccdproc.cosmicray_lacosmic(gain_corrected, sigclip=5)
 
+.. note::
+
+    A significant error in `~ccdproc.cosmicray_lacosmic` in version 2.1 of
```
crawfordsm: I think I'd just remove the first line here, but I think the rest gets across the point.

crawfordsm:
I just had one comment but this looks good to me.

mwcraig:
@AnthonyHorton -- thanks again for reporting the problem; we should have a new release out shortly.

PR #723, opened 2020-01-07
Mark tests that require remote data

olebole:
Some tests now use data that are not available locally but need to be retrieved from the data server. This requires remote access, and so the tests should be marked as such.

For documentation changes:

For bugfixes:

For new functionality:

### DIFF
--- a/ccdproc/tests/test_combiner.py
+++ b/ccdproc/tests/test_combiner.py
@@ -321,6 +321,7 @@ def test_combiner_mask_sum():
 
 
 # test combiner convenience function reads fits file and combine as expected
+@pytest.mark.remote_data
 def test_combine_average_fitsimages():
     fitsfile = get_pkg_data_filename('data/a8280271.fits')
     ccd = CCDData.read(fitsfile, unit=u.adu)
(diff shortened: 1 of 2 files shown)


MSeifert04:
Hey @olebole and thanks for submitting this issue.

It's not clear why these use remote data. I thought we ship these files as part of the [package](https://github.com/astropy/ccdproc/tree/master/ccdproc/tests/data).

@mwcraig Any idea why these tests require fetching the data remotely instead of reading them from the package?

olebole:
Ah, I see. I only looked in the base directory, not in `ccdproc/tests` for a `data` subdir.
The reason may be then that we run the tests on the files that were created with `setup.py build`, in the `.pybuild` subdirectory, and there the data files are not copied (and not found). Problem is then similar to scikit-image/scikit-image#4366, and looks like that our local build tool (pybuild) and `setuptools` have a different opinion of what is done at which step.
I think, this one can then just be closed.

MSeifert04:
Ah, that explains it. Thank you for the clarification.

If there's anything we can do from the ccdproc-side to assist you please let us know.

PR #726, opened 2020-01-14
ImageFileCollection fails if the fits headers contain the keyword 'FILE'

jensjessenhansen:
There's an assert statement on line 503 in image_collection.py in method _dict_from_fits_header:
assert 'file' not in h
to catch and raise the problem - but no handling is implemented. 
If not caught the 'file' value list in the summary dictionary would get a filename appended twice for each image.
This will cause the generation of the summary_table to fail.

The suggested way to fix this is to remove the header 'FILE' keyword if present.
This should be okay since it will either be the same as the input filename or likely incorrect - I guess?

For documentation changes:

For bugfixes:

For new functionality:

### DIFF
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -500,8 +500,6 @@ class ImageFileCollection:
 
         h = fits.getheader(file_name, self.ext)
 
-        assert 'file' not in h
-
         if self.location:
             # We have a location and can reconstruct the path using it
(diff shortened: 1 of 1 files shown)


mwcraig:
@jensjessenhansen -- thanks for the PR! I will take a look at it this evening and give you some feedback. This is definitely a case we need to handle better.

PR #727, opened 2020-01-29
Allow specification of FITS suffixes in `ImageFileCollection`

keflavich:
Our observatory writes `.FIT` files, and there appears to be no way at
present to read these with an `ImageFileCollection`.


mwcraig:
@keflavich -- do you have time to add a test for this? If not, I can do that Monday and then merge. I'm not sure why your files are not being picked up since `fit` was already supposed to be one of the recognized extensions but I'm 👍  on this feature in any event.

keflavich:
no time, sorry 🕐

PR #731, opened 2020-05-01
Update block_reduce and block_replicate imports

larrybradley:
In astropy 4.1, these functions moved from the `nddata.utils` to the `nddata.blocks` module (https://github.com/astropy/astropy/pull/10214).  Importing from `nddata.utils` will still work, but will raise a deprecation warning.

### DIFF
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -14,7 +14,7 @@ from astropy.units.quantity import Quantity
 from astropy import units as u
 from astropy.modeling import fitting
 from astropy import stats
-from astropy.nddata import utils as nddata_utils
+from astropy import nddata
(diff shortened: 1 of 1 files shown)


mwcraig:
Thanks @larrybradley -- are the changes to `test_table.txt` needed as part of this?

larrybradley:
@mwcraig No, I just removed the `test_table.txt` file.  It was generated when I ran the `ccdproc` tests locally and I inadvertently included in this PR.  Thanks for catching that.  It's gone now.

mwcraig:
Ah, that is a bug! Convenient timing, I'll open an issue for that for the scipy sprints! Thanks for the fix, merging.

PR #734, opened 2020-05-27
DO NOT MERGE: Test astropy 4.1rc1

pllim:
Test astropy 4.1rc1


pllim:
https://stackoverflow.com/questions/37710848/importerror-no-module-named-resource -- That test should not be running on Windows but fixing it is out of scope. I think given the rest passed, it is good enough a check.

PR #739, opened 2020-07-11
Add memory_profiler as a test requirement [#730]

### LINKED ISSUE #730: memory_profiler is an undeclared dependency
There is an error on test collection if one doesn't relies on the pytest flag of `doctest_ignore_import_error`. It has been done so in the astropy `pytest-doctestplus` plugin, but in the attempt of converging the pluging's behaviour to the default pytest doctest one we consider not ignoring this errors any more.

I can open a workaround PR to the test config, but ideally the maintainers should decide the best way to deal with this dependency.

```
(test_imports) $ ~/munka/devel/ccdproc [hacking|…1⚑ 1] $ pytest ccdproc
================================================= test session starts ==================================================
platform darwin -- Python 3.8.2, pytest-5.4.1, py-1.8.1, pluggy-0.13.1

Running tests with ccdproc version 2.1.1.dev1+gb9ec64d.
Running tests in ccdproc.

Date: 2020-04-30T19:49:39

Platform: macOS-10.13.6-x86_64-i386-64bit

Executable: /Users/bsipocz/miniconda/envs/test_imports/bin/python

Full Python Version: 
3.8.2 (default, Mar 26 2020, 10:43:30) 
[Clang 4.0.1 (tags/RELEASE_401/final)]

encodings: sys: utf-8, locale: UTF-8, filesystem: utf-8
byteorder: little
float info: dig: 15, mant_dig: 15

Package versions: 
Numpy: 1.18.3
Scipy: 1.4.1
Matplotlib: 3.2.1
Pandas: not available
Astropy: 4.0.1.post1
astroscrappy: 1.0.8
reproject: 0.7
astropy-helpers: 4.0.1

Using Astropy options: remote_data: none.

rootdir: /Users/bsipocz/munka/devel/ccdproc, inifile: setup.cfg
plugins: arraydiff-0.3, remotedata-0.3.2, hypothesis-5.10.4, filter-subpackage-0.1.1, openfiles-0.5.0, astropy-header-0.1.2, cov-2.8.1, doctestplus-0.7.0.dev0
collected 305 items / 1 error / 304 selected                                                                           

======================================================== ERRORS ========================================================
_______________________________ ERROR collecting ccdproc/tests/run_for_memory_profile.py _______________________________
ccdproc/tests/run_for_memory_profile.py:8: in <module>
    from memory_profiler import memory_usage
E   ModuleNotFoundError: No module named 'memory_profiler'
=============================================== short test summary info ================================================
ERROR ccdproc/tests/run_for_memory_profile.py - ModuleNotFoundError: No module named 'memory_profiler'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=================================================== 1 error in 0.67s ===================================================

```

jryon:
Fixes #730 by adding `memory_profiler` as a test requirement to `setup.cfg`. Also no longer ignores import errors in testing.

For bugfixes:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,10 @@ env:
         - CONDA_CHANNELS='astropy'
         - ASTROPY_USE_SYSTEM_PYTEST=1
 
+        # List runtime dependencies for the package that are available as conda
+        # packages here.
+        - CONDA_DEPENDENCIES='memory-profiler'
+
     matrix:
         - SETUP_CMD='egg_info'
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ env:
 
         # List runtime dependencies for the package that are available as conda
         # packages here.
-        - CONDA_DEPENDENCIES='memory-profiler'
+        - CONDA_DEPENDENCIES='memory_profiler'
 
     matrix:
         - SETUP_CMD='egg_info'
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ New Features
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
-- Add memory_profiler as a test requirement [#730]
+- Add memory_profiler as a test requirement [#739]
 
 Bug Fixes
 ^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ New Features
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+- Add memory_profiler as a test requirement [#730]
 
 Bug Fixes
 ^^^^^^^^^
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -46,6 +46,7 @@ Alphabetical list of contributors
 * Connor Stotts (@stottsco)
 * Ole Streicher (@olebole)
 * JVSN Reddy (@janga1997)
+* Jenna Ryon (@jryon)
 * Brigitta Sipocz (@bsipocz)
 * Erik Tollerud (@eteq)
 * Simon Torres (@simontorres)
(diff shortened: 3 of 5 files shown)


CHANGES.rst
```diff
@@ -6,6 +6,7 @@ New Features
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+- Add memory_profiler as a test requirement [#730]
```
crawfordsm, line 9: You can add the PR number here to have the change bot pass

mwcraig:
@crawfordsm -- the failures here were expected. Right now we still use `pytest` instead of using `tox`. THere are two possible fixes:

+ Use `tox` instead of `pytest`.
+ Add `memory_profiler` as a dependency in travis

Any preference?

crawfordsm:
As there might be some additional tox issues, it might be easiest to add it to jenkins otherwise there might be more than one issue to sort out.

mwcraig:
@jryon -- can you please make one more change? We need to add a line like this https://github.com/astropy/specutils/blob/master/.travis.yml#L49 to our `.travis.yml` file, but with `memory_profiler` as the dependency.

mwcraig:
Opened #740 to keep track of switching to tox

.travis.yml
```diff
@@ -24,6 +24,10 @@ env:
         - CONDA_CHANNELS='astropy'
         - ASTROPY_USE_SYSTEM_PYTEST=1
 
+        # List runtime dependencies for the package that are available as conda
+        # packages here.
+        - CONDA_DEPENDENCIES='memory-profiler'
```
mwcraig, line 29: Maybe it should be an underscore instead of a hyphen in `memory-profiler`?

jryon: whoops, thanks!

mwcraig:
Looks like this is one of those cases where for every two bugs squashed, one more appears. The issue here is that the `resource` module is not available on Windows. We are already skipping the tests that actually use that but I guess pytest still tries to import everything in the tests directory and fails on trying to import it in `run_with_file_number_limit.py` (see https://travis-ci.org/github/astropy/ccdproc/jobs/708038685#L544).

not sure what the elegant solution would be here -- what do you think about putting the import in a `try`/`except` block, @jryon?

mwcraig:
Thanks, @jryon, merging!

PR #741, opened 2020-07-24
Improve image combination performance

### LINKED ISSUE #719: combine uncertainty calculated on unscaled data
I was combining hundreds of sky flat images. I wanted each image to be scaled to median (divide by median of each frame) and combine, and see the uncertainty level.

The output combined image looked good (pixel values roughly 1) but the uncertainty was too large (mostly 2-5). 

I found the ccdproc does combining using scaled data (e.g., line 352 of [combiner.py](https://github.com/astropy/ccdproc/blob/070360ae22b788042ccd8adae81e9c36ab894578/ccdproc/combiner.py))
```python
data = median_func(self._get_scaled_data(scale_to), axis=0)
```
while the uncertainty is calculated using the original data (line 358-361 of the same file)
```python
# set the uncertainty
uncertainty = uncertainty_func(self.data_arr, axis=0)
# Divide uncertainty by the number of pixel (#309)
uncertainty /= np.sqrt(len(self.data_arr) - masked_values)
```

**Is this an expected behavior?**
If so, how can I get the uncertainty of the median-normalized pixel values?
If not, will the change ``uncertainty = uncertainty_func(self._get_scaled_data(scale_to), axis=0)`` correct all the things?

```
import astropy
print(astropy.__version__)
import ccdproc
print(ccdproc.__version__)
import numpy
print(numpy.__version__)
# 3.2.3
# 2.0.1
# 1.17.4
```

mwcraig:
This pull request attempts to improve the performance of ccdproc by using using numpy's `nan*` functions instead of numpy `MaskedArray`, and using by using `bottleneck`. It is not intended (yet) to change the API for `Combiner`. 

So far I have only switched `average_combine` to do this. If I get a couple :+1: on this I'll do the same for `median_combine` and the clipping routines, and also combine the implementations of `sum_combine` and `average_combine` to the extent possible.

To do:

- [x] Document weighting during image combination
- [ ] Add link to preliminary benchmarking
- [x] Unify sum and average combination
- [ ] Use `nan*` or `bottleneck` for clipping
- [x] Changelog entry
- [x] make bottleneck an optional dependency
- [x] Add prominent suggestion to use bottleneck
- [ ] ~~Consider making dtypes settable (might belong in `CCDData`).~~ -- this is a separate issue, really, and CCDData lives in astropy, core, not here.

Edit: Fixes #719


mwcraig:
# Performance improvements (no clipping done)

Some benchmark plots are below. left-most commit is master, right-most uses bottleneck. Second-to-last uses `nan*` instead of masked array.

## `average_combine` with weighting: Speedup is ~ x2

average-combine-weighted

## Unweighted `average_combine`, some entries masked. Speedup is ~ x3

average-no-weights-some-masked

## Unweigthed `average_combine`, no entries masked. Speedup is ~ x2.8

average-no-weights-no-mask

mwcraig:
Ping @crawfordsm @MSeifert04 @saimn @cmccully @ysBach -- you are either a ccdproc maintainer or someone who is working hard separate from this to improve image combination speed. Keep an eye out for an email invite later today to talk about whether we can pool efforts on this or not.

I don't need a detailed review of this -- but if you have objects to this stop-gap approach for improving performance please speak up!

saimn:
Great to see progress on this. I started to do some comparison with what we have in DRAGONS, but it is still a wip.

About bottleneck, since I had the opportunity to look at it more closely recently (https://github.com/astropy/astropy/issues/10553#issuecomment-658282236):
- It does have optimized algorithm only for little-endian data. For big-endian (FITS...) it has a fallback to `np.nan*` functions.
- And currently the fallback to numpy comes with a pretty bad memory leak.

Other than that, :+1: to use `nan*` functions instead of masked arrays, it is much faster.

saimn:
Looking more in detail at the code it seems that ccdproc converts the data to np.float64 by default, unless a dtype is specified. So the good news is that it will not be affected by the bottleneck issue. But in terms of performance it would be faster to use the original data's dtype.

mwcraig:
One other bottleneck note: there are some precision issues with sums of float32: https://github.com/pydata/bottleneck/issues/193

mwcraig:
@saimn @ysBach -- any chance either of you can take a look at this? It has become a little convoluted because I'm trying to make sure I don't change the API. THis si only a small step towards improving performance, but it is a step...

For median the only improve is if bottleneck is installed because `np.nanmedian` just calls `np.ma.median` after setting up a mask internally.

Once this is wrapped up (hopefully this week) I'd like to come back to further improvements, likely the second week of June.

PR #743, opened 2020-08-11
A workaround for unfixable header keys in ImageFileCollection

gully:
Unfixable header keys are currently populating the summary_dict with their genuine values and also missing value place holders, resulting in lists of values that are too long, breaking the ImageFileCollection for data with these malformed headers.  

I tried to track down the root cause, and could pinpoint it to the `_dict_from_fits_header()` method called within `_fits_summary`.  My hunch is there's something special about these unfixable keys that gets them double counted as `missing_in_this_file`.  A key clue is that the resulting list for offending keys is always 2*N-1 long, where N is the number of files.

This workaround is really just that-- a stopgap measure to get this to work on the types of files I have.  I suspect the right solution is to dig a bit deeper on `_dict_from_fits_header`, so I didn't go through with a full check of tests, assuming this workaround won't actually be in the final product.

Here are urls to two example files that reproduce the problem:
```
https://koa.ipac.caltech.edu/cgi-bin/getKOA/nph-getKOA?instrument=NS&filehand=/koadata11/NIRSPEC/20010117/lev0/spec/NS.20010117.25279.fits

https://koa.ipac.caltech.edu/cgi-bin/getKOA/nph-getKOA?instrument=NS&filehand=/koadata11/NIRSPEC/20010117/lev0/spec/NS.20010117.25430.fits
```
The malformed keys are `GAIN.SPE` and `FREQ.SPE`, though they do not appear in the error message, truncated below:

```python

    627                 pass#summary_dict[key]= [None]*len(summary_dict['file'])
    628 
--> 629         summary_table = Table(summary_dict, masked=True)
    630 
    631         for column in summary_table.colnames:


ValueError: Inconsistent data column lengths: {4, 7}
```

For documentation changes:

For bugfixes:

For new functionality:

### DIFF
--- a/ccdproc/log_meta.py
+++ b/ccdproc/log_meta.py
@@ -10,6 +10,9 @@ from astropy.nddata import NDData
 from astropy import units as u
 from astropy.io import fits
 
+import warnings
+warnings.filterwarnings("ignore", category=DeprecationWarning) 
+
 import ccdproc  # really only need Keyword from ccdproc
 
 __all__ = []
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -56,6 +56,7 @@ Alphabetical list of contributors
 * Nathan Walker (@walkerna22)
 * Benjamin Weiner (@bjweiner)
 * Jiyong Youn (@hletrd)
+* Michael Gully-Santiago (@gully)
 
 (If you have contributed to the ccdproc project and your name is missing,
 please send an email to the coordinators, or
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,8 @@ Bug Fixes
 - ``test_image_collection.py`` in the test suite no longer produces
  permanent files on disk and cleans up after itself. [#738]
 
+- Workaround malformed header keyword names  [#743]
+
 2.1.0 (2019-12-24)
 ------------------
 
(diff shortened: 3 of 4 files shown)


gully:
I am seeing that this PR breaks 1 test:  
`py.test -vsk test_filenames_are_set_properly` 
Currently investigating...

mwcraig:
Thanks -- can you please ping when you have fixed that? Tests are currently not running on travis because they've dropped support for open source projects. 

Also, don't be shy about pinging me multiple times 😬

gully:
I fixed the failed test which in hindsight was an edge case of a None header.  Also added a warning logger.
I'm a bit muddled on this PR.  On one hand it's a quick fix that makes it possible to move forward on any data that has malformed FITS keywords, which may be relatively common and hindering other folks from getting started.  On the other hand it just treats the symptom not the sickness, and it loses available information (currently just fills the keyword with `None`s rather than attempt to sort out the genuine values).  In other words the image collection class would come with the proviso: "Currently does not support malformed header keywords".

Probably best practices would be to add a test that has such a malformed keyword and then make sure the correct `key:value` pairs are returned.  Or if remote data can be tolerated in the test suite, then the two files I linked above would reveal the problem.

mwcraig:
I'm going to close/reopen to see if that picks up the new GitHub Actions CI. More substantive review tomorrow, I hope...

ccdproc/log_meta.py
```diff
@@ -10,6 +10,9 @@
 from astropy import units as u
 from astropy.io import fits
 
+import warnings
+warnings.filterwarnings("ignore", category=DeprecationWarning)
```
mwcraig, line 14:
```suggestion
warnings.filterwarnings("ignore", category=DeprecationWarning)
```

PR #750, opened 2020-12-08
Fix bug when ImageFileCollection.filter() would return an empty collection

tbowers7:
If the filtering criteria passed to ImageFileCollection.filter() would return
an empty collection, a crash occurs because the method instantiates a new
ImageFileCollection() with a list of filenames returned from .files_filtered().
When a new instance of ImageFileCollection() is called without a filenames
list, _get_files() -> _fits_files_in_directory() attempts to create a list
of all FITS files within the directory contained in self.location.  If this
new instance of ImageFileCollection() is called without setting location, then
self.location is set to "".  When _fits_files_in_directory() attempts to do a
directory listing via listdir(), an error is raised:
    FileNotFoundError: [Errno 2] No such file or directory: ''

This commit adds a check for empty self.location before any attempted call to
listdir().  The function simply returns an empty string, leading to an empty
ImageFileCollection instance.  This is the expected behavior for filtering a
collection such that no files meet the criteria.

- [x] For new contributors: Did you add yourself to the "Authors.rst" file?  Yes.

For documentation changes:

- [x] For documentation changes: Does your commit message include a "[skip ci]"?  N/A

For bugfixes:

- [x] Did you add an entry to the "Changes.rst" file? Yes.
- [x] Does the commit message include a "Fixes #issue_number" (replace "issue_number").  N/A
- [x] Does this PR add, rename, move or remove any existing functions or parameters?  No.

For new functionality:

- [x] Did you add an entry to the "Changes.rst" file?  N/A
- [x] Did you include a meaningful docstring with Parameters, Returns and Examples?  N/A
- [x] Does the commit message include a "Fixes #issue_number" (replace "issue_number").  N/A
- [x] Did you include tests for the new functionality?  N/A
- [x] Does this PR add, rename, move or remove any existing functions or parameters?  N/A

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -24,6 +24,7 @@ Alphabetical list of contributors
 * Mihai Cara (@mcara)
 * James Davenport (@jradavenport)
 * Christoph Deil (@cdeil)
+* Timothy P. Ellsworth-Bowers (@tbowers7)
 * Carlos Gomez (@carlgogo)
 * Hans Moritz Günther (@hamogu)
 * Forrest Gasdia (@EP-Guy)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,10 @@ Bug Fixes
 - ``test_image_collection.py`` in the test suite no longer produces
  permanent files on disk and cleans up after itself. [#738]
 
+- ``ImageFileCollection`` now correctly returns an empty collection when
+  an existing collection is filtered restrictively enough to remove all
+  files.  [#750]
+
 2.1.0 (2019-12-24)
 ------------------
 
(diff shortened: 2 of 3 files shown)


mwcraig:
@tbowers7 -- thanks for this!

Can you please add a changelog entry to https://github.com/astropy/ccdproc/blob/master/CHANGES.rst in the section for 2.1.1, and add yourself to the list of contributors here: https://github.com/astropy/ccdproc/blob/master/AUTHORS.rst

If you don't have time, let me know and I'll get them added.

tbowers7:
Adjusted the way I am fixing this bug, moving the check for self.location back a step into the `_get_files()` method instead of having it in the `_fits_files_in_directory()` method.

This adjustment makes more sense because the check is now in sequence with checking for a file list passed to ImageFileCollection.

mwcraig:
Thanks for the fix, @tbowers7! In the feature please try to make a separate branch for your fixes --- not a big deal, but makes it a little easier if fixes to a PR are needed.

Issue #725, opened 2020-01-12
Problem with Read-the-docs (RTD) on Google


MSeifert04:
When searching for ccdproc on google I get this search result:

![Unbenannt](https://user-images.githubusercontent.com/14200878/72218156-22697a00-3538-11ea-9b68-61a65501cc83.png)

I think something went wrong with the version because it reports "vunknown". Does someone know what could cause this?

mwcraig:
Closed by #755

PR #755, opened 2021-01-12
Fix Read-the-docs ccdproc version

Yash-10:
This PR addresses #725 with a minor fix by providing the relevant `release` of ccdproc in the `html_title` which currently shows up as `vunknown` in the docs. This was done by importing `project` and accessing the `version` from it, which can also be seen done in other astropy affiliated packages like `photutils`, reference: [here](https://github.com/astropy/photutils/blob/master/docs/conf.py#L70-L76)


mwcraig:
Thanks, merging!

PR #759, opened 2021-02-01
Reduce test warnings

Yash-10:
Addressed some test warnings that come up while testing. Also made comments consistent along the way.

Thanks!

### DIFF
--- a/ccdproc/extern/bitfield.py
+++ b/ccdproc/extern/bitfield.py
@@ -248,7 +248,7 @@ def interpret_bit_flags(bit_flags, flip_bits=None):
 
 def bitfield_to_boolean_mask(bitfield, ignore_flags=0, flip_bits=None,
                              good_mask_value=True, dtype=np.bool_):
-    """
+    r"""
     bitfield_to_boolean_mask(bitfield, ignore_flags=None, flip_bits=None, \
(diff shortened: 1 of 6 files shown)


Yash-10:
@mwcraig, could you please review this PR?

Thanks!

Yash-10:
For the windows CI test, it seems that `len(recwarn)` is 2 and hence the assertion fails. Although these tests on macos and ubuntu are cancelled. [This](https://github.com/astropy/ccdproc/blob/master/ccdproc/tests/test_cosmicray.py#L128) line renders a log message
```
array provided for uncertainty; assuming it is a StdDevUncertainty. [astropy.nddata.ccddata]
```
so probably this is been captured by `recwarn` in some way?

mwcraig:
It looks like that call is now generating two warnings because of a new deprecation warning from numpy. I think dropping the check for the number of warnings would work for now.

Yash-10:
Hello @mwcraig, thank you for your reply. One of the deprecation warnings has been fixed here and the other seems to be coming from `astroscrappy`, the equivalent fix of which was already done and is close to be merged. Also, as you had suggested, the check for the number of warnings is dropped now.

Thank you.

mwcraig:
Excellent, thanks! I'll try to get 2.1.1 released today.

PR #761, opened 2021-03-08
Update Zenodo badge to latest release

mwcraig:


mwcraig:
Merging even if this fails, because #759 will fix the failure, I think

PR #768, opened 2021-04-09
MNT: Do not write config on import

pllim:
This functionality will be deprecated in `astropy` in a future release. See astropy/astropy#11497 and https://groups.google.com/g/astropy-dev/c/KVzPgeihxO8

Bonus: Clean up test runner code.

### DIFF
--- a/ccdproc.cfg
+++ b/ccdproc.cfg
@@ -1,5 +1,0 @@
-[ccdproc]
-# auto_logging = True
-# Whether to automatically log operations to metadata
-# If set to False, there is no need to specify add_keyword=False
-# when calling processing operations
--- a/ccdproc/_astropy_init.py
+++ b/ccdproc/_astropy_init.py
@@ -1,6 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-__all__ = ['__version__', 'test']
+__all__ = ['__version__']
 
 # this indicates whether or not we are in the package's setup.py
 try:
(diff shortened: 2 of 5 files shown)


ccdproc.cfg
```diff
@@ -1,5 +0,0 @@
-[ccdproc]
-# auto_logging = True
-# Whether to automatically log operations to metadata
-# If set to False, there is no need to specify add_keyword=False
-# when calling processing operations
```
saimn, line 5: As ccdproc has some custom config items, this should be documented using the sphinx extension: https://github.com/astropy/sphinx-astropy/blob/main/sphinx_astropy/ext/generate_config.py

pllim: Good call, @saimn . Added a commit. How does it look now?

ccdproc/_astropy_init.py
```diff
@@ -14,120 +14,11 @@
...
-    test_runner = _get_test_runner()
-    return test_runner.run_tests(
-        package=package, test_path=test_path, args=args,
-        plugins=plugins, verbose=verbose, pastebin=pastebin,
-        remote_data=remote_data, pep8=pep8, pdb=pdb,
-        coverage=coverage, open_files=open_files, **kwargs)
-
-
-if not _ASTROPY_SETUP_:
+if not _ASTROPY_SETUP_:  # noqa
```
saimn, line 17: Can be removed too.

ccdproc.cfg
saimn: Looks good :+1:

ccdproc/_astropy_init.py
pllim: Removed. What about now?

mwcraig (approved):
Thanks, merging!

Issue #769, opened 2021-05-01
ImageFileCollection filtering and regex_match


mkelley:
I'm filtering an ImageFileCollection, trying to find files by filter name.  I've run into a bug with a filter named `'CO+'`.  I use `regex_match=False` to search for the literal filter name, but the code in `ImageFileCollection._find_keywords_by_values` still parses this as a regular expression.  According to the documentation, `regex_match=False` should search by equality.

```python

from tempfile import TemporaryDirectory
import numpy as np
from astropy.io import fits
from ccdproc import ImageFileCollection

h = fits.Header()
h['FILTER'] = 'CO+'

with TemporaryDirectory() as tmpdir:
    fits.writeto(tmpdir + '/test.fits', np.zeros(1), header=h)
    ic = ImageFileCollection(location=tmpdir)

print("filter='CO+'")
print("  regex_match=False", ic.files_filtered(filter='CO+', regex_match=False))
print("  regex_match=True", ic.files_filtered(filter='CO+', regex_match=True))
print()
print(r"filter='CO\+'")
print("  regex_match=False", ic.files_filtered(
    filter=r'CO\+', regex_match=False))
print("  regex_match=True", ic.files_filtered(
    filter=r'CO\+', regex_match=True))
```

The results:

```
filter='CO+'
  regex_match=False []
  regex_match=True ['test.fits']

filter='CO\+'
  regex_match=False ['test.fits']
  regex_match=True ['test.fits']
```

* Filtering with 'CO+' and regex=False should have returned the file, but the result is empty.
* Filtering with 'CO+' and regex=True returns the file, because the regex matches 'CO', as expected.
* Filtering with 'CO\\+' and regex=False should not have returned the file, because the strings are not equal: `'CO+' != r'CO\+'`.
* Filtering with 'CO\\+' and regex=True returns the file, because this regex matches the full string 'CO+'

```
import astropy
print(astropy.__version__)
import ccdproc
print(ccdproc.__version__)
import numpy
print(numpy.__version__)
```
4.0.1.post1
2.1.1
1.18.5

mwcraig:
Closed by #770 -- thanks for reporting this!

PR #770, opened 2021-05-10
Escape regex special characters in FITS keyword values used for IFC filtering

tbowers7:
In `ImageFileCollection()`, the method `_find_keywords_by_values()` uses regular
expressions to find matches for specified FITS keywords by value.  In the case
where a FITS keyword value has parentheses (whether for a mathematical or
facility reason), the regex search interprets the parenthesis as a control
character and does not correctly match the string as specified, returning none
of the desired frames.

As an example from the Lowell Observatory Discovery Telescope (LDT) DeVeny
Optical Spectrograph, we specify order-blocking filters both by name and filter
wheel position, e.g.:
`FILTREAR= 'Clear (C)'          / Rear Filter`

There is logic in the `_find_keywords_by_values()` method to specify that one is
supplying a regular expression for search, but the method also constructs a
simple regex search based on the specified keyword value if `regex_match=False`.

This commit adds escape characters to parentheses in the case where the user
has NOT specified an input regular expression.  The resulting pattern search is
successful.

Also, add the shell of a test for parenthetical FITS keyword values.  The actual 
test would require test data that contains a FITS keyword whose value includes
a parenthesis.  An example is given in the comment.

	modified:   ccdproc/image_collection.py
	modified:   CHANGES.rst
	modified:   ccdproc/tests/test_image_collection.py

- [x] For new contributors: Did you add yourself to the "Authors.rst" file?  N/A

For documentation changes:

- [x] For documentation changes: Does your commit message include a "[skip ci]"?  N/A

For bugfixes:

- [x] Did you add an entry to the "Changes.rst" file?  Yes.
- [x] Did you add a regression test?  Yes-ish.  Added a shell, but requires different test data.
- [x] Does the commit message include a "Fixes #issue_number" (replace "issue_number").  N/A
- [x] Does this PR add, rename, move or remove any existing functions or parameters?  No.

For new functionality:

- [x] Did you add an entry to the "Changes.rst" file?  N/A
- [x] Did you include a meaningful docstring with Parameters, Returns and Examples?  N/A
- [x] Does the commit message include a "Fixes #issue_number" (replace "issue_number").  N/A
- [x] Did you include tests for the new functionality?  N/A
- [x] Does this PR add, rename, move or remove any existing functions or parameters?  N/A

### DIFF
--- a/ccdproc/tests/test_image_collection.py
+++ b/ccdproc/tests/test_image_collection.py
@@ -1121,7 +1121,6 @@ class TestImageFileCollection:
                         'regex_special_{:d}.fits'.format(i)))
 
         ic = ImageFileCollection(triage_setup.test_dir)
-        #ic.summary.pprint()
         for kw in special_kwds:
             new_ic = ic.filter(regex_fl=kw)
             assert len(new_ic.files) == 1


ccdproc/image_collection.py
```diff
@@ -710,6 +710,8 @@ def _find_keywords_by_values(self, **kwd):
                         pattern = re.compile(value,
                                              flags=re.IGNORECASE)
                     else:
+                        # Add escape characters to parentheses:
+                        value = value.replace('(',r'\(').replace(')',r'\)')
```
mwcraig, line 714: I think this would fix #769 it were this instead, so that all special characters were escaped:

```suggestion
                        value = re.escape(value)
```

ccdproc/tests/test_image_collection.py
```diff
@@ -1100,3 +1100,8 @@ def test_filtered_collection_with_no_files(self, triage_setup):
         ifc = ImageFileCollection(triage_setup.test_dir)
 
         ifc_no_files = ifc.filter(object='really fake object')
+
+    def test_filter_with_parenthetical_keyword_values(self, triage_setup):
+        # Would need test data containing FITS keywords whose values have ()
+        # e.g.:  FILTREAR= 'Clear (C)'          / Rear Filter
+        pass
```
mwcraig, line 1107: I think the short example that @mkelley provided in #769 could be used as the body of the test here, or at least as the start of a test. It generates a FITS file with a special character in it then tries to filter by a value with the special character. 

I can add a credit to @mkelley for the commit with the test body if you end up using a lot of his sample code.

mwcraig (changes requested):
Thanks for opening this @tbowers7. I think with one small change it would fix #769, which reports a similar issue. When the regex search was added ~2 years ago I didn't even think about the fact that the `value` might have regex characters in it. 

Do you have time to revise this week? I'm hoping to release v2.2 by Monday and would like to include this.

If you don't have time, let me know -- I can either delay the release or do the revision.

Thanks!

ccdproc/tests/test_image_collection.py
tbowers7: Yes, I will add this test this week.  Change to use re.escape() already committed.

mwcraig: Excellent, thanks!

tbowers7:
Test written and committed.  The test was tested separately, passing when the `value = re.escape(value)` line is included in `ccdproc/tests/test_image_collection.py`, and failing when that line is commented out (_i.e._ pre-PR#770).

ccdproc/tests/test_image_collection.py
```diff
@@ -1100,3 +1100,29 @@ def test_filtered_collection_with_no_files(self, triage_setup):
...
+        special_kwds = ['CO+', 'GG420 (1)', 'V|R|I', 'O[III]', 'NaD^2']
+        for i,kw in enumerate(special_kwds,1):
+            hdu = fits.PrimaryHDU()
+            hdu.data = np.zeros((5, 5))
+            hdu.header['REGEX_FL'] = kw
+            hdu.writeto(os.path.join(triage_setup.test_dir,
+                        'regex_special_{:d}.fits'.format(i)))
+
+        ic = ImageFileCollection(triage_setup.test_dir)
+        #ic.summary.pprint()
```
mwcraig, line 1124: Can you please take this line out?
```suggestion

```

mwcraig (approved):
Excellent, thanks for  the quick turnaround -- just one line to delete in the test and this is good to go!

ccdproc/tests/test_image_collection.py
tbowers7: All done.

mkelley:
Thanks for the fix!

PR #778, opened 2021-11-20
Update for astroscrappy 1.1.0

### LINKED ISSUE #777: Many tests fail with the latest astroscrappy release
For example, `test_cosmicray_gain_correct` fails with:

```
E   TypeError: detect_cosmics() got an unexpected keyword argument 'pssl'
```

mwcraig:
To do:

- [x] Add docstrings for `inbkg`, `invar`.
- [x] Update documentation in `reduction_toolbox`
- [x] Deprecate `pssl`
- [ ] *Maybe* deprecate `gain_apply` -- that was introduced because `astroscrappy` used to always apply gain. Now it does not.
- [x] Commit the changelog entry.
- [x] Does the commit message include a "Fixes #issue_number" (replace "issue_number"). Fixes #777


mwcraig:
@saimn @cmccully -- would you mind taking a look at this when you get a chance? The last big question I have is this: 
How much work is it to make sure the change in astroscrappy version is transparent to users? 

I think I've handled the mechanical bits (adding new keyword args, removing the old ones). What I *haven't* done is address the fact that background should not be subtracted now, and the image should be in counts, not electrons. 

That does need to be addressed, right?

saimn:
I guess handling the compatibility with the old and new version could be a bit painful. Given that the old version did not install on Python 3.9 & 3.10 (with pip at least), maybe having a requirement on astroscrappy >= 1.1.0 could be acceptable ? This would make it much easier for you.

mwcraig:
After spending some time away from the keyboard, supporting both seems easier than I thought. I need to add a changelog and review the documentation, but then will merge. My plan is to have this be the last ccdproc release that supports 1.0.8.

PR #781, opened 2021-12-03
Fix tox astropy dependency issue

mwcraig:
Turns out the `oldestdeps` tests were running with astropy `5` not the `4.0.6` requested. Also turns out that the combination of astroscrappy `1.0.8` and numpy `<1.20` is difficult so the `oldestdeps` test now uses numpy `1.20` which is not really the oldest supported one.


mwcraig:
Ignoring codecov since the change is trivial

PR #784, opened 2021-12-31
MNT: Stop using deprecated astropy.tests stuff

pllim:
Please have a look at the following list and replace the "[ ]" with a "[x]" if
the answer to this question is yes.

- [x] For new contributors: Did you add yourself to the "Authors.rst" file?

For documentation changes:

- [x] For documentation changes: Does your commit message include a "[skip ci]"?
      Note that it should not if you changed any examples!

For bugfixes:

- [x] Did you add an entry to the "Changes.rst" file? Not needed
- [x] Did you add a regression test? Not needed
- [x] Does the commit message include a "Fixes #issue_number" (replace "issue_number"). N/A
- [x] Does this PR add, rename, move or remove any existing functions or parameters? No

For new functionality:

- [x] Did you add an entry to the "Changes.rst" file? Not needed
- [x] Did you include a meaningful docstring with Parameters, Returns and Examples? N/A
- [x] Does the commit message include a "Fixes #issue_number" (replace "issue_number"). N/A
- [x] Did you include tests for the new functionality? N/A
- [x] Does this PR add, rename, move or remove any existing functions or parameters? No

Please note that the last point is not a requirement. It is meant as a check if
the pull request potentially breaks backwards-compatibility.

-----------------------------------------

This pull request is to remove soon to be deprecated code; also see astropy/astropy#12633 .

PR #785, opened 2022-01-12
Fix doc build warning

mwcraig:
A warnign was happening during the documentation build because of   https://github.com/astropy/sphinx-automodapi/issues/141  (see [this failing run](https://github.com/astropy/ccdproc/runs/4757634889?check_suite_focus=true#step:7:79))

Once a release of `sphinx-automodapi` has been done the pin I added here for `sphinx-automodapi` version should be removed.


mwcraig:
Merging despite the failing test because it is fixed in #784...but this needs to get merged so the documentation build works there.

PR #820, opened 2024-04-08
TST: fix compatibility with pytest 8

neutrinoceros:
This should resolve the current CI instability. ([example logs](https://github.com/astropy/ccdproc/actions/runs/8575940098/job/23505969612?pr=819))

For context, what I think is happening is that the `pytest.warns` context manager's behavior was changed in pytest 8.0. In older versions, all warnings *other* than the one we were checking for were silently swallowed, and now they're not. In this specific test, `astropy` would also emit some `AstropyUserWarning`(s), which we don't control but are pointing to the very same problem we're testing for, so we should just filter them explicitly.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -76,7 +76,7 @@ jobs:
     - name: Check out repository
       uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python }}
     - name: Install base dependencies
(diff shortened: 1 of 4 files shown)


neutrinoceros:
there was another warning specificly in oldest deps, which took a bit longer to inspect. I think I fixed it but I'll undraft only when CI is actually green everywhere

neutrinoceros:
For context, I tracked the numpy deprecation warning down to astroscrappy. This was fixed upstream in https://github.com/astropy/astroscrappy/commit/e5c12fac46153336f12d7f03aa726d8136a777d7 (first published in v1.1.0)

I also threw in the `--color=yes` pytest flag while I was at it, as done in astropy, which helps reading logs in CI

neutrinoceros:
btw Python 3.11 and 3.12 are not tested at all. I'll add them to the test matrix in a follow up PR just in case it uncovers incompatibilities and I need to fix those too.

mwcraig:
Thanks, @neutrinoceros!

PR #822, opened 2024-04-09
TST: add CI coverage for Python 3.11 and 3.12

neutrinoceros:
Based off #820

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -96,12 +96,10 @@ jobs:
     - name: Install graphviz dependency
       if: "endsWith(matrix.tox_env, 'build_docs')"
       run: sudo apt-get -y install graphviz
-    - name: Print Python, pip, setuptools, and tox versions
+    - name: Print Python env
       run: |
-        python -c "import sys; print(f'Python {sys.version}')"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -86,12 +96,10 @@ jobs:
     - name: Install graphviz dependency
       if: "endsWith(matrix.tox_env, 'build_docs')"
       run: sudo apt-get -y install graphviz
-    - name: Print Python, pip, setuptools, and tox versions
+    - name: Print Python env
       run: |
-        python -c "import sys; print(f'Python {sys.version}')"
-        python -c "import pip; print(f'pip {pip.__version__}')"
-        python -c "import setuptools; print(f'setuptools {setuptools.__version__}')"
-        python -c "import tox; print(f'tox {tox.__version__}')"
```
neutrinoceros, line 94: the reason why I changed this is that, since Python 3.12.0, `setuptools` is *not* bundled with the standard lib, so this would crash. `pip list` is safer, less verbose, and gives even more information.

neutrinoceros:
rebased on main

mwcraig (approved):

PR #823, opened 2024-04-10
DEPR: deprecate 'clobber' alias for overwrite keyword argument

neutrinoceros:
While testing the package, I stumbled upon a `_ASTROPY_LT_1_3` constant so I pulled the thread.


mwcraig:
> While testing the package, I stumbled upon a _ASTROPY_LT_1_3 constant so I pulled the thread.

😬 there is some "vintage" code in there....thanks for catching this!

mwcraig (approved):

PR #824, opened 2024-04-10
BUG: fix compatibility with numpy 2.0.0 (copy semantics)

neutrinoceros:
There are a couple other failing tests with numpy 2 but this one is self contained.
I'll open a different PR for the rest.


neutrinoceros:
@mwcraig there's another patch for numpy 2 in https://github.com/astropy/ccdproc/pull/826
And I'm still seeing 3 failing tests locally, but they're about testing scripts and the only info we get from the tests is that the return value is non-zero. Debugging will probably take a bit longer but I'm getting to it !

mwcraig (approved):

PR #830, opened 2024-06-19
TST: pin numpy<2 for tests with old astroscrappy

neutrinoceros:
I noticed CI was broken since numpy 2.0 was released: [example logs](https://github.com/astropy/ccdproc/actions/runs/9542750335/job/26298102524)
This should address it.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -38,7 +38,7 @@ deps =
     numpy126: numpy==1.26.*
 
     astroscrappy11: astroscrappy==1.1.*
-    astroscrappy10: astroscrappy==1.0.*
+    astroscrappy11: numpy<2.0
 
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -38,7 +38,7 @@ deps =
     numpy126: numpy==1.26.*
 
     astroscrappy11: astroscrappy==1.1.*
-    astroscrappy10: astroscrappy==1.0.*
```
neutrinoceros, line 41: I'm also removing the astropyscrappy10 env as it doesn't seem to be used in CI, but obviously I'm happy to revert this if undesired.

mwcraig:
Thanks, @neutrinoceros!

PR #831, opened 2024-07-26
Fixes #796 - cosmicray_lacosmic data type return

### LINKED ISSUE #796: cosmicray_lacosmic does not return numpy.ndarray
I have found that **cosmicray_lacosmic** does not return numpy.ndarray, but a astropy.units.quantity.Quantity.
The problem comes from _astroscrappy_gain_apply_helper(), where the comparison with gain != 1 is always true, and I think it should be as 'gain.value != 1'.

So, when we run:
```
newdata, mask = cosmicray_lacosmic(data, sigclip=5)
```
newdata is a astropy.units.quantity.Quantity instead of numpy.ndarray.

The versions I am using are:
```
import astropy
print(astropy.__version__)
import ccdproc
print(ccdproc.__version__)
import numpy
print(numpy.__version__)
```
4.3.1
2.3.1
1.21.5

altair-above:

For documentation changes:

For bugfixes:

For new functionality:

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -31,6 +31,7 @@ Alphabetical list of code contributors
 * Christoph Deil (@cdeil)
 * Clément M.T. Robert (@neutrinoceros)
 * Timothy P. Ellsworth-Bowers (@tbowers7)
+* Allison Eto (@altair-above)
 * Forrest Gasdia (@fgasdia)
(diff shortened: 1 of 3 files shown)


mwcraig (approved):
Thanks for the PR! I'll try to get a new version out Monday.

PR #871, opened 2025-02-12
Update Numpy RNG calls.  Fixes #839

### LINKED ISSUE #839: Update numpy random number generation calls
There are a bunch of places that ccdproc uses old-style numpy random number generation that should be updated.

AlexN1234:

For documentation changes:

For bugfixes:

For new functionality:

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -55,7 +55,7 @@ jobs:
           - name: 'ubuntu-py312'
             os: ubuntu-latest
             python: '3.12'
-            tox_env: 'py310-test-alldeps-numpy126'
+            tox_env: 'py312-test-alldeps-numpy126'
 
           - name: 'macos-py312'
             os: macos-latest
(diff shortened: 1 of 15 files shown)


mwcraig (changes requested):
Thanks, @AlexN1234! I've marked this as "Request changes" for now so I don't accidentally merge it.

A couple things:

1. What was the reason for dropping `NumpyRNGContext`? 
2. Can you please add your name to AUTHORS.rst? You can do that on the branch you are working on for this and push the commit that adds your name.

AlexN1234:
Ok I think I pushed the commit adding my name to the authors to the pull request.
I dropped NumpyRNGContext because, as far as I can tell, that seeds the old version of the generator, as in, the one you get if you call np.random.randint, for example.  Since the new Numpy Generator creates an individual instance of a rng generator, I don't think there is a way to get it to play nice with NumpyRNGContext.
It appeared to still work without using NumpyRNGContext since each function starts a new generator, the bit generator initial state is the same between them, if that makes sense.

mwcraig (approved):

mwcraig:
> I dropped NumpyRNGContext because, as far as I can tell, that seeds the old version of the generator, as in, the one you get if you call np.random.randint, for example. Since the new Numpy Generator creates an individual instance of a rng generator, I don't think there is a way to get it to play nice with NumpyRNGContext.

That makes sense. 

Thanks again for the pull request -- I'll merge it once the tests have passed.

Issue #849, opened 2024-10-22
`Combiner.sigma_clipping` overwrites existing masks


kbwestfall:
astropy==6.1.4
ccdproc==2.4.2
numpy==2.1.2

The current implementation of `Combine.sigma_clipping` overwrites any existing mask.  Is that intended?

I.e., instead of:

```
        self.data_arr.mask = sigma_clip(self.data_arr.data,
                                        sigma_lower=low_thresh,
                                        sigma_upper=high_thresh,
                                        axis=kwd.get('axis', 0),
                                        copy=kwd.get('copy', False),
                                        maxiters=kwd.get('maxiters', 1),
                                        cenfunc=func,
                                        stdfunc=dev_func,
                                        masked=True,
                                        **kwd).mask
```

would it make sense to instead do:

```
        self.data_arr.mask |= sigma_clip(self.data_arr.data,
                                         sigma_lower=low_thresh,
                                         sigma_upper=high_thresh,
                                         axis=kwd.get('axis', 0),
                                         copy=kwd.get('copy', False),
                                         maxiters=kwd.get('maxiters', 1),
                                         cenfunc=func,
                                         stdfunc=dev_func,
                                         masked=True,
                                         **kwd).mask
```

If so, I can submit a short PR.

mwcraig:
Good catch, I think you are right

PR #873, opened 2025-02-13
Minor masking and flat-field updates

kbwestfall:
This addresses #849 .  It also makes some adjustments to the flat-field correction function to avoid some divide by zero errors and a minor change in the unit book-keeping.

For documentation changes:

For bugfixes:

For new functionality:

### DIFF
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -818,7 +818,9 @@ def flat_correct(ccd, flat, min_value=None, norm_value=None):
 
     # If a norm_value was input and is positive, use it to scale the flat
     if norm_value is not None and norm_value > 0:
-        flat_mean = norm_value if hasattr(norm_value, 'unit') else norm_value * use_flat.unit
+        flat_mean = (
+            norm_value if hasattr(norm_value, "unit") else norm_value * use_flat.unit
+        )
     elif norm_value is not None:
         # norm_value was set to a bad value
         raise ValueError("norm_value must be greater than zero.")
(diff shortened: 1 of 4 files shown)


mwcraig:
@kbwestfall -- just noticed that I need to approve to run the workflows. Looks like there is something to fix....

kbwestfall:
Sorry about that!  I fixed the bug and rebased.

ccdproc/core.py
```diff
@@ -818,18 +818,25 @@ def flat_correct(ccd, flat, min_value=None, norm_value=None):
...
+        flat_mean = (
+            norm_value if hasattr(norm_value, "unit") else norm_value * use_flat.unit
+        )
     elif norm_value is not None:
         # norm_value was set to a bad value
         raise ValueError("norm_value must be greater than zero.")
     else:
         # norm_value was not set, use mean of the image.
-        flat_mean_val = use_flat.data.mean()
+        flat_mean = use_flat.mean()
```
mwcraig, line 829:
```suggestion
        flat_mean = use_flat.data.mean()
```

mwcraig (changes requested):
Just one more minor fix, I think 😀

ccdproc/core.py
kbwestfall: Yeah, I'm seeing where this fails for some versions.  It's important that `flat_mean` have the unit given the other change I made below.  If we want to keep the old dependencies, I think I should change this to:

`flat_mean = use_flat.data.mean() * use_flat.unit`

What do you think?

mwcraig: Good point, I had forgotten about the unit. Your proposal seems like the right fix

mwcraig (approved):
Thanks @kbwestfall!

PR #885, opened 2025-03-19
Adopt the array api

mwcraig:
This PR makes the necessary changes to adopt the array API. There are still some things to sort out:

- [x] Run tests mostly against numpy, but do one run against jax (or some other non-numpy array library)
- [x] Check for remaining invocations of numpy in the code.
- [ ] Check whether changing `.data_arr.mask` to `.data_arr_mask` constitutes an API change. If it does (and I think it does) then patch things up so that  `.data_arr.mask` points to `.data_arr_mask`
- [x] Clean up any remaining handling of immutable arrays ([`array_api_extra`](https://github.com/data-apis/array-api-extra) makes handling those easy)

Once those are done the next step will be to find an outside reviewer for this...

### DIFF
--- a/ccdproc/_ccddata_wrapper_for_array_api.py
+++ b/ccdproc/_ccddata_wrapper_for_array_api.py
@@ -7,7 +7,9 @@ import numpy as np
 from astropy import units as u
 from astropy.nddata import (
     CCDData,
+    InverseVariance,
     StdDevUncertainty,
+    VarianceUncertainty,
 )
 from astropy.nddata.compat import NDDataArray
 from astropy.units import UnitsError
@@ -174,9 +176,98 @@ class _CCDDataWrapperForArrayAPI(CCDData):
             self._mask = np.ma.nomask
 
 
-class _StdDevUncertaintyWrapper(StdDevUncertainty):
+class _CupyOperationNamesMixin:
+    # Override the method below solely to allow CuPy operation names
+    def propagate(self, operation, other_nddata, result_data, correlation, axis=None):
+        """Calculate the resulting uncertainty given an operation on the data.
+
+        .. versionadded:: 1.2
+
+        Parameters
+        ----------
+        operation : callable
+            The operation that is performed on the `NDData`. Supported are
+            `numpy.add`, `numpy.subtract`, `numpy.multiply` and
+            `numpy.true_divide` (or `numpy.divide`).
+
+        other_nddata : `NDData` instance
+            The second operand in the arithmetic operation.
+
+        result_data : `~astropy.units.Quantity` or ndarray
+            The result of the arithmetic operations on the data.
+
+        correlation : `numpy.ndarray` or number
+            The correlation (rho) is defined between the uncertainties in
+            sigma_AB = sigma_A * sigma_B * rho. A value of ``0`` means
+            uncorrelated operands.
+
+        axis : int or tuple of ints, optional
+            Axis over which to perform a collapsing operation.
+
+        Returns
+        -------
+        resulting_uncertainty : `NDUncertainty` instance
+            Another instance of the same `NDUncertainty` subclass containing
+            the uncertainty of the result.
+
+        Raises
+        ------
+        ValueError
+            If the ``operation`` is not supported or if correlation is not zero
+            but the subclass does not support correlated uncertainties.
+
+        Notes
+        -----
+        First this method checks if a correlation is given and the subclass
+        implements propagation with correlated uncertainties.
+        Then the second uncertainty is converted (or an Exception is raised)
+        to the same class in order to do the propagation.
+        Then the appropriate propagation method is invoked and the result is
+        returned.
+        """
+        # Check if the subclass supports correlation
+        if not self.supports_correlated:
+            if isinstance(correlation, np.ndarray) or correlation != 0:
+                raise ValueError(
+                    f"{type(self).__name__} does not support uncertainty propagation"
+                    " with correlation."
+                )
+
+        if other_nddata is not None:
+            # Get the other uncertainty (and convert it to a matching one)
+            other_uncert = self._convert_uncertainty(other_nddata.uncertainty)
+
+            if operation.__name__ in ["add", "cupy_add"]:
+                result = self._propagate_add(other_uncert, result_data, correlation)
+            elif operation.__name__ in ["subtract", "cupy_subtract"]:
+                result = self._propagate_subtract(
+                    other_uncert, result_data, correlation
+                )
+            elif operation.__name__ in ["multiply", "cupy_multiply"]:
+                result = self._propagate_multiply(
+                    other_uncert, result_data, correlation
+                )
+            elif operation.__name__ in [
+                "true_divide",
+                "cupy_true_divide",
+                "divide",
+                "cupy_divide",
+            ]:
+                result = self._propagate_divide(other_uncert, result_data, correlation)
+            else:
+                raise ValueError(f"unsupported operation: {operation.__name__}")
+        else:
+            # assume this is a collapsing operation:
+            result = self._propagate_collapse(operation, axis)
+
+        return self.__class__(array=result, copy=False)
+
+
+class _StdDevUncertaintyWrapper(_CupyOperationNamesMixin, StdDevUncertainty):
     """
-    Override propagate methods to make sure they use the array API.
+    Override operation propagate methods to make sure they use the array API.
+
+    Override overall propagate method to allow cupy_-prefixed operation names.
     """
 
     def _propagate_add(self, other_uncert, result_data, correlation):
@@ -224,18 +315,58 @@ class _StdDevUncertaintyWrapper(StdDevUncertainty):
         )
 
 
+class _VarianceUncertaintyWrapper(_CupyOperationNamesMixin, VarianceUncertainty):
+    """This subclass is needed to allow CuPy operation names"""
+
+
+class _InverseVarianceWrapper(_CupyOperationNamesMixin, InverseVariance):
+    """This subclass is needed to allow CuPy operation names"""
+
+
+def _wrap_uncertainty(unc):
+    """
+    Wrap the uncertainty for a ccd
+    """
+    match unc:
+        case StdDevUncertainty():
+            return _StdDevUncertaintyWrapper(unc)
+        case VarianceUncertainty():
+            return _VarianceUncertaintyWrapper(unc)
+        case InverseVariance():
+            return _InverseVarianceWrapper(unc)
+        case _:
+            raise TypeError("Unsupported uncertainty type")
+
+
+def _unwrap_uncertainty(unc):
+    """
+    Unwrap the uncertainty for a ccd
+    """
+    # Unwrap using the uncertainty's .array so that the uncertainty
+    # setter will automatically handle the units for the uncertainty.
+    match unc:
+        case _StdDevUncertaintyWrapper():
+            return StdDevUncertainty(unc.array)
+        case _VarianceUncertaintyWrapper():
+            return VarianceUncertainty(unc.array)
+        case _InverseVarianceWrapper():
+            return InverseVariance(unc.array)
+        case _:
+            raise TypeError("Unsupported uncertainty type")
+
+
 def _wrap_ccddata_for_array_api(ccd):
     """
     Wrap a CCDData object for use with array API backends.
     """
     if isinstance(ccd, _CCDDataWrapperForArrayAPI):
-        if isinstance(ccd.uncertainty, StdDevUncertainty):
-            ccd.uncertainty = _StdDevUncertaintyWrapper(ccd.uncertainty)
+        if ccd.uncertainty is not None:
+            ccd.uncertainty = _wrap_uncertainty(ccd.uncertainty)
         return ccd
 
     _ccd = _CCDDataWrapperForArrayAPI(ccd)
-    if isinstance(_ccd.uncertainty, StdDevUncertainty):
-        _ccd.uncertainty = _StdDevUncertaintyWrapper(_ccd.uncertainty)
+    if _ccd.uncertainty is not None:
+        _ccd.uncertainty = _wrap_uncertainty(_ccd.uncertainty)
     return _ccd
 
 
@@ -244,8 +375,8 @@ def _unwrap_ccddata_for_array_api(ccd):
     Unwrap a CCDData object from array API backends to the original CCDData.
     """
 
-    if isinstance(ccd.uncertainty, _StdDevUncertaintyWrapper):
-        ccd.uncertainty = StdDevUncertainty(ccd.uncertainty.array)
+    if ccd.uncertainty is not None:
+        ccd.uncertainty = _unwrap_uncertainty(ccd.uncertainty)
 
     if isinstance(ccd, CCDData):
         return ccd


eteq:
As a quick FYI I ran the test suite in this PR on a machine with a NVidia GPU with `CCDPROC_ARRAY_LIBRARY=cupy` and got a whole pile of errors (sent the log directly to @mwcraig but can also share snippets here if someone else wants to see them)

ccdproc/combiner.py
```diff
@@ -132,22 +166,28 @@ def __init__(self, ccd_iter, dtype=None):
...
-        for i, ccd in enumerate(ccd_list):
-            self.data_arr[i] = ccd.data
+        mask_list = []
+        for ccd in ccd_list:
             if ccd.mask is not None:
-                self.data_arr.mask[i] = ccd.mask
+                mask_list.append(ccd.mask)
             else:
-                self.data_arr.mask[i] = ma.zeros(default_shape)
+                mask_list.append(xp.zeros(default_shape))
```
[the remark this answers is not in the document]
nstarman: If mask is a computed attribute and the computation is "expensive" then this can be sped up by 

```python
mask_list = [(m := ccd.mask) if m is not None else xp.zeros(default_shape) for ccd in ccd_list]
```

pyproject.toml
```diff
@@ -8,17 +8,19 @@ dynamic = ["version"]
 description = "Astropy affiliated package"
 readme = "README.rst"
 license = { text = "BSD-3-Clause" }
-requires-python = ">=3.8"
+requires-python = ">=3.11"
```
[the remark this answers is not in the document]
nstarman: https://scientific-python.org/specs/spec-0000/

ccdproc/core.py
```diff
@@ -441,7 +441,13 @@ def create_deviation(ccd_data, gain=None, readnoise=None, disregard_nan=False):
...
 
     # calculate the deviation
-    var = (xp.sqrt(data) ** 2 + readnoise_value**2) ** 0.5
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            message="invalid value encountered in sqrt",
+            category=RuntimeWarning,
+        )
+        var = (xp.sqrt(data) ** 2 + readnoise_value**2) ** 0.5
```
nstarman, line 450:
```suggestion
        var = xp.sqrt(xp.abs(data) + readnoise_value**2)
```

pyproject.toml
```diff
@@ -145,6 +148,7 @@ filterwarnings= [
     "ignore:numpy\\.ufunc size changed:RuntimeWarning",
     "ignore:numpy.ndarray size changed:RuntimeWarning",
     "ignore:`np.bool` is a deprecated alias for the builtin `bool`:DeprecationWarning",
+    "ignore:invalid value encountered in sqrt:RuntimeWarning",
```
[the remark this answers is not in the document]
nstarman: ✅

pyproject.toml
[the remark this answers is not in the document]
mwcraig: So the move here would be to bump it up to 3.12?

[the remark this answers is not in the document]
nstarman: 3.11 for a few more months. Then 3.12. I was also thinking about adding the badge saying "Spec 0", e.g. you can see in the README in https://github.com/GalacticDynamics/unxt

ccdproc/core.py
mwcraig: Turns out this is not equivalent in the case when `data` is negative -- then `sqrt` generates a `nan` which we want..

pyproject.toml
```diff
@@ -8,17 +8,19 @@ dynamic = ["version"]
...
-requires-python = ">=3.8"
+requires-python = ">=3.11"
 authors = [
     { name = "Steve Crawford", email = "ccdproc@gmail.com" },
     { name = "Matt Craig" },
     { name = "and Michael Seifert" },
 ]
 dependencies = [
-    "astropy>=5.0.1",
+    "array_api_compat",
```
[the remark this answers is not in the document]
mwcraig: Pinned it

ccdproc/combiner.py
```diff
@@ -2,52 +2,89 @@
...
+from numpy import ravel_multi_index as np_ravel_multi_index
 
 from .core import sigma_func
 
 __all__ = ["Combiner", "combine"]
 
 
-def _default_median():  # pragma: no cover
+def _default_median(xp=None):  # pragma: no cover
     if HAS_BOTTLENECK:
```
nstarman, line 27: Does Bottleneck work with the array api? If not, would it be better to change the order of this logic and only use bottleneck if the namespace is numpy / None?

ccdproc/combiner.py
```diff
@@ -2,52 +2,89 @@
...
+        return xp.nanmedian
+    except AttributeError as e:
+        raise RuntimeError(
+            "No NaN-aware median function available. Please install bottleneck."
+        ) from e
 
 
-def _default_average():  # pragma: no cover
+def _default_average(xp=None):  # pragma: no cover
     if HAS_BOTTLENECK:
```
nstarman, line 43: Ibid

ccdproc/combiner.py
```diff
@@ -2,52 +2,89 @@
...
+        return xp.nanmean
+    except AttributeError as e:
+        raise RuntimeError(
+            "No NaN-aware mean function available. Please install bottleneck."
+        ) from e
 
-def _default_sum():  # pragma: no cover
+
+def _default_sum(xp=None):  # pragma: no cover
     if HAS_BOTTLENECK:
```
nstarman, line 59: Ibid

ccdproc/combiner.py
```diff
@@ -2,52 +2,89 @@
...
+        return xp.nansum
+    except AttributeError as e:
+        raise RuntimeError(
+            "No NaN-aware sum function available. Please install bottleneck."
+        ) from e
 
 
-def _default_std():  # pragma: no cover
+def _default_std(xp=None):  # pragma: no cover
     if HAS_BOTTLENECK:
```
nstarman, line 75: Ibid

ccdproc/combiner.py
```diff
@@ -72,6 +109,12 @@ class Combiner:
         description. If ``None`` it uses ``np.float64``.
         Default is ``None``.
 
+    xp : array namespace, optional
+        The array namespace to use for the data. If not provided, it will
+        be inferred from the first `~astropy.nddata.CCDData` object in
+        ``ccd_iter``.
+        Default is ``None``.
```
nstarman, line 116:
```suggestion
    xp : array namespace, optional
        The array namespace to use for the data. If `None` or not provided, it will
        be inferred from the first `~astropy.nddata.CCDData` object in
        ``ccd_iter``.
        Default is `None`.
```

ccdproc/combiner.py
```diff
@@ -99,15 +142,12 @@ class Combiner:
                  [ 0.66666667,  0.66666667,  0.66666667,  0.66666667]]...)
     """
 
-    def __init__(self, ccd_iter, dtype=None):
+    def __init__(self, ccd_iter, dtype=None, xp=None):
```
nstarman, line 145: Not necessarily for this PR, but the validation logic in the later for-loop can be improved, e.g.

```python
ccd_list = list(ccd_iter)
if not all(isinstance(x, CCDData) for x in ccd_list):
    raise TypeError...

default_shape = ccd_list[0].shape
if not all(x.shape == default_shape):
    raise TypeError...

default_unit = ...
```

ccdproc/combiner.py
```diff
@@ -132,31 +172,47 @@ def __init__(self, ccd_iter, dtype=None):
                 if not (default_unit == ccd.unit):
                     raise TypeError("CCDData objects don't have the same unit.")
 
+        # Set array namespace
+        xp = xp or array_api_compat.array_namespace(ccd_list[0].data)
+        self._xp = xp
+        if dtype is None:
+            dtype = xp.float64
         self.ccd_list = ccd_list
```
nstarman, line 180: Having both ccd_list and _data_arr can double the memory footprint. Are both needed?

ccdproc/log_meta.py
```diff
@@ -134,7 +134,7 @@ def _replace_array_with_placeholder(value):
     return_type_not_value = False
     if isinstance(value, u.Quantity):
         return_type_not_value = not value.isscalar
-    elif isinstance(value, (NDData, np.ndarray)):
+    elif isinstance(value, NDData | np.ndarray):
```
nstarman, line 137: This is actually slower than
```suggestion
    elif isinstance(value, (NDData, np.ndarray)):
```

ccdproc/combiner.py
mwcraig: Yeah, good call. I can check later what libraries bottlneck works with. Mostly I don't want to change the behavior for current users

ccdproc/tests/test_cosmicray.py
```diff
@@ -21,17 +26,22 @@
...
+    crrays = xp.asarray(rng.integers(0, size, size=(ncrays, 2)))
     # use (threshold + 15) below to make sure cosmic ray is well above the
     # threshold no matter what the random number generator returns
     # add_cosmicrays is highly sensitive to the seed
     # ideally threshold should be set so it is not sensitive to seed, but
     # this is not working right now
-    crflux = 10 * scale * rng.random(NCRAYS) + (threshold + 15) * scale
+    crflux = xp.asarray(10 * scale * rng.random(ncrays) + (threshold + 15) * scale)
+
+    # Some array libraries (Dask) do not support setting individual elements,
```
nstarman, line 38: Does `array_api_extras` address this? I haven't looked into the details

ccdproc/tests/test_cosmicray.py
```diff
@@ -21,17 +26,22 @@
...
+    # so use NumPy.
+    data_as_np = np_array(data.data)
     for i in range(ncrays):
         y, x = crrays[i]
-        data.data[y, x] = crflux[i]
+        data_as_np[y, x] = crflux[i]
+    data.data = xp.asarray(data_as_np)
```
nstarman, line 44: Does this work for vectorized assignment?

```python
data_as_np = np_array(data.data)
y, x = crrays.T
data_as_np[y, x] = crflux
data.data = xp.asarray(data_as_np)
```

ccdproc/tests/test_cosmicray.py
```diff
@@ -50,8 +60,10 @@ def test_cosmicray_lacosmic_ccddata():
     ccd_data = ccd_data_func(data_scale=DATA_SCALE)
     threshold = 5
     add_cosmicrays(ccd_data, DATA_SCALE, threshold, ncrays=NCRAYS)
-    noise = DATA_SCALE * np.ones_like(ccd_data.data)
-    ccd_data.uncertainty = noise
+    noise = DATA_SCALE * xp.ones_like(ccd_data.data)
+    # Workaround for the fact that upstream checks for numpy array
```
nstarman, line 64: upstream is Astropy? So we can fix this to be array-namespace.

mwcraig:
> Hi @mwcraig. LMK how I can be of help with reviews. This PR looks great, and it seems to be mostly down to small line-by-line suggestions.

Thanks, @nstarman! I'll want one last look once I've tied down a few loose ends. One thing I learned since the coordination meeting is that using `np.testing.assert_allclose` is a bad idea because it converts any array-like arguments into numpy arrays for the comparison, hiding any mis-match in array name space between the two things being compared. As a result, there are still a few places that I thought were working that were actually converting the data or uncertainty or mask to numpy arrays.

I've switch to using `xp.all(xpx.isclose(a, b))` for numerical array comparisons, which raises an exception if `a` and `b` are in different namespaces.

I'll ping you when I'm confident that I have all of those cases fixed up.

ccdproc/tests/test_cosmicray.py
mwcraig: That raised an error...given that this is in a test I'm just going to leave it as a loop.

ccdproc/tests/test_cosmicray.py
mwcraig: It turns out the answer is no, but not for the reason the comment suggests. Using `xpx.at(...` raises a exception with the message `NotImplementedError: dask does not yet implement assignment to a scalar numpy or dask array index`. I think `array_api_extra` must convert the index to an array, and dask doesn't support indexing by an array.

Just using regular indexing (i.e. not `xpx.at`) works for everything except `jax`, where the arrays are immutable. 

After tinkering for a bit I think I'll just leave it as is.

ccdproc/tests/test_cosmicray.py
mwcraig: Found a workaround.

mwcraig:
@nstarman  -- this is ready for one more look. The only remaining item is how to handle bottleneck vs the array libraries if both are present. I'm leaning towards doing what you suggested: bottleneck if the array library is `numpy` or unspecified, or if the array namespace has no `nan...` as long as there are actually nans in the array.

Now that I write that out, I'm inclined to open an issue for this and to fix it separately.

ccdproc/combiner.py
mwcraig: Opened an issue to track this: https://github.com/astropy/ccdproc/issues/904

ccdproc/tests/pytest_fixtures.py
```diff
@@ -59,7 +62,7 @@ def ccd_data(
     data = rng.normal(loc=mean, size=[size, size], scale=scale)
 
     fake_meta = {"my_key": 42, "your_key": "not 42"}
-    ccd = CCDData(data, unit=u.adu)
+    ccd = CCDData(array_library.asarray(data), unit=u.adu)
```
mwcraig, line 65:
```suggestion
    ccd = CCDData(xp.asarray(data), unit=u.adu)
```

ccdproc/tests/test_ccdproc.py
```diff
@@ -641,7 +724,7 @@ def test_gain_correct():
     ccd_data = ccd_data_func()
     init_data = ccd_data.data
     gain_data = gain_correct(ccd_data, gain=3, add_keyword=None)
-    np.testing.assert_allclose(gain_data.data, 3 * init_data)
+    assert xp.all(xpx.isclose(gain_data.data, 3 * init_data))
```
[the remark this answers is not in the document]
mwcraig: I think you are right, though internally I flinched a bit at the idea 😬. At the end of the day they should be easy to find/replace/

ccdproc/tests/test_cosmicray.py
```diff
@@ -20,18 +25,25 @@
 
 
 def add_cosmicrays(data, scale, threshold, ncrays=NCRAYS):
+    from numpy import array as np_array
```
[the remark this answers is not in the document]
mwcraig: I hid this in here to make sure its scope was just this function -- this may always remain a numpy thing since I see it as setting up the input data for a bunch of tests that are the ones actually testing the array API. No need to be able to generate the data solely via the Array API, since I don't think things like random number generation will every land there.

docs/array_api.rst
```diff
@@ -0,0 +1,55 @@
...
+on a GPU using `CuPy`_ should be
+`reported as an issue <https://github.com/astropy/ccdproc/issues>`_. Though the
+`Sparse`_ array library supports the array API, `ccdproc`_ does not currently work
+with `sparse_`. A `pull request <https://github.com/astropy/ccdproc/pulls>` to FIX
+that would be a welcome contribution to the project.
+
+What limitations shuold I be aware of?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
++ The ``median`` function is not part of the array API, but most array libraries
```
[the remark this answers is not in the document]
mwcraig: I _think_, though I need to verify, that all of the array libraries can at least do a median along and axis, which is the main thing needed in combiner. I'll take a look at this after merging, tracking in #906

ccdproc/combiner.py
mwcraig: Opened an issue to track this: https://github.com/astropy/ccdproc/issues/904

mwcraig:
@nstarman -- thanks again for the thorough review! Can you please give this comment a thumbs-up if you think this is ready for merge?

nstarman:
🎉

PR #923, opened 2026-06-02
Remove scikit-image runtime dependency

karlhillx:
ccdproc's block_reduce/block_average/block_replicate wrappers all delegate to astropy.nddata. scikit-image was never imported at runtime -- only in the test file, and only for a version-check skip condition that used broken string comparison.

Removed it from runtime deps and cleaned up the test imports.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -37,6 +37,7 @@ Alphabetical list of code contributors
 * Yash Gondhalekar (@Yash-10)
 * Hans Moritz Günther (@hamogu)
 * Nathan Heidt (@heidtha)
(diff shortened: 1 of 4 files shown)


mwcraig (commented):
Thanks for the quick fix @karlhillx -- can you add yourself to the list of contributors or would you like me to do that for you?

karlhillx:
> Thanks for the quick fix @karlhillx -- can you add yourself to the list of contributors or would you like me to do that for you?

Done, just pushed a commit adding myself. Thanks Matt!

mwcraig:
Thanks @karlhillx!

Issue #941, opened 2026-07-07
array-api-strict triage: escape-site mapping of the remaining test failures


mwcraig:
### Context

The tooling from #937 and #939 is implemented on branch [`array-api-strict-triage-tooling`](https://github.com/mwcraig/ccdproc/tree/array-api-strict-triage-tooling) (mwcraig fork, commit 94ca9ec):

- `CCDPROC_ARRAY_LIBRARY=array-api-strict` test backend with fixture arrays placed on the non-default `Device("device1")`, so any implicit conversion to numpy raises — the same failure signal as CuPy's device-to-host error, with no GPU needed. Includes a `py313-strict` tox env and CI job.
- Escape-site dedupe report (`CCDPROC_TRIAGE_ESCAPES=1`): groups test failures by the innermost non-test ccdproc frame.
- `backend_xfail` / `backend_skip` markers applied per backend; documented limitations (astroscrappy, `scipy.ndimage.median_filter`, reproject, `block_reduce`/`block_replicate`) are now xfailed under array-api-strict.
- Escape logger (`CCDPROC_LOG_ARRAY_ESCAPES=1`): warns when `np.asarray`/`np.asanyarray`/`np.ma.asanyarray` receive a non-numpy array-API array, catching backends that convert silently (dask, JAX).

### Results

Full suite under `CCDPROC_ARRAY_LIBRARY=array-api-strict`: **139 failed / 192 passed / 5 skipped**. After marking documented limitations: **109 failed / 30 xfailed**. As predicted in #939, the dedupe report collapses the failures to about ten distinct escape sites.

### Escape-site → root-cause mapping

| Escape site | Failures | Root cause | Tracked in |
|---|---|---|---|
| `combiner.py:189` `Combiner.__init__` | 44 | Stacks per-image arrays by building a nested Array; array-api-strict rejects this ("Use stack instead"). Likely also the biggest CuPy failure source. | #965 |
| `_ccddata_wrapper_for_array_api.py:74` `_arithmetic_wrapper` | 11 | Device-mismatch / dtype coercion inside the CCDData arithmetic wrapper | possibly related to #927 |
| combiner scaling tests | ~9 | `Quantity` leaking into array-API arithmetic (`TypeError: Expected Array or Python scalar`) | #936 |
| `test_ccdproc.py:174` `test_subtract_overscan` | 6 | numpy-only `.copy()` method called on an Array (test-only) | #969 |
[...]

PR #942, opened 2026-07-07
Add array-api-strict test backend and array-escape triage tooling

### LINKED ISSUE #937: Add array-api-strict (non-default device) test backend as a CPU proxy for CuPy
### Proposal

Add `array-api-strict` as a test backend, configured to place arrays on a **non-default device**.

Why it is a good CPU-side proxy for CuPy:

- array-api-strict (≥ 2.0) simulates multiple devices; arrays on a non-default device **raise on conversion to numpy** — the same failure signal as CuPy's implicit device-to-host `TypeError`, with no GPU required.
- It also rejects any usage outside the array API spec (e.g. `arr.astype(...)`, `np`-style dtype strings), catching portability bugs that numpy/CuPy/JAX all tolerate.

### Work items

- [ ] Wire `array-api-strict` into the `CCDPROC_ARRAY_LIBRARY` mechanism in `conftest.py` / `pytest_fixtures.py`, including device placement of fixture arrays.
- [ ] Add a tox environment and CI job.
- [ ] Triage the initial failures: each should map to a root cause in #910–#913 or to a documented limitation (xfail/skip marker).

A failure that reproduces under both array-api-strict and `sparse` (#938) is almost certainly a real CuPy bug, so these two backends together let most of #909 be worked without GPU hardware.

---
Follow-up to #909. Found during a review of the array API implementation from #885.

### LINKED ISSUE #939: Triage tooling for array-library escapes: traceback dedupe, xfail markers, escape logger
### Proposal

Three pieces of tooling to turn the CuPy failure lists (#910–#913) into a tractable set of root causes and keep them fixed:

1. **Escape-site dedupe script**: run the failing suite (`CCDPROC_ARRAY_LIBRARY=cupy` or a proxy backend) with `--tb=long`, and group failures by the innermost non-test frame in each traceback. CuPy's `Implicit conversion` error is raised from `cupy.ndarray.__array__`, so the traceback names the exact call site of every escape. The ~60 failures across #910–#913 should collapse to well under ten distinct sites; record the mapping on each issue.

2. **Per-backend xfail/skip markers**: mark tests of documented limitations (astroscrappy, reproject, scipy.ndimage-based functions, …) per backend so a CuPy or proxy run reports only *regressions*, not known gaps.

3. **Escape-logger fixture**: a conftest fixture that monkeypatches `np.asarray` / `np.asanyarray` / `np.ma.MaskedArray` to log a stack trace whenever they receive a non-numpy array. This catches escapes even for backends that convert *silently* (Dask, JAX) instead of erroring like CuPy.

---
Follow-up to #909–#913. Found during a review of the array API implementation from #885.

mwcraig:
Closes #937
Closes #939

Adds the array-api-strict test backend and the array-escape triage tooling from the array-API review follow-ups. The triage results produced by this tooling are recorded in #941.

### array-api-strict backend (#937)

- `CCDPROC_ARRAY_LIBRARY=array-api-strict` (or `array_api_strict`) selects `array_api_strict` as the testing array library, with fixture arrays placed on the non-default `Device("device1")`. Conversion to numpy then raises — the same failure signal as CuPy's implicit device-to-host conversion, with no GPU required. The device can be overridden via `CCDPROC_ARRAY_DEVICE`.
- New `py313-strict` tox environment and `ubuntu-py313-strict` CI job.
- Documented in `docs/array_api.rst`.

### Triage tooling (#939)

- **Escape-site dedupe report** (`CCDPROC_TRIAGE_ESCAPES=1`): groups test failures by the innermost non-test ccdproc frame and prints a ranked summary at session end (`ccdproc/tests/_escape_triage.py`).
- **Per-backend markers**: `@pytest.mark.backend_xfail("cupy", "array-api-strict", reason=...)` and `@pytest.mark.backend_skip(...)`, applied via `pytest_collection_modifyitems`, so backend runs report only regressions rather than known gaps.
- **Escape logger** (`CCDPROC_LOG_ARRAY_ESCAPES=1`): monkeypatches `np.asarray`/`np.asanyarray`/`np.ma.asanyarray` to warn (with the ccdproc call site) whenever they receive a non-numpy array-API array — catches backends like dask and JAX that convert silently instead of raising like CuPy.

Tests of documented limitations (astroscrappy, `scipy.ndimage.median_filter`, reproject, `block_reduce`/`block_replicate`) are marked `backend_xfail` for array-api-strict.

### Verification

- numpy backend: 331 passed, 5 skipped — unchanged from main.
- array-api-strict backend: 139 failures before markers → 109 failed / 30 xfailed after; the dedupe report collapses the remaining failures to ~10 distinct escape sites, mapped to root causes in #941.
- `ruff check` and pre-commit hooks pass.

https://claude.ai/code/session_01Qri8rDgkWjCT7Yb7QKJrhf

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,15 +1,15 @@
-Unreleased
-----------
+2.6.0 (unreleased)
+------------------
 
 New Features
 ^^^^^^^^^^^^
 
 - Add ``array-api-strict`` (on a non-default device) as a test backend, a
-  CPU-only proxy for CuPy device behavior. [#937]
+  CPU-only proxy for CuPy device behavior. [#942]
 - Add triage tooling for array-API backend failures: ``backend_xfail`` /
   ``backend_skip`` markers, an escape-site report
   (``CCDPROC_TRIAGE_ESCAPES=1``), and an escape logger
-  (``CCDPROC_LOG_ARRAY_ESCAPES=1``). [#939]
+  (``CCDPROC_LOG_ARRAY_ESCAPES=1``). [#942]
 
 2.5.1 (2025-07-05)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,16 +4,12 @@ Unreleased
 New Features
 ^^^^^^^^^^^^
 
-- Added support for testing against ``array-api-strict`` on a non-default
-  device, which serves as a CPU-only proxy for CuPy-style GPU device
-  behavior. Set ``CCDPROC_ARRAY_LIBRARY=array-api-strict`` (and optionally
-  ``CCDPROC_ARRAY_DEVICE``) to use it. [#937]
-- Added test-infrastructure tooling for triaging array-API backend
-  failures: ``@pytest.mark.backend_xfail``/``@pytest.mark.backend_skip``
-  markers for per-backend expected failures, an escape-site dedupe report
-  enabled with ``CCDPROC_TRIAGE_ESCAPES=1``, and an array-escape logger
-  enabled with ``CCDPROC_LOG_ARRAY_ESCAPES=1`` that flags silent conversion
-  of non-numpy arrays back to numpy. [#939]
+- Add ``array-api-strict`` (on a non-default device) as a test backend, a
+  CPU-only proxy for CuPy device behavior. [#937]
+- Add triage tooling for array-API backend failures: ``backend_xfail`` /
+  ``backend_skip`` markers, an escape-site report
+  (``CCDPROC_TRIAGE_ESCAPES=1``), and an escape logger
+  (``CCDPROC_LOG_ARRAY_ESCAPES=1``). [#939]
 
 2.5.1 (2025-07-05)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,20 @@
+Unreleased
+----------
+
+New Features
+^^^^^^^^^^^^
+
+- Added support for testing against ``array-api-strict`` on a non-default
+  device, which serves as a CPU-only proxy for CuPy-style GPU device
+  behavior. Set ``CCDPROC_ARRAY_LIBRARY=array-api-strict`` (and optionally
+  ``CCDPROC_ARRAY_DEVICE``) to use it. [#937]
+- Added test-infrastructure tooling for triaging array-API backend
+  failures: ``@pytest.mark.backend_xfail``/``@pytest.mark.backend_skip``
+  markers for per-backend expected failures, an escape-site dedupe report
+  enabled with ``CCDPROC_TRIAGE_ESCAPES=1``, and an array-escape logger
+  enabled with ``CCDPROC_LOG_ARRAY_ESCAPES=1`` that flags silent conversion
+  of non-numpy arrays back to numpy. [#939]
+
 2.5.1 (2025-07-05)
 ------------------
 
--- a/ccdproc/conftest.py
+++ b/ccdproc/conftest.py
@@ -28,6 +28,7 @@ from .tests._escape_triage import (
     _env_truthy,
     pytest_runtest_makereport,  # noqa: F401 pytest hook, used via attribute lookup
     pytest_sessionfinish,  # noqa: F401 pytest hook, used via attribute lookup
+    pytest_sessionstart,  # noqa: F401 pytest hook, used via attribute lookup
     pytest_terminal_summary,  # noqa: F401 pytest hook, used via attribute lookup
     record_escape_log,
 )
--- a/ccdproc/conftest.py
+++ b/ccdproc/conftest.py
@@ -7,7 +7,7 @@ import logging
 import os
 import traceback
 
-import array_api_compat  # noqa: F401
+import array_api_compat
 import array_api_compat.numpy
 import numpy as np
 import pytest
@@ -153,8 +153,8 @@ def pytest_collection_modifyitems(items):
 # backends like dask and jax the conversion often succeeds silently, so the
 # bug is easy to miss. This monkeypatches np.asarray/np.asanyarray (and
 # np.ma.asanyarray) to log a warning, when active, any time they are handed
-# an array that reports its own `__array_namespace__` and that namespace is
-# not numpy.
+# an array whose array-API namespace (as resolved by
+# array_api_compat.array_namespace) is not numpy.
 #
 # Activated by setting the environment variable CCDPROC_LOG_ARRAY_ESCAPES to
 # a truthy value.
@@ -171,31 +171,31 @@ _LOG_ARRAY_ESCAPES = _env_truthy(os.environ.get("CCDPROC_LOG_ARRAY_ESCAPES", "")
 _NUMPY_LIKE_NAMESPACES = {np, array_api_compat.numpy}
 
 
-def _is_foreign_array(obj):
+def _foreign_namespace(obj):
     """
-    Return True if obj is an array from a non-numpy array-API library.
+    Return obj's array-API namespace if it is foreign, else None.
 
-    numpy arrays (including np.ma masked arrays) are never foreign. Anything
-    else that implements the array-API protocol marker __array_namespace__()
-    is foreign unless its namespace is numpy or array_api_compat's numpy
-    wrapper -- i.e. exactly the arrays whose conversion to numpy would fail
-    (CuPy on GPU) or silently densify/transfer (dask, jax). The try/except
-    guards against objects whose __array_namespace__() raises; those are
-    treated as not foreign so the logger never breaks the call it wraps.
+    numpy arrays (including np.ma masked arrays) are never foreign.
+    Detection goes through array_api_compat.array_namespace() rather than
+    the raw __array_namespace__ dunder because some backends (notably dask)
+    do not define the dunder on their array objects even though
+    array-api-compat can resolve a namespace for them. The namespace is
+    foreign unless it is numpy or array_api_compat's numpy wrapper -- i.e.
+    exactly the arrays whose conversion to numpy would fail (CuPy on GPU)
+    or silently densify/transfer (dask, jax). The try/except guards against
+    non-arrays and misbehaving objects (including unhashable namespaces);
+    those are treated as not foreign so the logger never breaks the call
+    it wraps.
     """
     if isinstance(obj, np.ndarray):
-        return False
-
-    get_namespace = getattr(obj, "__array_namespace__", None)
-    if get_namespace is None:
-        return False
-
+        return None
     try:
-        namespace = get_namespace()
+        namespace = array_api_compat.array_namespace(obj)
+        if namespace in _NUMPY_LIKE_NAMESPACES:
+            return None
     except Exception:
-        return False
-
-    return namespace not in _NUMPY_LIKE_NAMESPACES
+        return None
+    return namespace
 
 
 def _describe_escape_site():
@@ -227,14 +227,15 @@ def _make_escape_logging_wrapper(original, funcname, guard):
             guard.active = True
             try:
                 obj = args[0]
-                if _is_foreign_array(obj):
+                namespace = _foreign_namespace(obj)
+                if namespace is not None:
                     site = _describe_escape_site()
                     _escape_logger.warning(
                         "array-API escape: %s() called on a %r array "
                         "(namespace=%r) at %s",
                         funcname,
                         type(obj),
-                        obj.__array_namespace__(),
+                        namespace,
                         site,
                     )
             finally:
--- a/ccdproc/conftest.py
+++ b/ccdproc/conftest.py
@@ -27,7 +27,9 @@ except ImportError:
 from .tests._escape_triage import (
     _env_truthy,
     pytest_runtest_makereport,  # noqa: F401 pytest hook, used via attribute lookup
+    pytest_sessionfinish,  # noqa: F401 pytest hook, used via attribute lookup
     pytest_terminal_summary,  # noqa: F401 pytest hook, used via attribute lookup
+    record_escape_log,
 )
 from .tests.pytest_fixtures import (
     triage_setup,  # noqa: F401 this is used in tests
@@ -198,20 +200,24 @@ def _foreign_namespace(obj):
     return namespace
 
 
-def _describe_escape_site():
+def _escape_site_frame():
     """
-    Return a short "file:line function" string for the innermost stack
-    frame that is inside ccdproc but not inside ccdproc's own test suite,
-    for use in escape log messages.
+    Return the FrameSummary for the innermost stack frame that is inside
+    ccdproc but not inside ccdproc's own test suite -- the frame blamed for
+    an escape -- or None if there is no such frame. Frames in this module are
+    classified as test frames and are never chosen, so the exact call depth
+    here does not affect the result.
     """
     from .tests._escape_triage import locate_escape_site
 
-    # Drop this function's own frame before searching.
-    frames = traceback.extract_stack()[:-1]
-    site = locate_escape_site(frames)
-    if site is None:
+    return locate_escape_site(traceback.extract_stack())
+
+
+def _describe_escape_site(frame):
+    """Short "file:line function" string for an escape log message."""
+    if frame is None:
         return "<unknown location>"
-    return f"{site.filename}:{site.lineno} {site.name}"
+    return f"{frame.filename}:{frame.lineno} {frame.name}"
 
 
 class _ReentrancyGuard:
@@ -229,7 +235,8 @@ def _make_escape_logging_wrapper(original, funcname, guard):
                 obj = args[0]
                 namespace = _foreign_namespace(obj)
                 if namespace is not None:
-                    site = _describe_escape_site()
+                    frame = _escape_site_frame()
+                    site = _describe_escape_site(frame)
                     _escape_logger.warning(
                         "array-API escape: %s() called on a %r array "
                         "(namespace=%r) at %s",
@@ -238,6 +245,7 @@ def _make_escape_logging_wrapper(original, funcname, guard):
                         namespace,
                         site,
                     )
+                    record_escape_log(frame, funcname)
             finally:
                 guard.active = False
         return original(*args, **kwargs)
--- a/ccdproc/conftest.py
+++ b/ccdproc/conftest.py
@@ -3,9 +3,14 @@
 # this contains imports plugins that configure py.test for astropy tests.
 # by importing them here in conftest.py they are discoverable by py.test
 # no matter how it is invoked within the source tree.
+import logging
 import os
+import traceback
 
 import array_api_compat  # noqa: F401
+import array_api_compat.numpy
+import numpy as np
+import pytest
 
 try:
     # When the pytest_astropy_header package is installed
@@ -19,6 +24,10 @@ except ImportError:
     TESTED_VERSIONS = {}
 
 
+from .tests._escape_triage import (
+    pytest_runtest_makereport,  # noqa: F401 pytest hook, used via attribute lookup
+    pytest_terminal_summary,  # noqa: F401 pytest hook, used via attribute lookup
+)
 from .tests.pytest_fixtures import (
     triage_setup,  # noqa: F401 this is used in tests
 )
@@ -41,6 +50,14 @@ PYTEST_HEADER_MODULES.pop("h5py", None)
 # What happens here is controlled by an environmental variable
 array_library = os.environ.get("CCDPROC_ARRAY_LIBRARY", "numpy").lower()
 
+# Device to create test arrays on. This is only meaningful for backends that
+# support multiple devices (currently array-api-strict, as a CPU-only proxy
+# for testing non-default-device behavior like CuPy's GPU device). It can be
+# overridden with the CCDPROC_ARRAY_DEVICE environment variable, whose value
+# is passed to the array library's Device constructor. "default" (or leaving
+# it unset) means the library's usual default device.
+testing_array_device = None
+
 match array_library:
     case "numpy":
         import array_api_compat.numpy as testing_array_library  # noqa: F401
@@ -60,8 +77,197 @@ match array_library:
 
         PYTEST_HEADER_MODULES["cupy"] = "cupy"
 
+    case "array-api-strict" | "array_api_strict":
+        import array_api_strict as testing_array_library  # noqa: F401
+
+        PYTEST_HEADER_MODULES["array_api_strict"] = "array_api_strict"
+
+        # array-api-strict exposes a couple of extra fake devices in addition
+        # to its default CPU_DEVICE. Using one of those non-default devices
+        # here makes np.asarray() raise on the resulting arrays, the same way
+        # it would for a CuPy array living on a GPU. That makes
+        # array-api-strict a convenient CPU-only proxy for catching the
+        # "silent conversion to numpy" bugs that would otherwise only show up
+        # on CuPy.
+        device_name = os.environ.get("CCDPROC_ARRAY_DEVICE", "default").lower()
+        if device_name != "default":
+            testing_array_device = testing_array_library.Device(device_name)
+        else:
+            testing_array_device = testing_array_library.Device("device1")
+
     case _:
         raise ValueError(
             f"Unsupported array library: {array_library}. "
             "Supported libraries are listed at https://ccdproc.readthedocs.io/en/latest/array_api.html."
         )
+
+
+def _env_truthy(value):
+    return str(value).strip().lower() in {"1", "true", "yes", "on"}
+
+
+# ---------------------------------------------------------------------------
+# Per-backend xfail/skip markers
+#
+# @pytest.mark.backend_xfail("cupy", "array-api-strict", reason="...")
+# @pytest.mark.backend_skip("cupy", reason="...")
+#
+# These let individual tests be marked as expected-failures or skips only
+# when run against specific array backends (as set by CCDPROC_ARRAY_LIBRARY),
+# without affecting the default numpy-backed test run.
+# ---------------------------------------------------------------------------
+
+
+def _normalize_backend_name(name):
+    return name.lower().replace("_", "-")
+
+
+_ACTIVE_BACKEND = _normalize_backend_name(array_library)
+
+
+def pytest_collection_modifyitems(items):
+    for item in items:
+        for marker in item.iter_markers(name="backend_skip"):
+            backends = {_normalize_backend_name(b) for b in marker.args}
+            if _ACTIVE_BACKEND in backends:
+                reason = marker.kwargs.get(
+                    "reason", f"skipped for array backend {_ACTIVE_BACKEND!r}"
+                )
+                item.add_marker(pytest.mark.skip(reason=reason))
+                break
+
+        for marker in item.iter_markers(name="backend_xfail"):
+            backends = {_normalize_backend_name(b) for b in marker.args}
+            if _ACTIVE_BACKEND in backends:
+                reason = marker.kwargs.get(
+                    "reason", f"expected failure for array backend {_ACTIVE_BACKEND!r}"
+                )
+                item.add_marker(pytest.mark.xfail(reason=reason, strict=False))
+                break
+
+
+# ---------------------------------------------------------------------------
+# Escape logger: catch silent conversion of non-numpy array-API arrays back
+# to numpy. On CuPy such a conversion typically raises immediately (because
+# the array lives on a GPU), which is how those bugs are usually found. On
+# backends like dask and jax the conversion often succeeds silently, so the
+# bug is easy to miss. This monkeypatches np.asarray/np.asanyarray (and
+# np.ma.asanyarray) to log a warning, when active, any time they are handed
+# an array that reports its own `__array_namespace__` and that namespace is
+# not numpy.
+#
+# Activated by setting the environment variable CCDPROC_LOG_ARRAY_ESCAPES to
+# a truthy value.
+# ---------------------------------------------------------------------------
+
+_escape_logger = logging.getLogger("ccdproc.array_escape")
+
+_LOG_ARRAY_ESCAPES = _env_truthy(os.environ.get("CCDPROC_LOG_ARRAY_ESCAPES", ""))
+
+# array_api_compat.numpy wraps plain numpy; arrays built through it report
+# __array_namespace__() as one of these two modules. Neither counts as an
+# "escape" -- we only care about arrays from a genuinely different library
+# (jax, dask, cupy, array_api_strict, ...) ending up in a numpy-only call.
+_NUMPY_LIKE_NAMESPACES = {np, array_api_compat.numpy}
+
+
+def _is_foreign_array(obj):
+    if isinstance(obj, np.ndarray):
+        return False
+
+    get_namespace = getattr(obj, "__array_namespace__", None)
+    if get_namespace is None:
+        return False
+
+    try:
+        namespace = get_namespace()
+    except Exception:
+        return False
+
+    return namespace not in _NUMPY_LIKE_NAMESPACES
+
+
+def _describe_escape_site():
+    """
+    Return a short "file:line function" string for the innermost stack
+    frame that is inside ccdproc but not inside ccdproc's own test suite,
+    for use in escape log messages.
+    """
+    from .tests._escape_triage import locate_escape_site
+
+    # Drop this function's own frame before searching.
+    frames = traceback.extract_stack()[:-1]
+    site = locate_escape_site(frames)
+    if site is None:
+        return "<unknown location>"
+    return f"{site.filename}:{site.lineno} {site.name}"
+
+
+class _ReentrancyGuard:
+    """Small helper to keep our wrappers from recursing into themselves."""
+
+    def __init__(self):
+        self.active = False
+
+
+def _make_escape_logging_wrapper(original, funcname, guard):
+    def wrapper(*args, **kwargs):
+        if not guard.active and args:
+            guard.active = True
+            try:
+                obj = args[0]
+                if _is_foreign_array(obj):
+                    site = _describe_escape_site()
+                    _escape_logger.warning(
+                        "array-API escape: %s() called on a %r array "
+                        "(namespace=%r) at %s",
+                        funcname,
+                        type(obj),
+                        obj.__array_namespace__(),
+                        site,
+                    )
+            finally:
+                guard.active = False
+        return original(*args, **kwargs)
+
+    wrapper.__name__ = getattr(original, "__name__", funcname)
+    wrapper.__doc__ = getattr(original, "__doc__", None)
+    return wrapper
+
+
+@pytest.fixture(autouse=True, scope="session")
+def _log_array_escapes():
+    """
+    Session-scoped autouse fixture that, when CCDPROC_LOG_ARRAY_ESCAPES is
+    set, monkeypatches numpy's array-coercion entry points for the duration
+    of the test session so that any silent conversion of a non-numpy
+    array-API array is logged instead of passing unnoticed.
+    """
+    if not _LOG_ARRAY_ESCAPES:
+        yield
+        return
+
+    guard = _ReentrancyGuard()
+
+    originals = {
+        "np.asarray": np.asarray,
+        "np.asanyarray": np.asanyarray,
+        "np.ma.asanyarray": np.ma.asanyarray,
+    }
+
+    np.asarray = _make_escape_logging_wrapper(
+        originals["np.asarray"], "numpy.asarray", guard
+    )
+    np.asanyarray = _make_escape_logging_wrapper(
+        originals["np.asanyarray"], "numpy.asanyarray", guard
+    )
+    np.ma.asanyarray = _make_escape_logging_wrapper(
+        originals["np.ma.asanyarray"], "numpy.ma.asanyarray", guard
+    )
+
+    try:
+        yield
+    finally:
+        np.asarray = originals["np.asarray"]
+        np.asanyarray = originals["np.asanyarray"]
+        np.ma.asanyarray = originals["np.ma.asanyarray"]
--- a/ccdproc/conftest.py
+++ b/ccdproc/conftest.py
@@ -159,6 +159,12 @@ def pytest_collection_modifyitems(items):
 # an array whose array-API namespace (as resolved by
 # array_api_compat.array_namespace) is not numpy.
 #
+# Only Python-level calls through the module attributes are visible:
+# C-level coercions inside compiled dependencies (scipy, astroscrappy,
+# reproject) and references bound before the patch ("from numpy import
+# asarray") bypass the wrappers entirely, so absence of a warning is not
+# proof there was no conversion.
+#
 # Activated by setting the environment variable CCDPROC_LOG_ARRAY_ESCAPES to
 # a truthy value.
 # ---------------------------------------------------------------------------
--- a/ccdproc/conftest.py
+++ b/ccdproc/conftest.py
@@ -129,6 +129,17 @@ _ACTIVE_BACKEND = _normalize_backend_name(array_library)
 
 
 def pytest_collection_modifyitems(items):
+    """
+    Apply the ``backend_skip`` / ``backend_xfail`` markers at collection
+    time. For each collected test, the backends named in a marker's
+    positional args (normalized, so ``array_api_strict`` and
+    ``array-api-strict`` match) are compared against the active backend from
+    ``CCDPROC_ARRAY_LIBRARY``; on a match the corresponding standard pytest
+    marker is attached -- a plain skip, or a *non-strict* xfail so a test
+    that starts passing XPASSes instead of failing the run (see the "prune
+    on XPASS" note in docs/array_api.rst). Tests without a matching marker,
+    including everything in the default numpy run, are untouched.
+    """
     for item in items:
         for marker in item.iter_markers(name="backend_skip"):
             backends = {_normalize_backend_name(b) for b in marker.args}
@@ -137,6 +148,7 @@ def pytest_collection_modifyitems(items):
                     "reason", f"skipped for array backend {_ACTIVE_BACKEND!r}"
                 )
                 item.add_marker(pytest.mark.skip(reason=reason))
+                # One matching marker decides the outcome; skip the rest.
                 break
 
         for marker in item.iter_markers(name="backend_xfail"):
@@ -235,13 +247,31 @@ class _ReentrancyGuard:
 
 
 def _make_escape_logging_wrapper(original, funcname, guard):
+    """
+    Build the replacement for one numpy coercion entry point (``original`` is
+    the real ``np.asarray``, ``np.asanyarray`` or ``np.ma.asanyarray``).
+    The wrapper logs and tallies an "array-API escape" whenever its first
+    positional argument is an array from a non-numpy array-API library, then
+    always finishes by calling ``original`` -- behavior is unchanged, escapes
+    are only observed. ``funcname`` is the dotted numpy name (e.g.
+    ``numpy.asarray``) recorded with each escape. ``guard`` breaks recursion:
+    the namespace detection and stack extraction below can themselves end up
+    calling the patched functions.
+    """
+
     def wrapper(*args, **kwargs):
+        # Do nothing extra on re-entrant calls (guard held) or when there is
+        # no positional argument to inspect.
         if not guard.active and args:
             guard.active = True
             try:
                 obj = args[0]
                 namespace = _foreign_namespace(obj)
                 if namespace is not None:
+                    # A foreign array is about to be coerced to numpy: warn
+                    # with the innermost non-test ccdproc frame to blame, and
+                    # tally the site for the end-of-session summary and the
+                    # baseline ratchet.
                     frame = _escape_site_frame()
                     site = _describe_escape_site(frame)
                     _escape_logger.warning(
--- a/ccdproc/conftest.py
+++ b/ccdproc/conftest.py
@@ -25,6 +25,7 @@ except ImportError:
 
 
 from .tests._escape_triage import (
+    _env_truthy,
     pytest_runtest_makereport,  # noqa: F401 pytest hook, used via attribute lookup
     pytest_terminal_summary,  # noqa: F401 pytest hook, used via attribute lookup
 )
@@ -52,10 +53,12 @@ array_library = os.environ.get("CCDPROC_ARRAY_LIBRARY", "numpy").lower()
 
 # Device to create test arrays on. This is only meaningful for backends that
 # support multiple devices (currently array-api-strict, as a CPU-only proxy
-# for testing non-default-device behavior like CuPy's GPU device). It can be
-# overridden with the CCDPROC_ARRAY_DEVICE environment variable, whose value
-# is passed to the array library's Device constructor. "default" (or leaving
-# it unset) means the library's usual default device.
+# for testing non-default-device behavior like CuPy's GPU device). Leaving
+# CCDPROC_ARRAY_DEVICE unset selects the backend's testing default: the
+# non-default "device1" for array-api-strict, None (the library's usual
+# device) for everything else. Setting it to "default" selects the library's
+# normal default device; any other value is passed to the library's Device
+# constructor.
 testing_array_device = None
 
 match array_library:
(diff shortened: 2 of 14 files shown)


CHANGES.rst
```diff
@@ -1,3 +1,20 @@
+Unreleased
+----------
+
+New Features
+^^^^^^^^^^^^
+
+- Added support for testing against ``array-api-strict`` on a non-default
```
mwcraig: These entries are too long, make then briefer.

ccdproc/conftest.py
```diff
@@ -60,8 +77,197 @@ def pytest_configure(config):
...
+_LOG_ARRAY_ESCAPES = _env_truthy(os.environ.get("CCDPROC_LOG_ARRAY_ESCAPES", ""))
+
+# array_api_compat.numpy wraps plain numpy; arrays built through it report
+# __array_namespace__() as one of these two modules. Neither counts as an
+# "escape" -- we only care about arrays from a genuinely different library
+# (jax, dask, cupy, array_api_strict, ...) ending up in a numpy-only call.
+_NUMPY_LIKE_NAMESPACES = {np, array_api_compat.numpy}
+
+
+def _is_foreign_array(obj):
```
mwcraig: Add comments explaining this.

ccdproc/tests/_escape_triage.py
```diff
@@ -0,0 +1,131 @@
...
+import os
+import traceback
+from collections import defaultdict
+
+import pytest
+
+_TRUTHY = {"1", "true", "yes", "on"}
+
+
+def _env_truthy(value):
```
mwcraig, line 38: Isn't this in another place in this PR?

ccdproc/tests/_escape_triage.py
```diff
@@ -0,0 +1,131 @@
...
+
+    ccdproc_frames = [f for f in frames if _is_ccdproc_frame(f.filename)]
+    if ccdproc_frames:
+        return ccdproc_frames[-1]
+
+    return frames[-1]
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item, call):
```
mwcraig, line 285: Add comments to this and the next

ccdproc/conftest.py
```diff
@@ -60,8 +77,197 @@ def pytest_configure(config):
...
+        # here makes np.asarray() raise on the resulting arrays, the same way
+        # it would for a CuPy array living on a GPU. That makes
+        # array-api-strict a convenient CPU-only proxy for catching the
+        # "silent conversion to numpy" bugs that would otherwise only show up
+        # on CuPy.
+        device_name = os.environ.get("CCDPROC_ARRAY_DEVICE", "default").lower()
+        if device_name != "default":
+            testing_array_device = testing_array_library.Device(device_name)
+        else:
+            testing_array_device = testing_array_library.Device("device1")
```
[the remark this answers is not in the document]
mwcraig: Confirmed — this is a real mismatch, and the docs are the authority on intent here. `docs/array_api.rst` says "set it to ``default`` to use the library's normal CPU device instead", but this code maps `default` (and unset) to `device1`, so the documented keyword is broken. (Verified that `array_api_strict.Device("CPU_DEVICE")` does work as a literal value, so the capability exists — it's the `default` keyword that's wired backwards.)

Planned fix, keeping unset = `device1` since forcing the non-default device is the whole point of this backend:

```python
device_name = os.environ.get("CCDPROC_ARRAY_DEVICE", "device1")
if device_name.lower() == "default":
    # The library's normal CPU device, on which np.asarray() succeeds.
    testing_array_device = testing_array_library.Device("CPU_DEVICE")
else:
    testing_array_device = testing_array_library.Device(device_name)
```

ccdproc/conftest.py
```diff
@@ -41,6 +50,14 @@ def pytest_configure(config):
 # What happens here is controlled by an environmental variable
 array_library = os.environ.get("CCDPROC_ARRAY_LIBRARY", "numpy").lower()
 
+# Device to create test arrays on. This is only meaningful for backends that
+# support multiple devices (currently array-api-strict, as a CPU-only proxy
+# for testing non-default-device behavior like CuPy's GPU device). It can be
+# overridden with the CCDPROC_ARRAY_DEVICE environment variable, whose value
+# is passed to the array library's Device constructor. "default" (or leaving
+# it unset) means the library's usual default device.
```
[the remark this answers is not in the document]
mwcraig: Agreed — same root cause as the thread on line 96. This comment describes the intended behavior ("default"/unset → library default) but the code does the opposite, and the docs promise yet another combination (unset → device1, `default` → CPU device). The docs' version is the right one. The fix will make the code match the docs and reword this comment to: unset → the backend's non-default device (`device1` for array-api-strict, `None` for everything else); `CCDPROC_ARRAY_DEVICE=default` → the library's normal default device; any other value → passed to the library's `Device` constructor.

tox.ini
```diff
@@ -55,6 +57,7 @@ deps =
     oldestdeps: reproject==0.9.1
 
     dask: dask
+    strict: array-api-strict
```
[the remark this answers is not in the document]
mwcraig: Valid — the backend depends on `array_api_strict.Device` and multi-device simulation, which landed in array-api-strict 2.0 (the design notes in #937 already assume ≥ 2.0; local runs verified against 2.3.1). Will change the tox dep to `strict: array-api-strict>=2.0`.

CHANGES.rst
mwcraig: Will shorten to:

```rst
- Add ``array-api-strict`` (on a non-default device) as a test backend, a
  CPU-only proxy for CuPy device behavior. [#937]
- Add triage tooling for array-API backend failures: ``backend_xfail`` /
  ``backend_skip`` markers, an escape-site report
  (``CCDPROC_TRIAGE_ESCAPES=1``), and an escape logger
  (``CCDPROC_LOG_ARRAY_ESCAPES=1``). [#939]
```

The full usage details stay in ``docs/array_api.rst``.

ccdproc/conftest.py
mwcraig: Will add an explanation along these lines to `_is_foreign_array`:

```python
def _is_foreign_array(obj):
    """
    Return True if obj is an array from a non-numpy array-API library.

    numpy arrays (including np.ma masked arrays) are never foreign. Anything
    else that implements the array-API protocol marker __array_namespace__()
    is foreign unless its namespace is numpy or array_api_compat's numpy
    wrapper -- i.e. exactly the arrays whose conversion to numpy would fail
    (CuPy on GPU) or silently densify/transfer (dask, jax). The try/except
    guards against objects whose __array_namespace__() raises (e.g. on an
    unsupported api version argument); those are treated as not foreign so
    the logger never breaks the call it wraps.
    """
```

ccdproc/tests/_escape_triage.py
mwcraig: Yes — `_env_truthy` is duplicated in `ccdproc/conftest.py` (line 105). Since conftest already imports the triage hooks from this module, the cleanest consolidation is to keep the definition here and have conftest import it: `from .tests._escape_triage import _env_truthy`. Will do that.

ccdproc/tests/_escape_triage.py
mwcraig: Will add docstrings to both hooks:

```python
@pytest.hookimpl(hookwrapper=True)
def pytest_runtest_makereport(item, call):
    """
    Record the escape site of every failing test.

    Runs as a wrapper around report generation for each test phase; only the
    'call' phase (the test body itself, not setup/teardown) of failing tests
    is recorded. The failure's traceback is reduced to its most informative
    frame by locate_escape_site() and the test id is filed under that
    (file, line, function) key for the end-of-session summary.
    """

def pytest_terminal_summary(terminalreporter):
    """
    Print the escape-site summary at the end of the test session.

    One section listing each escape site with its failure count (most
    common first) and up to five example test ids, so a large batch of
    backend failures collapses to a short list of root-cause call sites.
    """
```

mwcraig:
## Review

**Overview**: This PR adds an `array-api-strict` test backend (with fixture arrays on a non-default device so `np.asarray` raises, emulating CuPy device behavior), per-backend `backend_xfail`/`backend_skip` markers, an escape-site triage report (`CCDPROC_TRIAGE_ESCAPES`), and an escape logger (`CCDPROC_LOG_ARRAY_ESCAPES`), plus a `py313-strict` tox env and CI job.

I verified the happy paths locally: the default numpy backend still passes 331/331 (5 skipped), the strict backend applies the markers correctly, and the triage summary groups failures by sensible sites. However, the escape *logger* has three real bugs, one of which makes its main output wrong in every case.

### Bug 1 (high): the escape logger always reports its own wrapper as the escape site

`_describe_escape_site()` in `ccdproc/conftest.py` drops only its own frame (`extract_stack()[:-1]`), but the caller is the monkeypatch `wrapper` defined in `ccdproc/conftest.py` — which is itself inside the package root and *not* under `ccdproc/tests/`, so `locate_escape_site()` always selects it as the "innermost non-test ccdproc frame". Every escape message points at the wrapper, never at the real call site:

```
array-API escape: numpy.asarray() called on a <class 'array_api_strict._array_object.Array'> array
(namespace=...) at /Users/.../ccdproc/conftest.py:231 wrapper
```

(Reproduced by wrapping `np.asarray` via `_make_escape_logging_wrapper` and triggering an escape from a frame compiled with a `ccdproc/core.py` filename — the report names `conftest.py:231 wrapper` instead of `core.py`.)

Since the call site is the entire point of the logger, this defeats the feature. Dropping two frames (`[:-2]`) fixes the immediate symptom, but a more robust fix covers Bug 3 as well: have the frame classifiers treat `ccdproc/conftest.py` as test infrastructure (i.e., exclude it alongside `ccdproc/tests/`), so it can never win "innermost non-test ccdproc frame".

### Bug 2 (medium): dask escapes are never detected, contradicting the stated purpose

The PR body and the comments in `conftest.py` say the logger "catches backends like dask and JAX that convert silently". But `_is_foreign_array()` keys on `obj.__array_namespace__`, and dask arrays don't have that attribute — even via `array_api_compat.dask.array` (verified with dask 2025.7.0 / array-api-compat 1.12.0: `hasattr(a, "__array_namespace__")` is `False`, while `array_api_compat.array_namespace(a)` correctly returns the compat dask namespace). So for the dask backend the logger silently reports nothing — the exact failure mode it was written to prevent.

Suggested fix: detect via `array_api_compat.array_namespace(obj)` in a `try/except TypeError` instead of the raw dunder (the `isinstance(obj, np.ndarray)` short-circuit already handles plain numpy, and the existing `_NUMPY_LIKE_NAMESPACES` exclusion still applies).

### Bug 3 (medium): enabling both env vars corrupts the triage report

When `CCDPROC_LOG_ARRAY_ESCAPES=1` and `CCDPROC_TRIAGE_ESCAPES=1` are set together, any failure whose exception is raised *inside* the wrapped `np.asarray` carries the wrapper's frame as the innermost non-test ccdproc frame, so the triage summary attributes it to `conftest.py … wrapper` rather than the real site. Reproduced on this branch:

```
$ CCDPROC_ARRAY_LIBRARY=array-api-strict CCDPROC_TRIAGE_ESCAPES=1 CCDPROC_LOG_ARRAY_ESCAPES=1 \
    pytest ccdproc/tests/test_ccdproc.py ccdproc/tests/test_combiner.py
...
/Users/.../ccdproc/conftest.py:242 wrapper  (1 failures)
```

These are exactly the failures the dedupe report is supposed to localize (the ones that would raise on CuPy), so grouping them under `wrapper` hides the root cause. Excluding `conftest.py` in the frame classifiers (the Bug 1 fix) resolves this too.

### Bug 4 (medium, usability): logger output is invisible in the scenario it targets

`_escape_logger.warning(...)` goes through `logging`, and pytest only displays captured log records for *failing* tests. The logger's stated purpose is catching **silent** escapes on dask/JAX — where the test **passes** — and in that case the warning is swallowed entirely (verified: a passing test emitting `logging.getLogger("ccdproc.array_escape").warning(...)` produces no visible output under this repo's pytest config). As shipped, `CCDPROC_LOG_ARRAY_ESCAPES=1` on the dask/jax tox envs would print nothing even if Bug 2 were fixed.

Options: emit `warnings.warn(...)` instead (shows up in pytest's warnings summary for passing tests), or document that the flag requires `--log-cli-level=WARNING` (and consider setting that in the relevant tox envs).

### Minor notes

- `docs/array_api.rst` describes array-api-strict as "performs no computation itself" — it's a thin wrapper over numpy and does compute; its distinguishing feature is strict API/device validation. Suggest rewording.
- `_escape_triage.pytest_runtest_makereport` only records the `call` phase, so escapes that fail during fixture setup won't be triaged. Probably fine for this codebase (fixtures are mostly plain functions here), but worth a comment.
- In the wrapper, `obj.__array_namespace__()` is invoked a second time just to build the log message; reusing the namespace from `_is_foreign_array` (e.g. returning it) would avoid re-calling arbitrary third-party code outside the guarded check.
- The new CI job references `${{ matrix.toxposargs }}` which is never defined in its matrix (harmless — evaluates to empty — and consistent with the main job, where some entries define it).
- The `ci-tests-expected-failures` job will show a red ✗ on every PR until the ~109 strict failures are fixed. The comment says that's deliberate; just flagging that reviewers of unrelated PRs will see a failing check.

### What checks out

- Frame-classification anchoring on the package directory (`fdf651e`) correctly avoids the CI-checkout-named-`ccdproc` substring trap; verified sites resolve to real `ccdproc/*.py` locations in a strict-backend run.
- Marker plumbing (`pytest_collection_modifyitems`, non-strict xfail, marker registration in `pyproject.toml`) works as advertised; `py313-strict` naming matches the existing `py313-jax` / `py312-alldeps-dask` pattern, and `CCDPROC_ARRAY_DEVICE=default` / `device1` handling matches `array_api_strict.Device`'s accepted names.
- Hook re-export through `conftest.py` (`pytest_runtest_makereport`, `pytest_terminal_summary`) is picked up by pytest, and the hookwrapper yields exactly once on all paths.
- Default numpy run unaffected: 331 passed, 5 skipped, matching main.

Bugs 1–3 share one root cause (frame/namespace detection in the logger) and are small fixes; the triage report and markers — the parts the strict CI job actually exercises — are solid.

https://claude.ai/code/session_01Qri8rDgkWjCT7Yb7QKJrhf

ccdproc/tests/_escape_triage.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    Record the escape site of every failing test.
+
+    Runs as a wrapper around report generation for each test phase; only the
+    'call' phase (the test body itself, not setup/teardown) of failing tests
+    is recorded. The failure's traceback is reduced to its most informative
+    frame by locate_escape_site() and the test id is filed under that
+    (file, line, function) key for the end-of-session summary.
+    """
```
mwcraig, line 297: The pytest documentation says a hook wrapper needs to return a result or raise an exception....

ccdproc/tests/_escape_triage.py
```diff
@@ -0,0 +1,155 @@
...
+        return non_test_ccdproc[-1]
+
+    ccdproc_frames = [f for f in frames if _is_ccdproc_frame(f.filename)]
+    if ccdproc_frames:
+        return ccdproc_frames[-1]
+
+    return frames[-1]
+
+
+@pytest.hookimpl(hookwrapper=True)
```
mwcraig: The pluggy documentation describe this as an "old-style" wrapper; new style are to be preferred, so why not do that?

ccdproc/tests/_escape_triage.py
```diff
@@ -0,0 +1,155 @@
...
+
+    site = locate_escape_site(traceback.extract_tb(excinfo.tb))
+    if site is None:
+        return
+
+    key = (site.filename, site.lineno, site.name)
+    _ESCAPE_SITES[key].append(item.nodeid)
+
+
+def pytest_terminal_summary(terminalreporter):
```
mwcraig, line 316: Point me to where the pytest documentation on this are.

ccdproc/conftest.py
```diff
@@ -19,6 +24,11 @@ def pytest_configure(config):
     TESTED_VERSIONS = {}
 
 
+from .tests._escape_triage import (
```
mwcraig, line 27: Why are these functions in tests instead of just being in conftest?

ccdproc/conftest.py
```diff
@@ -60,8 +80,205 @@ def pytest_configure(config):
...
+
+
+def _normalize_backend_name(name):
+    return name.lower().replace("_", "-")
+
+
+_ACTIVE_BACKEND = _normalize_backend_name(array_library)
+
+
+def pytest_collection_modifyitems(items):
```
mwcraig, line 131: point me to the pytest documentation on this.

ccdproc/conftest.py
```diff
@@ -60,8 +80,205 @@ def pytest_configure(config):
...
+                if _is_foreign_array(obj):
+                    site = _describe_escape_site()
+                    _escape_logger.warning(
+                        "array-API escape: %s() called on a %r array "
+                        "(namespace=%r) at %s",
+                        funcname,
+                        type(obj),
+                        obj.__array_namespace__(),
+                        site,
+                    )
```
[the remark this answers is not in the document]
mwcraig: Fair point, though the practical risk is nearly nil: this line only runs after `_is_foreign_array(obj)` has just called `obj.__array_namespace__()` successfully, so raising here would require a non-idempotent `__array_namespace__` — no real backend does that. Still, calling it twice is the underlying smell, and removing the double call closes the theoretical hole for free. Planned fix: turn the predicate into a helper that returns the namespace (or `None`), and reuse that value in the log message:

```python
def _foreign_namespace(obj):
    """Return obj's array-API namespace if it is foreign, else None."""
    if isinstance(obj, np.ndarray):
        return None
    get_namespace = getattr(obj, "__array_namespace__", None)
    if get_namespace is None:
        return None
    try:
        namespace = get_namespace()
    except Exception:
        return None
    return None if namespace in _NUMPY_LIKE_NAMESPACES else namespace
```

```python
namespace = _foreign_namespace(obj)
if namespace is not None:
    site = _describe_escape_site()
    _escape_logger.warning(
        "array-API escape: %s() called on a %r array (namespace=%r) at %s",
        funcname, type(obj), namespace, site,
    )
```

ccdproc/tests/_escape_triage.py
mwcraig: That requirement applies to *new-style* wrappers (`@pytest.hookimpl(wrapper=True)`), not to this one. This hook is an old-style wrapper (`hookwrapper=True`), and for those the rule is the opposite: the pluggy docs state old-style wrappers "can **not** return results; they can only modify them using the `force_result()` API" — the generator's own return value is ignored (https://pluggy.readthedocs.io/en/stable/#wrappers, "Old-style wrappers"). Here `yield` hands back a `pluggy.Result`, we read it with `outcome.get_result()`, and the early `return`s just end the generator without touching the hook result — which is exactly what we want, since this hook only observes.

That said, the sibling thread asking to convert to a new-style wrapper makes this moot: after that conversion the "return a result or raise" rule *does* apply, and the rewrite ends with `return report` accordingly (see the code in that thread).

ccdproc/tests/_escape_triage.py
mwcraig: No good reason not to — agreed, will convert. New-style wrappers need pluggy ≥ 1.1 (June 2023); this repo doesn't pin pytest or pluggy anywhere (not even in `oldestdeps`, which lets them float), and the tox envs currently resolve to pytest 8.4.1 / pluggy 1.6.0, so there's no compatibility cost. Planned rewrite:

```python
@pytest.hookimpl(wrapper=True)
def pytest_runtest_makereport(item, call):
    # With a new-style wrapper, yield hands back the TestReport itself
    # (not a pluggy.Result), and the wrapper must return it.
    report = yield

    if (
        TRIAGE_ACTIVE
        and report.when == "call"
        and report.failed
        and call.excinfo is not None
    ):
        site = locate_escape_site(traceback.extract_tb(call.excinfo.tb))
        if site is not None:
            key = (site.filename, site.lineno, site.name)
            _ESCAPE_SITES[key].append(item.nodeid)

    return report
```

This also settles the "needs to return a result" question in the other thread — that rule is exactly the new-style contract, and `return report` satisfies it.

ccdproc/tests/_escape_triage.py
mwcraig: Two relevant pieces of documentation:

- The hook spec for `pytest_terminal_summary(terminalreporter, exitstatus, config)`: https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_terminal_summary — called "after command line options have been parsed" to add a section to the terminal summary reporting; unused parameters can be omitted from the signature, which is why this implementation only takes `terminalreporter`.
- The `TerminalReporter` API that provides the `section()` and `write_line()` methods used here: https://docs.pytest.org/en/stable/reference/reference.html#pytest.TerminalReporter

ccdproc/conftest.py
mwcraig: Three reasons, in decreasing order of weight:

1. **`locate_escape_site` is shared.** The pytest triage hooks use it on failure tracebacks, and the runtime escape logger in `conftest.py` (`_describe_escape_site`) uses it on live stacks. Keeping it in a plain module lets both import it; importing things *from* a conftest (`from ccdproc.conftest import ...`) works but is unconventional and fragile — conftest files are meant to be discovered by pytest, not used as a library.
2. **Testability.** `locate_escape_site` and the frame-classification helpers are pure functions that can be unit-tested directly from a regular module.
3. **Size.** `conftest.py` is already ~285 lines with the backend selection, marker handling, and escape logger; this keeps a self-contained feature self-contained.

pytest still picks up the hooks because conftest imports them into its namespace — that's what this import block is for (hook discovery is by attribute lookup on the conftest module).

The one thing that genuinely *must* stay in `ccdproc/tests/` is the path anchoring: `_TESTS_ROOT`/`_PACKAGE_ROOT` are derived from the module's own `__file__`, though that could be spelled explicitly if the code moved. Happy to inline it all into conftest if you'd rather have one file — the sharing in point 1 is the main thing that would get uglier.

ccdproc/conftest.py
mwcraig: - Hook spec for `pytest_collection_modifyitems(session, config, items)`: https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems — "called after collection has been performed"; may filter or re-order the items in-place. As with other hooks, unused parameters may be dropped from the signature, hence `def pytest_collection_modifyitems(items)`.
- A worked example of exactly this pattern (loop over `items`, conditionally `item.add_marker(pytest.mark.skip(...))`): https://docs.pytest.org/en/stable/example/simple.html#control-skipping-of-tests-according-to-command-line-option
- `Node.iter_markers()`, used here to read the `backend_xfail`/`backend_skip` markers: https://docs.pytest.org/en/stable/reference/reference.html#pytest.nodes.Node.iter_markers

mwcraig:
Re-verified all four escape-logger bugs independently on `fdf651e` — every one is real:

- **Bug 1 (wrapper reported as escape site)**: reproduced directly. Wrapping `np.asarray` via `_make_escape_logging_wrapper` and calling it from a plain function reports `ccdproc/conftest.py:231 wrapper` as the site, exactly as described. The root cause is as stated: `extract_stack()[:-1]` drops only `_describe_escape_site` itself, leaving the wrapper as the innermost frame, and the wrapper lives in `ccdproc/conftest.py`, which the classifiers count as a non-test ccdproc frame.
- **Bug 2 (dask never detected)**: confirmed with the `py312-alldeps-dask` tox env (dask 2025.7.0, array-api-compat 1.12.0): `hasattr(a, "__array_namespace__")` is `False` for a dask array built through `array_api_compat.dask.array`, while `array_api_compat.array_namespace(a)` correctly resolves it. The suggested `array_namespace()`-based detection is the right fix.
- **Bug 3 (triage report corruption when both flags are set)**: same root cause as Bug 1; excluding `conftest.py` in the frame classifiers fixes both.
- **Bug 4 (logger invisible on passing tests)**: confirmed under this repo's actual pytest config — a passing test emitting the warning produces zero visible output.

Two refinements to the proposed fixes:

1. **Bugs 1/3**: agree that classifying `ccdproc/conftest.py` as test infrastructure (rather than `extract_stack()[:-2]`) is the way to go — it fixes both bugs at once and doesn't break if a frame is ever added or removed between the wrapper and the site helper. Concretely, `_escape_triage.py` will treat frames from `os.path.join(_PACKAGE_ROOT, "conftest.py")` the same as `ccdproc/tests/` frames.
2. **Bug 4**: of the two options given, `warnings.warn(...)` is **not** viable here — `pyproject.toml` sets `filterwarnings = ["error", ...]`, so every escape would become a hard test failure on backends where the test otherwise passes, turning a logging-only flag into a behavior change. The other option works with almost no code: `log_cli_level = "INFO"` is already set in `pyproject.toml`, and verified that adding `-o log_cli=true` makes the warning visible on passing tests. Plan: enable `log_cli` in the tox envs (or document the flag) wherever `CCDPROC_LOG_ARRAY_ESCAPES` is set.

One small extra found while reproducing Bug 1: `_is_foreign_array`'s final `namespace not in _NUMPY_LIKE_NAMESPACES` test sits *outside* the try/except and requires the namespace to be hashable. Real namespaces are modules (hashable), so it's theoretical, but the spec only promises "an object" — the planned `_foreign_namespace()` refactor (also requested in Copilot's inline comment about the double `__array_namespace__()` call) will move that test inside the guarded helper.

The minor notes all look right too; the `docs/array_api.rst` "performs no computation" wording and the missing comment on `call`-phase-only triage will be fixed in the same pass. The red ✗ on `ci-tests-expected-failures` remains deliberate until the ~109 strict failures are burned down.

ccdproc/conftest.py
mwcraig: Keep it where it is.

ccdproc/conftest.py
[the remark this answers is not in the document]
mwcraig: Fixed in a9313a6 (pushed). `_is_foreign_array` is now `_foreign_namespace()`, which returns the namespace (or `None`), and the log message reuses that returned value — so the namespace machinery runs exactly once, inside the guarded helper. Two refinements over the sketch above, per the follow-up review comment: detection goes through `array_api_compat.array_namespace()` rather than the raw dunder (so dask arrays, which lack `__array_namespace__`, are now caught too), and the `_NUMPY_LIKE_NAMESPACES` membership test moved inside the try/except so an unhashable namespace can't raise either.

ccdproc/tests/_escape_triage.py
mwcraig: Settled by the new-style conversion in a9313a6: the hook is now `@pytest.hookimpl(wrapper=True)` and ends with `return report`, so the "return a result or raise" rule from the pytest docs — which is exactly the new-style contract — is satisfied.

ccdproc/tests/_escape_triage.py
mwcraig: Converted in a9313a6 (pushed), as proposed above: `@pytest.hookimpl(wrapper=True)`, `report = yield` (the `TestReport` itself, not a `pluggy.Result`), the early-returns folded into a single `if` so the wrapper always returns, and a final `return report`. Verified with a strict-backend triage run — the escape-site summary is unchanged from the old-style wrapper.

ccdproc/conftest.py
mwcraig: Kept in `ccdproc/tests/` as directed — no move in a9313a6.

mwcraig:
All four escape-logger bugs are fixed in a9313a6 (pushed), along with the agreed minor notes. Each fix was verified end-to-end before committing:

- **Bug 1** (wrapper reported as escape site): `ccdproc/conftest.py` frames are now classified as test infrastructure in `_escape_triage.py`, so the monkeypatch wrapper can never win "innermost non-test ccdproc frame". A synthetic escape triggered from a frame compiled with a `ccdproc/core.py` filename now logs that frame, not `conftest.py … wrapper`.
- **Bug 2** (dask never detected): detection now goes through `array_api_compat.array_namespace()` via the new `_foreign_namespace()` helper. Both plain `dask.array` arrays and compat-built ones resolve to the compat dask namespace; numpy arrays, masked arrays, and non-arrays return `None`.
- **Bug 3** (triage corruption with both flags): same classifier fix. A strict-backend run of `test_ccdproc.py` + `test_combiner.py` with `CCDPROC_TRIAGE_ESCAPES=1 CCDPROC_LOG_ARRAY_ESCAPES=1` now produces a triage summary with zero `conftest.py` entries, byte-identical to a logger-off run.
- **Bug 4** (logger invisible on passing tests): documented in `docs/array_api.rst` — the triage flags and backend markers now have a section there, including the `-o log_cli=true` requirement (`warnings.warn` stays ruled out because of `filterwarnings = ["error"]`). Verified on the dask backend: a fully passing `test_combiner.py` run (74 passed) with the flag and `-o log_cli=true` surfaced 125 escape warnings pointing at real sites (e.g. `combiner.py:680 average_combine`, `core.py:1233 sigma_func`).

Minor notes also addressed: the array-api-strict description in `docs/array_api.rst` no longer claims it "performs no computation", and the triage hook's docstring now states that fixture-setup failures are deliberately not recorded. The default numpy run is unchanged (331 passed, 5 skipped) and ruff/pre-commit pass. Thread-level replies posted alongside this comment; the open threads should be resolvable now.

mwcraig (commented):
## Critical review

### Summary

The mechanism design is solid in the places people usually get wrong — `outcome` vs `conclusion` with `continue-on-error`, cross-workflow artifact download (`run-id` + `github-token`), the `workflows: ["CI"]` name match, `checks: write`/`actions: read` permissions, `__file__`-anchored baseline path surviving tox `changedir`, line-number-free baseline keys, and a missing-baseline-file that fails loud rather than silent. Those are all correct.

The real problems are in the tooling's failure modes and the CI plumbing's edge cases: **the baseline regeneration mode can silently destroy the baseline file**, **enforcement silently stops enforcing if one env var is missing**, and the strict-status pipeline **conflates infrastructure failure with "expected test failure" (or reports nothing at all)**. Plus a changelog pointing at PRs that don't exist, a stale jax xfail that never fires, and the ratchet workflow — the heart of the PR — is undocumented in `docs/array_api.rst`.

All reproductions below were run against the PR head (`b564834`) in a fresh py3.12 venv (jax 0.10.2, dask 2026.7.1).

---

### Bugs (confirmed, with reproductions)

**B1. `CCDPROC_WRITE_ESCAPE_BASELINE=1` silently wipes or truncates the baseline** — `ccdproc/tests/_escape_triage.py:412-414`, `170-206`

`_write_escape_baseline()` rewrites the file from `_observed_library_sites()`, which reads `_ESCAPE_LOG_COUNTS` — populated *only* by the `_log_array_escapes` fixture in `conftest.py`, which returns immediately unless `CCDPROC_LOG_ARRAY_ESCAPES` is truthy (`conftest.py:266`). Write mode neither enables the logger nor checks that it ran. The file header says only "Regenerate with CCDPROC_WRITE_ESCAPE_BASELINE=1 (preserves tags)."

Two failure modes:
- Follow the header's instruction literally (no `CCDPROC_LOG_ARRAY_ESCAPES`, or on the default numpy backend): observed set is empty → the entire baseline is replaced with the header plus `# (no library escapes observed this run)` (`_escape_triage.py:192-194`).
- Even with everything set correctly, running a *subset* of tests drops every baseline entry for code the subset didn't exercise — "preserves tags" preserves tags only for sites still observed.

Verified live:
```
CCDPROC_WRITE_ESCAPE_BASELINE=1 python -m pytest ccdproc/tests/test_ccdproc.py::test_create_deviation -q
# → 7 passed in 0.95s; array_escape_baseline.txt reduced from 13 entries to
#   "# (no library escapes observed this run)"  (git diff: 1 insertion, 13 deletions)
```
Suggested fix: refuse to write when the logger is inactive (`if not _LOG_ARRAY_ESCAPES: raise UsageError(...)`) and warn (or refuse) when the new site set is smaller than the existing baseline / empty.

**B2. Enforcement silently no-ops without `CCDPROC_LOG_ARRAY_ESCAPES`** — `_escape_triage.py:415-416`, `371-377`

`CCDPROC_ENFORCE_ESCAPE_BASELINE` and `CCDPROC_LOG_ARRAY_ESCAPES` are coupled only in the tox `enforce` factor, not in code. Set enforce without the logger and `_observed_library_sites()` is empty → `_new_escapes()` is empty → session passes, and the terminal summary prints the reassuring "No foreign-array escapes observed (numpy backend?); baseline not checked." A CI config that copies only the enforce var gets a permanently-green, non-enforcing gate.

Verified live (deleted the `combiner.py average_combine` baseline line to simulate a new escape):
```
CCDPROC_ENFORCE_ESCAPE_BASELINE=1 CCDPROC_ARRAY_LIBRARY=dask python -m pytest ccdproc/tests/test_combiner.py -q
# → 74 passed, exit 0: "No foreign-array escapes observed (numpy backend?); baseline not checked."
CCDPROC_LOG_ARRAY_ESCAPES=1 CCDPROC_ENFORCE_ESCAPE_BASELINE=1 CCDPROC_ARRAY_LIBRARY=dask python -m pytest ccdproc/tests/test_combiner.py -q
# → exit 1: "NEW escapes not in baseline (1) -- these fail the session: + combiner.py  average_combine  numpy.asarray"
```
Note the message even says "numpy backend?" while running on dask — the diagnostic itself misleads. Suggested fix: in `pytest_sessionfinish` (or at import), error out if `ENFORCE_BASELINE and not _LOG_ARRAY_ESCAPES`.

**B3. Changelog cites nonexistent PRs** — `CHANGES.rst:8,12`

The two new entries are tagged `[#937]` and `[#939]`; neither PR exists. Both should be `[#942]`. Also, the new section heading is bare `Unreleased`, where this repo's convention is `x.y.z (unreleased)`.

**B4. Hardcoded artifact name breaks the moment the expected-failures matrix grows** — `ci_tests.yml:172-176`, `strict_status.yml:32,41`

`ci-tests-expected-failures` is deliberately a matrix, but the artifact name is the constant `strict-job-outcome`. `upload-artifact` v4+ (v6.0.0 here) **errors when an artifact of the same name already exists in the run** — it is no longer last-write-wins. Add a second `include:` row and its upload fails with "an artifact with this name already exists". `strict_status.yml` likewise downloads the one fixed name and reads a single file, so the whole reporting path is structurally single-entry. Inert today (one row), guaranteed breakage on the first extension.

Suggested fix: `name: strict-job-outcome-${{ matrix.name }}`, download by pattern in `strict_status.yml`, iterate; or document that the job must stay single-entry.

**B5. Infra failures are reported as "expected test failure" — or produce no check at all** — `ci_tests.yml:158-176`, `strict_status.yml:36`

Two related holes:
- *Conflation:* `steps.tests.outcome == 'failure'` for *any* nonzero tox exit — env-creation failure, `array-api-strict>=2.0` failing to resolve, tox internal error — not just failing tests. All get labeled ":warning: tests failed (expected until the remaining array-API bugs are fixed)" and a grey neutral check. A genuinely broken strict environment is rebranded as routine.
- *Silent gap:* "Record test outcome" and "Upload test outcome" have no `if: always()`. If an earlier step fails ("Install base dependencies"), they're skipped → no artifact → `strict_status.yml`'s download fails (tolerated) → "Create check run" is skipped → **no strict check is posted for that commit at all**.

Note the naive fix interacts badly: with bare `if: always()`, a skipped tests step writes `skipped` to the file, which `strict_status.yml` maps to `passed === false` → "failing (expected)". The pipeline needs a third state ("infra error", e.g. `conclusion: failure` on the check run), or at minimum a comment acknowledging the limitation.

**B6. Check-run name encodes the outcome** — `strict_status.yml:47`

`py313-strict: passing` vs `py313-strict: failing (expected)` are two different checks to GitHub. Consequences: it can never be a branch-protection required check; the check "identity" flips names across pushes as the outcome changes; anything keyed on a stable check name breaks. If informational-only is the intent, say so in a comment — better, use a stable `name: 'py313-strict'` and vary only `conclusion` and `output.title`.

**B7. The `jax` xfail on `test_generator_ccds_without_unit` is stale — the test XPASSes** — `ccdproc/tests/test_image_collection.py:~386`

The reason says "ccds() on the jax backend does not raise ValueError when the files have no unit" — but that `ValueError` comes from `fits_ccddata_reader` at `image_collection.py:960`, *before* the backend conversion block (`:963-973`), so the backend cannot affect whether it raises. Verified empirically:
```
CCDPROC_ARRAY_LIBRARY=jax python -m pytest ccdproc/tests/test_image_collection.py -q -rX
# → 76 passed, 1 xpassed (this test) — same result in isolation and full-module runs
```
The marker is wrong (or was recorded against an environment quirk that no longer exists) and, because `backend_xfail` maps to `strict=False`, it XPASSes silently forever — a live demonstration of D5 below. Remove the marker; if it originally failed for someone, the failure was a different exception (e.g. a warning promoted to error), not the stated mechanism. (The other backend_xfails — astroscrappy/scipy/reproject/block_reduce — all sit on genuine numpy-only boundaries; this is the one outlier.)

---

### Bugs (plausible — mechanism verified, occurrence depends on conditions)

**P1. Duplicate strict checks on same-repo PRs** — `ci_tests.yml:3-13`, `strict_status.yml`

`on: push` (all branches) + `pull_request` both fire for a branch pushed to the main repo; the concurrency group keys on `github.ref`, which differs (`refs/heads/…` vs `refs/pull/N/merge`), so neither run cancels the other. Two completed "CI" runs → two `workflow_run` events → two `checks.create` calls on the same `head_sha` → two (possibly contradictory, since the merge-ref run can differ) strict checks on the PR. Fix: gate `strict_status.yml` on `github.event.workflow_run.event == 'pull_request'` (which also stops noise checks from `schedule`/`push`-to-main runs).

**P2. Baseline keys use frame *function names* → fragile across Python versions and comprehensions** — `_escape_triage.py` (`record_escape_log` keying on `frame.name`)

An escape inside a genexpr/lambda keys as `<genexpr>`/`<lambda>` on every version; list comprehensions key as `<listcomp>` on py<3.12 but as the enclosing function on 3.12+ (PEP 709). The enforce CI job is py312; the documented refresh is "locally", unpinned. A contributor regenerating on 3.11 produces keys that mismatch CI → spurious NEW-escape failure + spurious stale entry. Fix: document a Python floor for regeneration, or normalize `<listcomp>`-style names to the enclosing frame.

**P3. Escape logging is not thread-safe; the reentrancy guard is process-global** — `conftest.py` (`_ReentrancyGuard`, shared across all three wrappers)

Under dask's threaded scheduler, thread A inside a wrapped `np.asarray` sets the guard; a concurrent escape on thread B sees `active` and is skipped entirely; and the `defaultdict[key] += 1` first-insert isn't atomic. Both drop real escapes (ratchet false-negatives), never false alarms. Fix: `threading.local()` guard + a lock around the tally.

---

### Design limitations that should at least be documented

**D1. The logger only sees Python-level `np.asarray`/`np.asanyarray`/`np.ma.asanyarray` module-attribute calls.** C-level coercions inside scipy/astroscrappy/reproject (exactly the BOUNDARY deps the baseline enumerates) and any `from numpy import asarray` bound before the patch are invisible — the ratchet has structural false negatives. The module docstring ("catch silent conversion back to numpy") oversells it.

**D2. xdist silently defeats the ratchet.** `_ESCAPE_LOG_COUNTS` is a per-process global; with `pytest -n`, workers accumulate and the controller sees nothing → "baseline not checked", green. Not live today (no `-n` in tox.ini) — worth an assert or comment before someone adds it.

**D3. Strict-job device mismatch noise.** Fixture data is created on `device1` (`pytest_fixtures.py`), but inline `xp.ones(...)`/`xp.asarray(...)` in test bodies land on the default device; array-api-strict raises on cross-device ops. Part of the 109 strict failures are harness artifacts, not ccdproc bugs — diluting the signal the job exists to produce. Worth a helper (or fixture-provided `make_array`) that pins the device.

**D4. `py313-strict` installs only `extras = test` + `array-api-strict` — no alldeps** (`tox.ini`), so numpy-boundary paths through scipy/reproject/astroscrappy aren't exercised there. Probably intentional; one comment line would prevent someone "fixing" it.

**D5. `strict=False` backend_xfails never flag stale markers.** Correct choice given `xfail_strict = true` globally, but a fixed backend bug becomes a silent XPASS forever (see B7). A note in the marker docs ("prune on XPASS") is enough.

---

### Refactoring suggestions

**R1. `_escape_triage.py` is three modules in a trenchcoat**: (1) failure-triage grouping (`pytest_runtest_makereport` + `_ESCAPE_SITES` report), (2) live escape-log tally + reporting, (3) baseline load/write/enforce. Meanwhile the monkeypatching lives in `conftest.py` but its helpers (`locate_escape_site`, `record_escape_log`) live here, creating a conftest↔module ping-pong. Suggest `_escape_log.py` (patching + recording + summary) and `_escape_ratchet.py` (load/write/enforce/report); "triage" would then honestly name what's left.

**R2. Hook registration via `from ._escape_triage import pytest_sessionfinish  # noqa`** (`conftest.py:27-33`) works and is commented, but `pytest_plugins = ["ccdproc.tests._escape_triage"]` is the idiomatic, less-fragile spelling (the noqa'd imports are load-bearing and an overzealous linter autofix could silently remove them).

**R3. Implicit contract between the two workflows** — the outcome file format (`success`/`failure` text) is shared between `ci_tests.yml` and `strict_status.yml` with no shared definition; a one-line schema comment in both places (or writing JSON with an explicit `state:` field, which also solves B5's third state) would harden it.

**R4. tox factor coupling**: `enforce` sets both `CCDPROC_LOG_ARRAY_ESCAPES` and `CCDPROC_ENFORCE_ESCAPE_BASELINE` — currently the only place the B2 invariant is maintained. After fixing B2 in code, keep the factor as the single source of the pairing. Also: nothing explains that the CI `strict`/`enforce` envs deliberately omit the `test` factor so they run from the source tree (not `.tmp/{envname}`); a future "normalization" to `py312-test-…` would silently change import paths.

---

### Documentation

`docs/array_api.rst` documents the strict backend, `CCDPROC_TRIAGE_ESCAPES`, `CCDPROC_LOG_ARRAY_ESCAPES`, and the markers — but has *zero* mention of the ratchet (`grep -iE 'baseline|enforce|ratchet' docs/array_api.rst` → nothing). Missing, in rough priority order:

1. **Regenerating the baseline** — that it requires *all three* of `CCDPROC_WRITE_ESCAPE_BASELINE=1` + `CCDPROC_LOG_ARRAY_ESCAPES=1` + a non-numpy `CCDPROC_ARRAY_LIBRARY`, over the **full** suite, in the **source tree**. Today that's split between a tox.ini comment and the baseline-file header, and the header's one-liner is the B1 footgun.
2. **What to do when the ratchet fires** on your PR (`py312-alldeps-dask-enforce` goes red): read the "NEW escapes" section of the log, either migrate the call site or (deliberately) add a baseline line.
3. **How to read the strict CI signal**: the `ubuntu-py313-strict` check is always green (carve-out); the real outcome is the grey `py313-strict: failing (expected)` / green `py313-strict: passing` check (only after this PR merges, since `workflow_run` executes from the default branch), plus the run's step summary. Nothing user-facing explains the grey check.
4. `CCDPROC_ENFORCE_ESCAPE_BASELINE` and `CCDPROC_WRITE_ESCAPE_BASELINE` aren't listed alongside the other env vars at all.

---

### Verified non-issues (so nobody re-litigates them)

Backend-name spellings in all `backend_xfail`/`backend_skip` marks match `_normalize_backend_name` (no silent never-applies typos); markers registered for `--strict-markers`; no filterwarnings loosening; no weakened assertions anywhere in the test diff; action SHAs/versions all real and current; cross-workflow artifact download inputs correct; `outcome` (not `conclusion`) correctly read under `continue-on-error`; `session.exitstatus = 1` in `pytest_sessionfinish` genuinely fails the run; baseline path is `__file__`-anchored (changedir-safe); missing baseline file fails loud; `_env_truthy` handles unset/"0"/"true".

---

### Suggested fix priority

1. B1 + B2 (one-line guards; they undermine the tool's core promise) and B3 (trivial).
2. B5 + B6 (strict-status robustness: third state, stable check name, `if: always()` with skipped-handling).
3. B4 + P1 (artifact-per-matrix-name; gate `workflow_run` on `pull_request`).
4. B7 (delete the stale jax marker).
5. Docs section (regeneration + ratchet-fired + grey-check).
6. P2/P3/D1–D5/R1–R4 as follow-ups or comments.

ccdproc/tests/_escape_triage.py
```diff
@@ -0,0 +1,533 @@
...
+        "# that requires CCDPROC_LOG_ARRAY_ESCAPES=1, a non-numpy",
+        "# CCDPROC_ARRAY_LIBRARY (e.g. dask), and a *full* test-suite run --",
+        "# a subset run drops the entries its tests never exercise.",
+        "#",
+        "# Tags are for humans, not the ratchet: TODO = still to migrate,",
+        "# BOUNDARY = a numpy-only dependency (scipy/astroscrappy/reproject)",
+        "# that will never leave. Verify/adjust the seeded tags by hand.",
+        "#",
+    ]
+    w_file = max(len(f) for f, _, _ in sites)
```
mwcraig: Repeating this comprehension 3 times seems odd...

ccdproc/conftest.py
```diff
@@ -60,8 +83,218 @@ def pytest_configure(config):
...
+
+
+def _normalize_backend_name(name):
+    return name.lower().replace("_", "-")
+
+
+_ACTIVE_BACKEND = _normalize_backend_name(array_library)
+
+
+def pytest_collection_modifyitems(items):
```
mwcraig, line 131: Add a docstring and add some inline comments to t code.

ccdproc/conftest.py
```diff
@@ -60,8 +83,218 @@ def pytest_configure(config):
...
+
+
+class _ReentrancyGuard:
+    """Small helper to keep our wrappers from recursing into themselves."""
+
+    def __init__(self):
+        self.active = False
+
+
+def _make_escape_logging_wrapper(original, funcname, guard):
```
mwcraig, line 249: Add comments explaining what this does.

ccdproc/tests/_escape_triage.py
mwcraig: Agreed, it's repetitive — all three are just computing per-column widths so the baseline file's columns line up. They can collapse to one transpose-and-measure, e.g.:

```python
w_file, w_func, w_co = (max(len(s) for s in col) for col in zip(*sites))
```

(`sites` is guaranteed non-empty at that point because of the empty-run guard above, so `max()` can't blow up.) Will fold that into the next commit on this branch rather than pushing a one-liner now.

ccdproc/conftest.py
mwcraig: Will do in the next commit. For the record, what it does: this is the pytest collection hook that makes the `backend_skip` / `backend_xfail` markers work. For every collected test it checks each marker's backend arguments (normalized so `array_api_strict` and `array-api-strict` match) against the active backend from `CCDPROC_ARRAY_LIBRARY`; on a match it attaches a plain `pytest.mark.skip` or a **non-strict** `pytest.mark.xfail` with the marker's reason. The `break` after each match just stops scanning further markers of the same kind once one has fired. Docstring + inline comments to that effect coming.

ccdproc/conftest.py
mwcraig: Will add comments in the next commit. Summary of what it does: this builds the wrapper that replaces one numpy coercion entry point (`np.asarray`, `np.asanyarray`, or `np.ma.asanyarray`) while the escape logger is active. On every call it (1) skips the check if the reentrancy guard is held or there are no positional args, otherwise (2) asks `_foreign_namespace()` whether the first argument is an array from a non-numpy array-API library, and if so (3) logs the "array-API escape" warning with the blamed ccdproc frame and tallies it via `record_escape_log()` for the terminal summary and the baseline ratchet. In all cases it ends by calling the original numpy function, so behavior is unchanged. The guard exists because the namespace detection and frame extraction can themselves call the patched functions — without it the wrapper would recurse.

ccdproc/tests/_escape_triage.py
mwcraig: Done in d329ac5 — one caveat: ruff (B905) requires an explicit `strict=` on `zip()`, so it's `zip(*sites, strict=True)` split across two lines by black, not quite the one-liner sketched above.

ccdproc/conftest.py
mwcraig: Docstring added in d329ac5, plus a short comment on the `break`.

ccdproc/conftest.py
mwcraig: Documented in d329ac5: a docstring on the factory covering the log-and-tally behavior and the reentrancy guard, plus inline comments at the guard check and the foreign-array branch.

PR #925, opened 2026-06-05
Use array-API dtype in Combiner; fix stale test

karlhillx:
Fixes #924.

Two changes:

- `ccdproc/combiner.py:603` — `weights.astype("float64")` → `xp.astype(weights, xp.float64)`. The actual JAX deprecation.
- `ccdproc/tests/test_image_collection.py:387-401` — `ccds()` raises `ValueError` in every backend except JAX, so the test branches: asserts the raise in numpy/dask, asserts a CCDData is returned in JAX. Same pattern as `test_memory_use.py`.

The py313-jax job still has 6 failures from `np.result_type` in `astropy.nddata.mixins.ndarithmetic:769` when scaling CCDData. That's astropy code, not ccdproc, and predates this PR — worth a separate issue upstream.

cc @mwcraig

PR #952, opened 2026-07-19
Exclude masked weights from average combinations

### LINKED ISSUE #893: Weights are not properly applied for masked and clipped arrays in `combine()`
I'm trying to combine images with `ccdproc.combine()`. 

The following is an example to show the issue. Note that I set the `mask = [[1, 0, 0], [0, 0, 0]]` for the first image. 

```python
import numpy as np
from astropy.nddata import CCDData
from ccdproc import combine

ccd1 = CCDData([[1, 1, 1], [2, 2, 2]], mask=[[1, 0, 0], [0, 0, 0]], unit="adu")
ccd2 = CCDData([[1, 1, 1], [2, 2, 2]], unit="adu")

weights = np.array([1, 1])
weights = None

combined_ccd = combine(
    [ccd1, ccd2],
    method="average",
    weights=weights,
)

print(combined_ccd.data)
```

With `weights = None`, the result is `[[1., 1., 1.], [2., 2., 2.]]` as expected. 

On the other hand, with `weights = np.array([1, 1])`, the result becomes `[[0.5 1.  1. ], [2., 2., 2.]]`, while I still expect the result is `[[1, 1, 1], [2, 2, 2]]`.

I think that the `combine()` does not remove the masked elements from the `weights` array, so smaller value are returned. 

A similar behavior is also seen in the clipped arrays.

My Python environment is the following:
- Python 3.11.12
- ccdproc 2.4.3
- Numpy 2.2.5
- Astropy 7.0.2

nomad3:
## Summary

- exclude weights belonging to masked, clipped, or non-finite samples from the weighted-average denominator
- preserve the existing weighted numerator, custom callback behavior, weighted sums, and clean-stack allocation path
- add public and direct regression coverage for frame and pixel weights, clipping, fully masked output, custom callbacks, and weighted sums

Fixes #893

## Validation

- `python -m pytest ccdproc -q` — 337 passed, 29 skipped
- `python -m pytest ccdproc/tests/test_combiner.py -q` — 80 passed
- NumPy weight-focused tests — 12 passed
- Dask weight-focused tests without Bottleneck — 12 passed
- JAX weight-focused tests with 64-bit enabled — 12 passed
- `ruff check ccdproc/combiner.py ccdproc/tests/test_combiner.py`
- `black --check ccdproc/combiner.py ccdproc/tests/test_combiner.py`
- `git diff --check`

The array-api-strict backend was not counted as patch validation because current `main` fails earlier in existing `Combiner` construction and public `combine()` code. PR #925 changes an adjacent dtype-conversion line; this PR deliberately leaves that conversion untouched.

## AI assistance disclosure

I used OpenAI Codex to assist with repository inspection, test design, implementation, and review. I reviewed the resulting changes and take responsibility for their correctness and maintenance.

## Checklist

- [x] For new contributors: Did you add yourself to the `AUTHORS.rst` file?
- [ ] For documentation changes: Does your commit message include a `[skip ci]`? (Not applicable.)
- [x] Did you add an entry to the `CHANGES.rst` file?
- [x] Does the commit message include a `Fixes #issue_number`?
- [ ] Does this PR add, rename, move or remove any existing functions or parameters? (No.)


ccdproc/combiner.py
```diff
@@ -661,6 +662,8 @@ def average_combine(
         # we get to the uncertainty calculation.
         if self.weights is not None:
             weighted_sum, weights = self._weighted_sum(data, sum_func, xp=xp)
+            if use_default_scale_func and masked_values.any():
```
mwcraig: Please use the array api `any` instead of the `any` method:


```suggestion
            if use_default_scale_func and xp.any(masked_values):
```

mwcraig (changes requested):
Thanks @nomad3 -- this looks great, just one minor change so that we do not create an additional array API issue down the road.

ccdproc/combiner.py
nomad3: Updated in 933610c to use `xp.any(masked_values)`, keeping the conditional namespace-aware as suggested. The focused combiner suite passes (80 tests), along with Ruff, Black, and `git diff --check`. Thanks!

ccdproc/tests/test_combiner.py
```diff
@@ -168,6 +168,94 @@ def test_pixelwise_weights():
...
+    expected = xp.asarray([[7.75, 15.5]])
+    assert xp.all(xpx.isclose(combined.data, expected))
+
+
+@pytest.mark.filterwarnings(
+    "ignore:Degrees of freedom <= 0 for slice.:RuntimeWarning",
+    "ignore:invalid value encountered in divide:RuntimeWarning",
+)
+def test_combiner_weighted_average_fully_masked():
+    mask = xp.ones((1, 1), dtype=xp.bool)
```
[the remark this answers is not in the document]
nomad3: Updated in 933610c to use the built-in `bool` dtype for portability across array backends. The focused combiner suite passes (80 tests), along with Ruff, Black, and `git diff --check`.

mwcraig (approved):
Thanks for the PR< and apologies for the delayed merge!

PR #953, opened 2026-07-19
Return plain CCDData from array API wrappers

### LINKED ISSUE #927: Array API: _unwrap_ccddata_for_array_api never converts the wrapper back to CCDData
### Problem

`_CCDDataWrapperForArrayAPI` is a subclass of `CCDData`:

https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/_ccddata_wrapper_for_array_api.py#L37

In `_unwrap_ccddata_for_array_api`, the `isinstance(ccd, CCDData)` check comes first, so it is *always* true for the wrapper and the function returns the wrapper unchanged. The conversion back to plain `CCDData` and the `TypeError` branch are unreachable:

https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/_ccddata_wrapper_for_array_api.py#L373-L390

### Consequence

Public functions in `core.py` that wrap/unwrap (e.g. `flat_correct`) return the private wrapper type to users instead of `CCDData`. The uncertainty is still unwrapped (that happens before the early return), which also means a returned wrapper carries a non-wrapped uncertainty — an inconsistent state if it is passed back into another ccdproc function.

### Suggested fix

Check for `_CCDDataWrapperForArrayAPI` *before* the `CCDData` check (or use `type(ccd) is CCDData`).

---
Found during a review of the array API implementation from #885; follow-up to #909.

nomad3:
## Summary

- make `_unwrap_ccddata_for_array_api` validate inputs before accessing uncertainty
- unwrap private uncertainty wrappers even when arithmetic has already produced a plain `CCDData`
- convert `_CCDDataWrapperForArrayAPI` results to concrete `CCDData` while preserving ordinary `CCDData` subclasses
- add focused public and helper-level regressions across the supported uncertainty types

Fixes #927

## Validation

- `python -m pytest ccdproc -q` — 339 passed, 29 skipped
- focused regression module — 8 passed with NumPy, Dask, and JAX
- focused regression module with `array-api-strict` — 7 passed, 1 documented xfail for the pre-existing Astropy uncertainty-propagation boundary
- Ruff, Black, and `git diff --check` — passed

## Checklist

- [ ] For new contributors: Did you add yourself to the `AUTHORS.rst` file? (The entry is already proposed in draft #952, so it is intentionally not duplicated here.)
- [x] For bugfixes: Does the commit message include `Fixes #927`?

## AI assistance disclosure

OpenAI Codex was used for source inspection, test design, implementation, and automated review. This PR remains a draft pending the contributor's personal review; the contributor will take responsibility for the final contribution before marking it ready.

### DIFF
--- a/ccdproc/_ccddata_wrapper_for_array_api.py
+++ b/ccdproc/_ccddata_wrapper_for_array_api.py
@@ -375,16 +375,22 @@ def _unwrap_ccddata_for_array_api(ccd):
     Unwrap a CCDData object from array API backends to the original CCDData.
     """
 
-    if ccd.uncertainty is not None:
-        ccd.uncertainty = _unwrap_uncertainty(ccd.uncertainty)
-
-    if isinstance(ccd, CCDData):
-        return ccd
-
-    if not isinstance(ccd, _CCDDataWrapperForArrayAPI):
+    if not isinstance(ccd, CCDData):
         raise TypeError(
             "Input must be a CCDData or _CCDDataWrapperForArrayAPI instance."
         )
 
-    # Convert back to CCDData
-    return CCDData(ccd)
+    if isinstance(
+        ccd.uncertainty,
+        (
+            _StdDevUncertaintyWrapper,
+            _VarianceUncertaintyWrapper,
+            _InverseVarianceWrapper,
+        ),
+    ):
+        ccd.uncertainty = _unwrap_uncertainty(ccd.uncertainty)
+
+    if isinstance(ccd, _CCDDataWrapperForArrayAPI):
+        return CCDData(ccd)
+
+    return ccd
--- a/ccdproc/_ccddata_wrapper_for_array_api.py
+++ b/ccdproc/_ccddata_wrapper_for_array_api.py
@@ -391,6 +391,13 @@ def _unwrap_ccddata_for_array_api(ccd):
         ccd.uncertainty = _unwrap_uncertainty(ccd.uncertainty)
 
     if isinstance(ccd, _CCDDataWrapperForArrayAPI):
-        return CCDData(ccd)
+        # Do not copy-construct through CCDData here. Astropy's mask setter
+        # coerces masks with np.asarray(..., dtype=bool), which would pull
+        # non-NumPy masks out of their array namespace. The wrapper is a
+        # state-free Python subclass created as a copy by
+        # _wrap_ccddata_for_array_api, so changing its public class identity
+        # does not mutate the caller's CCDData or disturb its backend arrays.
+        ccd.__class__ = CCDData
+        return ccd
 
     return ccd
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,8 @@ Bug Fixes
   namespace form ``xp.astype(weights, xp.float64)`` instead of the deprecated
   string-based ``.astype("float64")``. This resolves the ``DeprecationWarning``
   that was being promoted to a test failure in the JAX CI job. [#924]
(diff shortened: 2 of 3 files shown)


ccdproc/_ccddata_wrapper_for_array_api.py
```diff
@@ -375,16 +375,22 @@ def _unwrap_ccddata_for_array_api(ccd):
...
+        (
+            _StdDevUncertaintyWrapper,
+            _VarianceUncertaintyWrapper,
+            _InverseVarianceWrapper,
+        ),
+    ):
+        ccd.uncertainty = _unwrap_uncertainty(ccd.uncertainty)
+
+    if isinstance(ccd, _CCDDataWrapperForArrayAPI):
+        return CCDData(ccd)
```
mwcraig: This is the line that's breaking the dask CI jobs (`ubuntu-py312-dask-escape-baseline` and `macos-py312-dask` both fail `test_transform_image[True-True]` with `TypeError: Multiple namespaces for array inputs`).

`CCDData(ccd)` re-runs the astropy constructor, and astropy's `NDDataArray.mask` setter coerces the mask with `np.asarray(value, dtype=np.bool_)`. So for a dask- or jax-backed image, `data` and `uncertainty.array` survive in the backend library but `mask` comes back as a plain numpy array -- reintroducing exactly the coercion `_CCDDataWrapperForArrayAPI` exists to prevent (its own `mask` setter uses `xp.asarray` from the data's namespace instead).

Since the wrapper is a plain-Python subclass of `CCDData` with no extra state, you can convert it without round-tripping through the constructor:

```suggestion
        # Avoid CCDData(ccd): astropy's mask setter coerces the mask with
        # np.asarray(..., dtype=bool), which would pull non-numpy masks
        # back to numpy.
        ccd.__class__ = CCDData
        return ccd
```

There's no aliasing concern -- `_wrap_ccddata_for_array_api` creates a new wrapper object, so mutating it here never touches a caller-owned `CCDData`. I've verified this variant locally: the full suite passes under numpy (339), dask (334), and jax with x64 (spot-checked, including the previously failing transform test), and after a wrap->unwrap round trip under dask, `data`, `mask`, and `uncertainty.array` all stay dask arrays while `type(result) is CCDData`.

mwcraig (changes requested):
Thanks for digging into #927 -- the diagnosis is right, and the reordering here is a real fix for the wrapper leaking out of public functions (on `main`, the early `isinstance(ccd, CCDData)` return made the `CCDData(ccd)` conversion unreachable, since the wrapper is a `CCDData` subclass). The test coverage is also very welcome.

Two requests:

1. **The conversion itself needs a different mechanism** -- see the inline comment. Copy-constructing through `CCDData(ccd)` coerces the mask to numpy via astropy's setter, which is what's failing the dask jobs on this PR. Swapping `__class__` in place fixes the type identity without disturbing any of the backend arrays.
2. **Please rebase onto current `main` and mark the PR ready for review.** `main` has moved since this branch was cut (#952, #925, and #961 have merged, including changes to `CHANGES.rst` that will conflict trivially), and CI here is a month stale. Once it's rebased with the fix above, the dask jobs should go green and I'm happy to take a final look.

nomad3:
Addressed the requested in-place change, rebased on main, and reran the relevant tests. Ready for another look—thanks!

mwcraig (approved):
Thanks @nomad3!

PR #954, opened 2026-07-19
Keep clip_extrema indices in the array namespace

### LINKED ISSUE #928: Array API: clip_extrema uses numpy ravel_multi_index on index arrays from the array namespace
### Problem

`Combiner.clip_extrema` calls numpy's `ravel_multi_index` on index arrays that come from the user's array namespace:

https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/combiner.py#L21

https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/combiner.py#L361

With CuPy data the `where` indices are CuPy arrays, so passing them to a numpy function triggers the implicit device-to-host conversion error (`TypeError: Implicit conversion to a NumPy array is not allowed`). This is one of the failure modes in #911.

### Suggested fix

`ravel_multi_index` is not part of the array API standard, but the flat index can be computed with plain namespace arithmetic (e.g. `i0 * stride0 + i1 * stride1 + ...`), keeping everything in `xp`.

---
Found during a review of the array API implementation from #885; follow-up to #909 / #911.

nomad3:
## Summary

- replace NumPy `ravel_multi_index` with equivalent C-order arithmetic in the selected array namespace
- keep coordinate indices on the input array device and use Array API-compatible reshaping/indexing
- add an asymmetric regression that checks the exact high/low flat indices and non-default-device preservation

Fixes #928

## Validation

- `python -m pytest ccdproc -q` — 380 passed, 29 skipped
- `python -m pytest ccdproc/tests/test_combiner.py -q` — 87 passed with NumPy and 87 passed with Dask
- JAX Combiner module with 64-bit enabled — 81 passed, 6 established xfails
- focused regression with `array-api-strict` on a non-default device — 1 passed
- Ruff, Black, `git diff --check`, and `git show --check` — passed

`array-api-strict` deliberately excludes integer-array assignment, so the focused strict regression replaces only the final `xpx.at` update with a test spy. The production update path is exercised by the NumPy, Dask, and JAX runs; strict verifies that the exact C-order indices are computed in the selected namespace and remain on the data device.

## Checklist

- [ ] For new contributors: Did you add yourself to the `AUTHORS.rst` file? (The entry is already present on current main, so it is intentionally not duplicated here.)
- [x] For bugfixes: Does the commit message include `Fixes #928`?

## AI assistance disclosure

OpenAI Codex was used for source inspection, test design, implementation, and automated review. I have personally reviewed the current diff and validation evidence and take responsibility for this contribution.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,13 @@ Bug Fixes
 - Use Bottleneck combination functions only with NumPy arrays, preserving the
   selected Array API namespace for other backends. [#904, #959]
 
+Bug Fixes
+^^^^^^^^^
+
+- Keep ``Combiner.clip_extrema`` index arithmetic in the selected array
+  namespace so GPU-backed arrays do not require an implicit conversion to
+  NumPy.
+
 2.5.1 (2025-07-05)
 ------------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,10 +37,6 @@ Bug Fixes
 
 - Use Bottleneck combination functions only with NumPy arrays, preserving the
   selected Array API namespace for other backends. [#904, #959]
-
-Bug Fixes
-^^^^^^^^^
-
 - Keep ``Combiner.clip_extrema`` index arithmetic in the selected array
   namespace so GPU-backed arrays do not require an implicit conversion to
   NumPy. [#954]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -43,7 +43,7 @@ Bug Fixes
 
 - Keep ``Combiner.clip_extrema`` index arithmetic in the selected array
   namespace so GPU-backed arrays do not require an implicit conversion to
-  NumPy.
+  NumPy. [#954]
 
 2.5.1 (2025-07-05)
 ------------------
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -18,7 +18,6 @@ from astropy.nddata import CCDData, StdDevUncertainty
 from astropy.stats import sigma_clip
 from astropy.utils import deprecated_renamed_argument
 from numpy import mgrid as np_mgrid
-from numpy import ravel_multi_index as np_ravel_multi_index
 
 from .core import sigma_func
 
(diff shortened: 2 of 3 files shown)


mwcraig:
> This PR remains a draft pending the contributor's personal review; the contributor will take responsibility for the final contribution before marking it ready.

Can you please go ahead and make this non-draft when you have a chance to look it over?

mwcraig:
@nomad3 -- is this one ready for review?

nomad3:
Yes—thank you for checking. I have personally reviewed the rebased diff and fresh validation, corrected the changelog placement, and marked it ready for review. All completed exact-head CI jobs are green; the remaining Dask+Bottleneck matrix job is still queued.

mwcraig (approved):
*This review was drafted by Claude Code and is being posted on behalf of @mwcraig, who reviewed and approved its contents.*

Thanks @nomad3 — this is a careful fix and an unusually well-constructed test.

I verified the replacement arithmetic by hand: the accumulation loop over `shape[1:]` is exactly C-order `ravel_multi_index`, and the `strict=True` zip guards against a coordinate/shape mismatch. Two quiet improvements ride along: `argsorted[(i, ...)]` removes the old hardcoded assumption of a 3D stack (which also violated array-api-strict indexing rules for 3D-image stacks), and creating `mg` on the data device removes the implicit host-to-device transfer that #928 was about. I also recomputed the test's hardcoded flat indices and expected mask for the 3×2×3 data and they're correct.

One optional suggestion — happy to merge as-is if you'd rather not:

- The test builds the `Combiner` via `object.__new__` and sets `_xp` / `_data_arr` / `_data_arr_mask` directly, which couples it to private internals for every backend even though only array-api-strict needs the bypass (it can't stack a list of arrays). If you're willing, branching so that non-strict backends construct the `Combiner` normally (from a list of `CCDData`) and only strict uses the bypass would make the test more robust against future changes to `clip_extrema`'s internals. Entirely non-blocking.

mwcraig (approved):
Thanks again @nomad3!

PR #955, opened 2026-07-19
Make array API escape logging thread-safe

### LINKED ISSUE #944: Escape logger is not thread-safe (missed escapes under threaded schedulers)
Follow-up from PR #942 (review item P3, https://github.com/astropy/ccdproc/pull/942#pullrequestreview-4695447403).

The escape logger installed by `ccdproc/conftest.py` uses a single process-global `_ReentrancyGuard` shared by all three wrapped functions (`np.asarray`, `np.asanyarray`, `np.ma.asanyarray`). Under a threaded scheduler (e.g. dask's default threaded scheduler):

- While thread A holds the guard (`guard.active = True`), a genuine escape occurring concurrently on thread B is silently skipped — the wrapper treats it as a re-entrant call.
- The tally update in `record_escape_log()` (`_ESCAPE_LOG_COUNTS[key] += 1` on a `defaultdict`) is not atomic on first insert.

Both failure modes *drop* real escapes, so the baseline ratchet (`CCDPROC_ENFORCE_ESCAPE_BASELINE=1`) can produce false negatives — never false alarms — meaning a new escape site reached only from worker threads could slip past CI.

Fix sketch:
- Replace the shared guard with `threading.local()` so re-entrancy is tracked per thread.
- Take a `threading.Lock` around the tally increment (or use a thread-safe counter).

nomad3:
## Summary

- keep the shared escape-wrapper reentrancy guard while making its active state local to each worker thread
- serialize the complete escape-count read-modify-write operation
- add deterministic regressions for cross-wrapper same-thread recursion, concurrent wrapper inspection, and protected tally updates

Fixes #944

## Validation

- `python -m pytest ccdproc/tests/test_escape_triage.py -q` — 2 passed
- refreshed full NumPy suite — 348 passed, 29 skipped
- refreshed Dask escape-baseline enforcement — 343 passed, 34 skipped; 191 escapes at 22 sites; no escapes outside the baseline
- Ruff, Black, `git diff --check`, and `git show --check` — passed

The guard remains shared across all three NumPy wrappers, preserving same-thread cross-wrapper recursion suppression. Only its `active` state becomes thread-local. The tally lock covers the full `counts[key] += 1` operation while preserving the existing keys and reporting format.

## Checklist

- [x] For new contributors: Did you add yourself to the `AUTHORS.rst` file? (The entry is present on `main` through merged PR #952.)
- [x] For bugfixes: Does the commit message include `Fixes #944`?

## AI assistance disclosure

### DIFF
--- a/ccdproc/tests/test_escape_triage.py
+++ b/ccdproc/tests/test_escape_triage.py
@@ -1,0 +1,102 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+from concurrent.futures import ThreadPoolExecutor
+from threading import Event, Lock, get_ident
+
+from ccdproc import conftest as escape_log
+from ccdproc.tests import _escape_triage
+
+
+def test_escape_wrapper_guard_is_thread_local(monkeypatch):
+    first_entered = Event()
+    release_first = Event()
+    inspected = []
+    logged = []
+    original_calls = []
+
+    def original(value):
+        original_calls.append(value)
+        return value
+
+    guard = escape_log._ReentrancyGuard()
+    wrapper = escape_log._make_escape_logging_wrapper(original, "numpy.asarray", guard)
+    nested_wrapper = escape_log._make_escape_logging_wrapper(
+        original, "numpy.asanyarray", guard
+    )
+
+    def foreign_namespace(value):
+        inspected.append(value)
+        if value == "first":
+            assert nested_wrapper("nested") == "nested"
+            first_entered.set()
+            assert release_first.wait(timeout=5)
+        return "foreign"
+
+    def record_escape(_frame, funcname):
+        logged.append(funcname)
+
+    monkeypatch.setattr(escape_log, "_foreign_namespace", foreign_namespace)
+    monkeypatch.setattr(escape_log, "_escape_site_frame", lambda: None)
+    monkeypatch.setattr(escape_log, "record_escape_log", record_escape)
+
+    with ThreadPoolExecutor(max_workers=2) as executor:
+        first = executor.submit(wrapper, "first")
+        try:
+            assert first_entered.wait(timeout=5)
+            second = executor.submit(wrapper, "second")
+            assert second.result(timeout=5) == "second"
+        finally:
+            release_first.set()
+        assert first.result(timeout=5) == "first"
+
+    assert inspected == ["first", "second"]
+    assert logged == ["numpy.asarray", "numpy.asarray"]
+    assert original_calls == ["nested", "second", "first"]
+
+
+def test_escape_log_tally_increment_is_serialized(monkeypatch):
+    class TrackingLock:
+        def __init__(self):
+            self.lock = Lock()
+            self.owner = None
+
+        def __enter__(self):
+            self.lock.acquire()
+            self.owner = get_ident()
+            return self
+
+        def __exit__(self, exc_type, exc_value, traceback):
+            self.owner = None
+            self.lock.release()
+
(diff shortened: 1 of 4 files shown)


ccdproc/tests/test_escape_triage.py
```diff
@@ -0,0 +1,102 @@
...
+            release_first.set()
+        assert first.result(timeout=5) == "first"
+
+    assert inspected == ["first", "second"]
+    assert logged == ["numpy.asarray", "numpy.asarray"]
+    assert original_calls == ["nested", "second", "first"]
+
+
+def test_escape_log_tally_increment_is_serialized(monkeypatch):
+    class TrackingLock:
```
mwcraig, line 62: Please add brief explanations of what each class is for -- either inline comments or a short docstring is fine. The purpose is to make it easier for future maintainers to understand the purpose of each of the components here.

mwcraig (changes requested):
Thanks — the production fix here is correct, and I verified it carefully: the `threading.local` guard semantics are right (`__init__` reruns per thread, so each thread gets its own `active` flag), the lock around the tally's read–modify–write closes the race, and both regression tests genuinely fail without the fix. The requested changes are entirely about the tests' isolation, not their substance.

**The tests monkeypatch the *live* escape logger, which makes them sensitive to unrelated numpy traffic.** `from ccdproc import conftest as escape_log` imports the same module object pytest loaded, and the session-installed `np.asarray`/`np.asanyarray` wrappers resolve `_foreign_namespace` and `record_escape_log` from that module's globals *at call time*. So in any escapes-enabled run (`CCDPROC_LOG_ARRAY_ESCAPES=1` — e.g. the dask escape-baseline CI job), these lines:

```python
monkeypatch.setattr(escape_log, "_foreign_namespace", foreign_namespace)
monkeypatch.setattr(escape_log, "record_escape_log", record_escape)
```

redirect *every* wrapped numpy call in the process — not just the test's own wrappers — into fakes that classify all inputs as foreign. Any incidental `np.asarray()` from another thread or plugin during the test window then lands in `inspected`/`logged` and breaks the exact-equality asserts. This is reproducible: with escapes enabled and a background thread doing nothing but `np.asarray([1, 2, 3])` in a loop, `test_escape_wrapper_guard_is_thread_local` fails with

```
AssertionError: assert ['first', [1, 2, 3], [1, 2, 3], ...] == ['first', 'second']
```

Today's serial CI happens to be quiet during that window, so it passes — but the test's correctness depends on an invariant (no concurrent numpy traffic) that nothing enforces, which is an awkward property for a test whose whole point is thread-safety. The window also silently diverts any *real* escapes that occur during it away from the tally.

**Requested changes:**

1. Make the fakes sentinel-scoped so incidental traffic passes through untouched, e.g.:

   ```python
   def foreign_namespace(value):
       if not isinstance(value, str) or value not in ("first", "second"):
           return None  # not one of this test's sentinels -- not foreign
       ...
   ```

   With that, unrelated calls take the not-foreign path, `record_escape` never sees them, and the exact-equality asserts are safe. (Alternatively, decouple the tests from the live module entirely by injecting the hooks into `_make_escape_logging_wrapper` — fine as a follow-up, the sentinel guard is enough here.)

2. Drop `raising=False` on the `_ESCAPE_LOG_COUNTS_LOCK` monkeypatch — the attribute exists (this PR adds it), and `raising=False` would let the test silently stop testing the lock if it's ever renamed.

With those two adjustments this is good to go — the fix itself is solid.

mwcraig (approved):
Thanks @nomad3

PR #956, opened 2026-07-19
Support immutable arrays in flat correction and masking

### LINKED ISSUE #931: Array API: in-place item assignment in flat_correct and ccdmask breaks immutable backends
### Problem

A few spots still mutate arrays with in-place item assignment, which fails on immutable backends (JAX) and is not part of the array API standard:

1. `flat_correct`:

   https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/core.py#L1009

2. `ccdmask` (`byblocks` branch): `csum[csum <= 0] = 0`, `csum[all_masked] = 1`, `block_mask[:, :] |= colmask[xp.newaxis, :]`, and `mask[l1:l2, c1:c2] = block_mask`, roughly here:

   https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/core.py#L2255-L2290

### Suggested fix

The codebase already uses `array_api_extra`'s `xpx.at(...)` for exactly this (e.g. in `background_deviation_box`, https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/core.py#L1327) — apply the same pattern here.

---
Found during a review of the array API implementation from #885; follow-up to #909.

nomad3:
## Summary

- rebind functional `xpx.at(...).set(...)` results in `flat_correct` and `ccdmask` so immutable Array API backends retain the updates
- use backend-portable stacking for column statistics and flatten blocks with `reshape((-1,))` so JAX/Dask reach the update paths
- add deterministic, backend-native regressions for masked flats, minimum flat values, block masks, all-masked columns, and column-gap filling while checking that inputs remain unchanged

## Testing

- `pytest -q ccdproc`: 350 passed, 29 skipped
- `CCDPROC_ARRAY_LIBRARY=jax JAX_ENABLE_X64=True pytest -q ccdproc`: 343 passed, 29 skipped, 7 expected failures
- `CCDPROC_ARRAY_LIBRARY=dask pytest -q ccdproc`: 345 passed, 34 skipped
- focused Array API strict run: 4 passed, 1 expected failure at the pre-existing CCDData mask-copy boundary
- Black, Ruff, and `git diff --check`

Fixes #931

## Checklist

- [x] For bugfixes: Does the commit message include `Fixes #931`?

## AI assistance disclosure

### DIFF
--- a/ccdproc/tests/test_ccdmask.py
+++ b/ccdproc/tests/test_ccdmask.py
@@ -5,7 +5,35 @@ import pytest
 from astropy.nddata import CCDData
 from numpy.testing import assert_array_equal
 
+from ccdproc.conftest import testing_array_device as xp_device
+from ccdproc.conftest import testing_array_library as xp
 from ccdproc.core import ccdmask
+
+
+def _ccdmask_in_active_namespace(data, **kwargs):
+    ratio = CCDData(xp.asarray(data, device=xp_device), unit="adu")
+    data_before = xp.asarray(data, device=xp_device, copy=True)
+
+    result = ccdmask(ratio, xp=xp, **kwargs)
+
+    assert xp.all(ratio.data == data_before)
+    return result
+
+
+def _ones_like_filter(data, size):
+    assert size
+    return xp.ones_like(data)
+
+
+def _zeros_like_percentile_filter(data, percentile, size):
+    assert percentile
+    assert size
+    return xp.zeros_like(data)
+
+
+def _zero_percentile(array, percentiles):
+    assert percentiles
+    return array[0] * 0
 
 
 def test_ccdmask_no_ccddata():
@@ -215,3 +243,65 @@ def test_ccdmask_pixels():
     mask = ccdmask(ratio, ncsig=11, nlsig=15, findbadcolumns=True)
     target_mask[:, 2] = True
     assert_array_equal(mask, target_mask)
+
+
+@pytest.mark.parametrize("findbadcolumns", [False, True])
+def test_ccdmask_byblocks_with_immutable_array(monkeypatch, findbadcolumns):
+    monkeypatch.setattr(
+        "ccdproc.core.ndimage.median_filter",
+        _ones_like_filter,
+    )
+    monkeypatch.setattr(
+        "ccdproc.core._percentile_fallback",
+        _zero_percentile,
+    )
+    data = np.ones((8, 8))
+    data[:4, 2] = 1000
+
+    mask = _ccdmask_in_active_namespace(
+        data,
+        byblocks=True,
+        findbadcolumns=findbadcolumns,
+        ncsig=4,
+        nlsig=4,
+        ncmed=3,
+        nlmed=3,
+        lsigma=3,
+        hsigma=3,
+        ngood=3,
+    )
+
+    expected = np.zeros(data.shape, dtype=bool)
+    expected[:4, 2] = True
+    assert xp.all(mask == xp.asarray(expected, device=xp_device))
+
+
+def test_ccdmask_column_gap_with_immutable_array(monkeypatch):
+    monkeypatch.setattr(
+        "ccdproc.core.ndimage.median_filter",
+        _ones_like_filter,
+    )
+    monkeypatch.setattr(
+        "ccdproc.core.ndimage.percentile_filter",
+        _zeros_like_percentile_filter,
+    )
+    data = np.ones((8, 8))
+    data[1, 3] = 1000
+    data[3, 3] = 1000
+
+    mask = _ccdmask_in_active_namespace(
+        data,
+        byblocks=False,
+        findbadcolumns=True,
+        ncsig=3,
+        nlsig=3,
+        ncmed=3,
+        nlmed=3,
+        lsigma=3,
+        hsigma=3,
+        ngood=4,
+    )
+
+    expected = np.zeros(data.shape, dtype=bool)
+    expected[1:4, 3] = True
+    assert xp.all(mask == xp.asarray(expected, device=xp_device))
--- a/ccdproc/tests/test_ccdmask.py
+++ b/ccdproc/tests/test_ccdmask.py
@@ -10,6 +10,8 @@ from ccdproc.conftest import testing_array_library as xp
 from ccdproc.core import ccdmask
 
 
+# Construct the CCD in the selected test namespace so every backend exercises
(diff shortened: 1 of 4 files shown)


ccdproc/tests/test_ccdmask.py
```diff
@@ -5,9 +5,37 @@
...
+
+def _zeros_like_percentile_filter(data, percentile, size):
+    assert percentile
+    assert size
+    return xp.zeros_like(data)
+
+
+def _zero_percentile(array, percentiles):
+    assert percentiles
+    return array[0] * 0
```
mwcraig: These asserts are truthiness checks on values that are always truthy at the call sites: `percentile` is `69.1` or `30.9` (so this is `assert 69.1`, which can never fail) and `size` is `(3, 3)` (a non-empty tuple). They read like argument validation but would pass if, say, `(nlsig, ncsig)` were passed where `(nlmed, ncmed)` belongs, or if the two percentiles were swapped. Checking the actual values costs nothing:

```suggestion
def _ones_like_filter(data, size):
    assert size == (3, 3)
    return xp.ones_like(data)


def _zeros_like_percentile_filter(data, percentile, size):
    assert percentile in (30.9, 69.1)
    assert size == (3, 3)
    return xp.zeros_like(data)


def _zero_percentile(array, percentiles):
    assert percentiles in (30.9, 69.1)
    return array[0] * 0
```

Comment generated by Claude Code

ccdproc/tests/test_ccdmask.py
```diff
@@ -215,3 +243,65 @@ def test_ccdmask_pixels():
...
+        ncmed=3,
+        nlmed=3,
+        lsigma=3,
+        hsigma=3,
+        ngood=3,
+    )
+
+    expected = np.zeros(data.shape, dtype=bool)
+    expected[:4, 2] = True
+    assert xp.all(mask == xp.asarray(expected, device=xp_device))
```
mwcraig, line 289: This `expected` is identical for `findbadcolumns=False` and `=True`, so the parametrization adds execution coverage of the ported column-statistics code but no assertion coverage: mutation testing shows that changing the all-masked fill to `.set(99)` and reintroducing the discarded-result form of the `set(0)` fill both pass on numpy *and* jax.

Two data tweaks would make the branch observable (numbers derived from the actual thresholds, with 4-row blocks so `csum_sigma = sqrt(4 - csum)`):

1. Add a column whose *unmasked* residuals sum to about 3.5 — then `3.5 > 3 * sqrt(0.5) ≈ 2.12`, so `colmask` fires, the whole column is masked via the `logical_or`, and the expected masks finally differ between the two parametrizations.
2. Optionally, add a column with a strongly negative sum (about −20) and assert it is **not** masked. Without the `set(0)` clamp, `−20 < −3 * sqrt(24) ≈ −14.7` would fire, so this pins the clamp's behavior directly — the reintroduced-discard mutation would then fail on jax instead of passing.

(The `set(1)` all-masked fill is only observable in degenerate narrow edge blocks — e.g. a 9-column image with `ncsig=4` leaving a 1-wide block where `csum_sigma` becomes 0 — so I think it's fine to leave that one executed-but-unasserted, ideally with a brief comment saying so.)

Comment generated by Claude Code

ccdproc/tests/test_ccdmask.py
```diff
@@ -5,9 +5,37 @@
 from astropy.nddata import CCDData
 from numpy.testing import assert_array_equal
 
+from ccdproc.conftest import testing_array_device as xp_device
+from ccdproc.conftest import testing_array_library as xp
 from ccdproc.core import ccdmask
 
 
+def _ccdmask_in_active_namespace(data, **kwargs):
```
mwcraig, line 15: Explain why this is necessary (inline comments are fine)

mwcraig (changes requested):
Thanks for tackling #931 — the production changes here are exactly right. Independent verification: numpy results are bit-identical before/after this PR across a broad sweep of `ccdmask`/`flat_correct` configurations, jax now matches numpy, and on `main`+jax `flat_correct(min_value=...)` was *silently ignoring* `min_value` (the `xpx.at` result was discarded), so this fixes silent wrong results as well as crashes. The new tests genuinely fail against unfixed code on jax, so the regression value is real.

I am requesting one round of changes, confined to the new `ccdmask` tests: the `findbadcolumns=True` case executes the ported column-statistics code, but its assertions cannot detect a value-level regression in it. Concretely (verified by mutation testing): changing the all-masked fill to `.set(99)` **and** reintroducing the discarded-result form `xpx.at(csum)[csum <= 0].set(0)` — the exact bug class this PR fixes — leaves all three new tests passing on both numpy and jax. The cause is the test data: with the mocked filters, the unmasked residuals are all zero, so `csum == [0, 0, 0, 0]` in every block, the `set(0)` fill maps zeros to zeros, and `colmask` needs `|csum| > 3 * csum_sigma ≈ 6` to fire, which never happens. The tell: the `expected` mask is identical for `findbadcolumns=False` and `=True`.

*Disclosure: these review comments were generated by Claude Code at my suggestion — including the code trace and the mutation testing described above — and reviewed/edited by me before posting.*

ccdproc/tests/test_ccdmask.py
nomad3: Added an inline comment explaining that the helper constructs the CCD in the selected test namespace to exercise functional-update paths and verifies that ccdmask leaves its input unchanged.

ccdproc/tests/test_ccdmask.py
nomad3: Updated the helpers to assert the exact (3, 3) window and the expected 30.9 and 69.1 percentile values.

ccdproc/tests/test_ccdmask.py
nomad3: Strengthened the fixture so the positive residual column is masked only with findbadcolumns=True, while the strongly negative residual column stays unmasked after clamping. I also mutation-checked the discarded clamp result: the JAX findbadcolumns=True case now fails as intended. The all-masked fill remains executed and is explicitly documented as not separately asserted in this non-degenerate geometry.

mwcraig (approved):
Thanks again, @nomad3!

PR #957, opened 2026-07-19
Fix fallback percentile calculation

### LINKED ISSUE #888: Add test for `core._percentile_fallback`
Right no this is marked as `#pragma: no cover` but Ideally we would test it. The [Sparse array library](https://sparse.pydata.org/en/stable/index.html) might be a good option, since it has no `percentile` function or method.

nomad3:
## Summary

- exercise `_percentile_fallback` with both the ordinary NumPy coverage lane and the existing `array-api-strict` backend on a non-default device
- calculate fallback indexes with Array API shape, dtype, percent-scale, and device-aware operations
- support integer and list-like percentiles and keep the 100th percentile inside the valid index range
- keep the existing sort-and-order-statistic behavior without selecting a new percentile interpolation convention

Fixes #888

## Validation

- failing-first strict regression: 2 failures at `len(Array)` before the original fix
- refreshed NumPy regression: 6 passed, with a test-only namespace proxy that omits `percentile` so the production fallback executes in the coverage job
- refreshed strict regression: 6 passed on the non-default device, including scalar floats and integer `[0, 50, 100]` endpoints
- refreshed full NumPy suite: 349 passed, 29 skipped
- the earlier full JAX suite passed with 324 passed, 31 skipped, and 7 existing xfailed
- the earlier full Dask suite passed with 326 passed and 36 skipped
- Ruff, Black, and `git diff --check`

## Checklist

For documentation changes:

For bugfixes:

- [x] Does the commit message include a "Fixes #issue_number"?

For new functionality:

- [ ] Does the commit message include a "Fixes #issue_number"?

## AI assistance

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ Bug Fixes
   average combinations. [#952]
 - Make ``flat_correct`` and ``ccdmask`` use functional array updates so they
   support immutable array-API backends. [#956]
+- Fix the fallback percentile calculation for array namespaces that do not
+  provide ``percentile``. [#957]
 - Fix dtype conversion in ``Combiner._weighted_sum`` to use the array-API
   namespace form ``xp.astype(weights, xp.float64)`` instead of the deprecated
   string-based ``.astype("float64")``. This resolves the ``DeprecationWarning``
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -125,14 +125,17 @@ def _percentile_fallback(array, percentiles, xp=None):
     except AttributeError:
         pass
 
-    # Fall back to using sort
-    sorted_array = xp.sort(array)
-
     percentile_array = xp.asarray(
         percentiles,
         dtype=xp.float64,
-        device=array_api_compat.device(sorted_array),
+        device=array_api_compat.device(array),
     )
+    if bool(xp.any((percentile_array < 0) | (percentile_array > 100))):
+        raise ValueError("Percentiles must be in the range [0, 100]")
+
+    # Fall back to using sort
+    sorted_array = xp.sort(array)
+
     indexes = xp.astype(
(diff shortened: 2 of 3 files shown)


mwcraig (changes requested):
Thanks for the fix — the approach here is right, and I appreciate the fallback finally getting real test coverage on both the numpy and array-api-strict lanes. Verifying against the pre-PR code confirmed the tests genuinely exercise the fixed path.

One change requested before this is ready:

**The index clamp is one-sided.** The new code clamps the high end:

```python
indexes = xp.minimum(
    indexes,
    xp.asarray(sorted_array.shape[0] - 1, ...),
)
```

but there's no low-side clamp or range validation. A negative percentile produces a negative index, which wraps Python-style to the *top* of the sorted array. I verified: with an array of `0..99`, `_percentile_fallback(arr, -5)` returns `95.0`, where `np.percentile` raises `ValueError`. Similarly, percentiles above 100 silently clamp to the maximum instead of raising. So both out-of-range directions silently return large values, and the fallback disagrees with the namespace-`percentile` branch on error behavior — the same call would raise under numpy but return garbage under a fallback namespace.

The current callers hard-code 30.9/69.1 so this is latent today, but since this PR is establishing the fallback's contract, could you either:

1. validate `0 <= p <= 100` up front and raise `ValueError` (preferred — matches `np.percentile`), or
2. at minimum add the low-side clamp (`xp.maximum(indexes, 0)`),

and add an out-of-range case to `test_percentile_fallback`? That's exactly the region the current parametrization doesn't touch.

Everything else looks good — the `shape[0]`/`÷100`/dtype/device handling is exactly what strict-on-non-default-device needs, and keeping floor-rank order statistics rather than picking an interpolation convention is the right call for `ccdmask`'s use.

mwcraig (approved):
Thanks once again @nomad3!

PR #958, opened 2026-07-19
Propagate gain correction to L.A.Cosmic uncertainties

### LINKED ISSUE #729: lacosmic corrects for gain in data array but not uncertainties
Hi, I noticed something strange about the behaviour of lacosmic and I'm not sure if this is a feature or a bug. If I feed it an image in adu (with an uncertainty array in adu) and the correct value for the gain, it corrects the data and updates the data unit to e-, but not the uncertainty array.
Is that how it's supposed to work?
I'm using astropy v3.2.2, ccdproc 2.1.0 and numpy 1.17.2.

nomad3:
## Summary

- propagate L.A.Cosmic gain correction to `CCDData` uncertainties using the existing uncertainty arithmetic
- preserve the reported ADU data and uncertainty units when `gain_apply=False`
- cover standard deviation, variance, and inverse-variance representations without changing cosmic-ray detection or replacement-pixel semantics

This deliberately does not claim support for every equivalent scaled uncertainty unit; that is an existing Array API wrapper boundary outside #729.

Fixes #729

## Validation

- `python -m pytest ccdproc/tests/test_cosmicray.py -q` — 35 passed
- `python -m pytest ccdproc -q` — 352 passed, 29 skipped
- `JAX_ENABLE_X64=1 CCDPROC_ARRAY_LIBRARY=jax python -m pytest ccdproc -q` — 345 passed, 29 skipped, 7 expected failures
- `CCDPROC_ARRAY_LIBRARY=dask python -m pytest ccdproc -q` — 347 passed, 34 skipped
- Black, Ruff, and `git diff --check`

The non-default `array-api-strict` L.A.Cosmic tests remain expected-failed because Astro-SCRAPPY requires NumPy.

## Checklist

For documentation changes:

- [ ] For documentation changes: Does your commit message include a `[skip ci]`?

For bugfixes:

- [x] Did you add an entry to the `CHANGES.rst` file?
- [x] Does the commit message include a `Fixes #issue_number`?

## AI assistance disclosure

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,7 +26,7 @@ Bug Fixes
   public uncertainty types from arithmetic paths. [#953]
 - Propagate gain correction to CCDData uncertainties in
   ``cosmicray_lacosmic``, and preserve data and uncertainty units when gain
-  correction is disabled. [#729]
+  correction is disabled. [#958, #729]
 
 2.5.1 (2025-07-05)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,9 @@ Bug Fixes
   that was being promoted to a test failure in the JAX CI job. [#924]
 - Return plain ``CCDData`` objects from array-API wrapper-copy paths, and
   public uncertainty types from arithmetic paths. [#953]
+- Propagate gain correction to CCDData uncertainties in
+  ``cosmicray_lacosmic``, and preserve data and uncertainty units when gain
+  correction is disabled. [#729]
 
 2.5.1 (2025-07-05)
 ------------------
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -1874,10 +1874,6 @@ def cosmicray_lacosmic(
                 )
                 nccd.uncertainty = gain_corrected.uncertainty
             nccd.unit = _ccd.unit * gain.unit
-            if nccd.uncertainty is not None:
-                nccd.uncertainty.unit = nccd.uncertainty._data_unit_to_uncertainty_unit(
-                    nccd.unit
-                )
 
         nccd.data = xp.asarray(cleanarr)
         if nccd.mask is None:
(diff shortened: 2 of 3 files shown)


mwcraig (changes requested):
Thanks for taking on #729 — routing the gain through the wrapper arithmetic so it propagates into the uncertainty is the right approach, and the `gain_apply=False` unit fix addresses real mislabeling. Two confirmed issues to resolve before this is ready, plus a couple of test notes.

**1. Variance/InverseVariance uncertainty arrays silently escape to numpy on jax.** Only `_StdDevUncertaintyWrapper` has an xp-aware `_propagate_multiply`; the Variance and InverseVariance wrappers fall through to astropy's numpy propagation path. Verified under jax: after `cosmicray_lacosmic` with gain applied, `result.uncertainty.array` is a numpy array for Variance and InverseVariance, while StdDev correctly stays jax. The wrapper gap pre-exists this PR, but this adds the first public call site that exercises it while the PR claims coverage of all three uncertainty types — and the tests can't see it because `assert_allclose` coerces to numpy before comparing. Please either fix the escape (xp-aware `_propagate_multiply` for the other two wrappers) or explicitly track it: add a namespace assertion to the test and mark the jax case xfail (or record it as a known escape) so the gap is visible rather than silently passing.

**2. Dead code — the unit-reassignment block is a no-op.** This line (and the surrounding 4-line block):

```python
nccd.uncertainty.unit = nccd.uncertainty._data_unit_to_uncertainty_unit(nccd.unit)
```

has no effect: `_unwrap_uncertainty` rebuilds the uncertainty from `.array` alone, dropping the unit, and astropy then re-derives it from the parent CCDData. I deleted the whole block and got byte-identical results across every case tested. It also reaches into a private astropy API (`_data_unit_to_uncertainty_unit`) for nothing. Please delete it, along with the redundant second `is not None` guard.

Smaller test points:

- The tests derive their expected units via the same private `_data_unit_to_uncertainty_unit` the code calls — that makes the assertion circular. Please state the expected units literally (e.g. `electron` for StdDev, `electron**2` for Variance).
- The `backend_xfail` reason says "uses astroscrappy", but astroscrappy is monkeypatched out in these tests — please update the reason to the actual limitation.

The numerics, units, and non-mutation coverage are otherwise good — with the escape either fixed or explicitly tracked and the dead block removed, this will be in good shape.

mwcraig (approved):
Thanks @nomad3!

PR #960, opened 2026-07-19
Allow per-call CCD extension overrides

### LINKED ISSUE #817: ImagerFileCollection.ccds() extension override
When using `ImageFileCollection.ccds()` to generate `CCDData` objects from the files in an `ImageFileCollection` it is not possible to change which FITS extension the `CCDData` objects will be generated for.  It is only possible to generate `CCDData` objects for the extension that was selected when the `ImageFileCollection` was created. This is a problem when working with some multi-extension FITS files, such as those from some ESO instruments.

Here's the `hdulist.info()`  from an example FITS file from the ESPRESSO instrument:

    Filename: ESPRESSO/bias_20231203/ESPRE.2023-12-03T12:49:30.096.fits.bz2
    No.    Name      Ver    Type      Cards   Dimensions   Format
      0  PRIMARY       1 PrimaryHDU    1440   (0, 0)      
      1  CCD290blue    1 ImageHDU       720   (9920, 9296)   int16 (rescales to uint16)   
      2  CCD290red     1 ImageHDU       720   (9920, 9296)   int16 (rescales to uint16)  

The primary only contains a header, no data, and there are two image data extensions corresponding to the blue and red arms of the spectograph. The primary header contains most of the information you would want to use for selecting/filtering files (e.g. image type, target object, etc.) so you need to create the `ImageFileCollection` with `ext=0`, but having selected the FITS files that you want to process you can't tell `ImageFileCollection.ccds()` to use the data extensions instead.  `ImageFileCollection.ccds()`  doesn't have an `ext` argument, and if you try to specify it using the `ccd_kwargs` argument instead you get a `TypeError` due to multiple keyword values.

Example:
```
 # Use keyword in primary header to select bias frames
ifc = ccdproc.ImageFileCollection(data_path, keywords="*", ext=0)
bias_ifc = ifc.filter(**{'eso dpr type':'bias'})  

# Now try to get CCDDatas for blue channel data in extension 1 for combining.
# This raises TypeError:
blue_combiner = ccdproc.Combiner(bias_ifc_dec.ccds(ext='CCD290blue'))
# This raises TypeError too: 
blue_combiner = ccdproc.Combiner(bias_ifc_dec.ccds(ccd_kwargs={'hdu':'CCD290blue'}))
```

Of course it's possible to work around this by writing your own `CCDData` generator and feeding the `ImageFileCollection.files` into it, but it would be more convenient if `ImageFileCollection.ccds()` had an optional `ext` argument that allowed you to override the value that gets passes to the `astropy.nddata.ccddata.fits_ccddata_reader()` `hdu` parameter.

* Python 3.12.2
* astropy 6.0.0
* ccdproc 2.4.1
* numpy 1.26.4

nomad3:
## Summary

- add an optional `ext` argument to `ImageFileCollection.ccds()` for selecting a FITS data extension per call
- keep the collection's own extension unchanged for primary-header summaries and filtering
- continue supporting `ccd_kwargs["hdu"]`, avoid mutating caller dictionaries, and reject contradictory selectors explicitly
- cover integer, extension-name, `(name, version)`, and zero-valued selectors with offline multi-extension FITS regressions
- correct the generated NumPy-style parameter indentation so the `ccds()` docstring builds cleanly under Sphinx warnings-as-errors

Compatibility boundary: `ext` is now reserved as the per-call selector; code that needs to filter an `EXT` header keyword should filter the collection before calling `ccds()`.

Fixes #817

## Validation

- initial override/conflict regression on untouched `main`: 7 failed
- `python -m pytest ccdproc/tests/test_image_collection.py -q` — 85 passed
- full NumPy suite — 354 passed, 29 skipped
- full Dask suite — 349 passed, 34 skipped
- full JAX x64 suite — 347 passed, 29 skipped, 7 established xfails
- inspected the final cleaned docstring: `ext` is a parameter and every description line is indented beneath it
- Black, Ruff, `git diff --check`, and `git show --check`

## AI assistance disclosure

OpenAI Codex was used for repository inspection, test design, implementation, CI diagnosis, and automated review. The contributor has completed the final review and takes responsibility for the contribution's correctness and maintenance.

## Checklist

- [ ] For documentation changes: Does the commit message include a `[skip ci]`? (Not a documentation-only change.)

For new functionality:

- [x] Did you add an entry to `CHANGES.rst`?
- [x] Did you include a meaningful docstring with Parameters?
- [x] Does the commit message include `Fixes #817`?
- [x] Does this PR add, rename, move, or remove an existing function or parameter? (It adds the optional `ext` parameter.)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,8 +10,9 @@ New Features
   ``backend_skip`` markers, an escape-site report
   (``CCDPROC_TRIAGE_ESCAPES=1``), and an escape logger
   (``CCDPROC_LOG_ARRAY_ESCAPES=1``). [#942]
-- Allow ``ImageFileCollection.ccds`` to read a per-call FITS extension while
-  retaining the collection extension for summaries and filtering. [#960]
+- Allow ``ImageFileCollection.ccds`` to override the collection's FITS
+  extension per call with ``ccd_kwargs["hdu"]`` while preserving ``ext=`` as
+  a header filter. [#960]
 
 Bug Fixes
 ^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,7 @@ New Features
   (``CCDPROC_TRIAGE_ESCAPES=1``), and an escape logger
   (``CCDPROC_LOG_ARRAY_ESCAPES=1``). [#942]
 - Allow ``ImageFileCollection.ccds`` to read a per-call FITS extension while
-  retaining the collection extension for summaries and filtering. [#817]
+  retaining the collection extension for summaries and filtering. [#960]
 
 Bug Fixes
 ^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ New Features
   ``backend_skip`` markers, an escape-site report
   (``CCDPROC_TRIAGE_ESCAPES=1``), and an escape logger
   (``CCDPROC_LOG_ARRAY_ESCAPES=1``). [#942]
+- Allow ``ImageFileCollection.ccds`` to read a per-call FITS extension while
+  retaining the collection extension for summaries and filtering. [#817]
 
 Bug Fixes
 ^^^^^^^^^
--- a/ccdproc/image_collection.py
+++ b/ccdproc/image_collection.py
@@ -897,6 +897,7 @@ class ImageFileCollection:
             See `~astropy.nddata.fits_ccddata_reader` for a complete list of
             parameters that can be passed through ``ccd_kwargs``.
 
+        {ccd_extension}
 
         regex_match : bool, keyword-only
             If ``True``, then string values in the ``**kwd`` dictionary are
@@ -933,7 +934,8 @@ class ImageFileCollection:
         if kwd:
             self._find_keywords_by_values(**kwd)
 
-        ccd_kwargs = ccd_kwargs or {}
+        ccd_kwargs = dict(ccd_kwargs or {})
+        ccd_hdu = ccd_kwargs.pop("hdu", self.ext)
 
         for full_path in self._paths():
             add_kwargs = {"do_not_scale_image_data": do_not_scale_image_data}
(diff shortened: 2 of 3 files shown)


mwcraig (changes requested):
Thanks for picking up #817 — per-call extension selection is a real gap, and the mechanics here are sound: the `_generator` change (`ccd_hdu = ccd_kwargs.pop("hdu", self.ext)` on a copy, so the caller's dict isn't mutated) is exactly right, and the tests for the selector paths are thorough.

However, the `ext=` parameter on `ccds()` introduces a silent breaking change, and after discussion we'd like the PR reshaped to avoid it.

**The problem:** `ccds()`'s `**kwd` namespace is a header-keyword *filter* — on current main, `ccds(ext=1)` means "the files whose `EXT` header card equals 1". This PR intercepts `ext` unconditionally, so that same call now silently yields **HDU 1 of every file in the collection** — wrong data, no warning. String values that used to filter (`ccds(ext='SCI')`) now raise `KeyError` instead. And the sibling methods (`hdus()`, `headers()`, `data()`, `files_filtered()`) still treat `ext=` as a filter, so one keyword would mean opposite things across methods that share a docstring template promising that any `**kwd` keyword filters.

**Requested change — drop the `ext=` parameter and make `ccd_kwargs={"hdu": ...}` the supported route.** The nice thing is that your `_generator` fix already does all the real work: with it, `ccds(ccd_kwargs={"hdu": 1})` works today, overrides the collection-level default, and can't collide with filtering because `ccd_kwargs` was never part of the filter namespace. `hdu` also matches the underlying `fits_ccddata_reader(hdu=...)` keyword, so it's the honest name. Concretely:

1. Remove the `ext=None` parameter and the `ext`/`ccd_kwargs['hdu']` conflict check from `ccds()`.
2. Keep the `_generator` change as-is — it's the actual fix.
3. Re-point the tests at the `hdu` route (your parametrization already covers it — it's mostly deleting the `ext` variants and the conflict-error test).
4. Update the docstring addition to document `ccd_kwargs["hdu"]` (int, name string, or `(name, ver)` tuple) as the way to override the collection's extension per call.
5. Reword the `CHANGES.rst` entry to match.

That keeps everything #817 asked for while leaving the filter contract of `**kwd` untouched. Sorry to ask for a restructure after you've built the `ext=` plumbing — the selector logic and tests all carry over, it's really just moving the entry point.

mwcraig (approved):
Thanks @nomad3  -- I really appreciate all of the PRs!

PR #976, opened 2026-08-22
Build Combiner data and mask arrays with xp.stack

### LINKED ISSUE #965: Combiner.__init__ builds a nested Array via xp.asarray([ccd.data ...]); array-api-strict requires xp.stack (44 failures)
Under `CCDPROC_ARRAY_LIBRARY=array-api-strict` (triage tooling from #937/#939, branch `array-api-strict-triage-tooling`), 44 of the 63 failures in `ccdproc/tests/test_combiner.py` (63 failed, 11 passed total) trace to a single line in `Combiner.__init__`. This is the biggest single failure source in the #941 escape-site table.

**Offending code** — `ccdproc/combiner.py:189`:

```python
# set up the data array
self._data_arr = xp.asarray([ccd.data for ccd in ccd_list], dtype=dtype)
```

**Root cause**: each `ccd.data` is already an `array_api_strict.Array`. Passing a Python list of `Array` objects to `xp.asarray` triggers array-api-strict's explicit guard against nested arrays — it refuses to infer a stacking axis and requires `xp.stack` instead.

**Representative traceback**:

```
ccdproc/tests/test_combiner.py:87: in test_combiner_create
    c = Combiner(ccd_list)
ccdproc/combiner.py:189: in __init__
    self._data_arr = xp.asarray([ccd.data for ccd in ccd_list], dtype=dtype)
array_api_strict/_creation_functions.py:101: in asarray
    raise TypeError("Nested Arrays are not allowed. Use `stack` instead.")
E   TypeError: Nested Arrays are not allowed. Use `stack` instead.
```

**Suggested fix**: replace the list-comprehension + `asarray` with `xp.stack([ccd.data for ccd in ccd_list], axis=0)` (casting dtype separately via `xp.astype` if needed). The mask construction at `combiner.py:196` (`self._data_arr_mask = xp.asarray(mask_list, dtype=bool)`) has the identical pattern and will hit the same error once the data-array construction is fixed, so it should be converted at the same time.

This is likely also the biggest CuPy failure source, since CuPy hits the same nested-array conversion path.

Part of the array-API triage in #941; found with the array-api-strict tooling from #937/#939. Verified at branch tip d329ac5 (2026-08-01): failure count and line number confirmed by re-running the suite and grepping tracebacks.

mwcraig:
array-api-strict rejects a nested list of arrays passed to `xp.asarray`, so every `Combiner` construction failed under the strict backend (50 of the 75 `test_combiner.py` failures). This PR:

- builds `_data_arr` with `xp.stack([...])` + `xp.astype(..., dtype)`;
- builds `_data_arr_mask` by coercing each mask with `xp.asarray(mask, dtype=xp.bool, device=<data device>)` (astropy keeps `CCDData.mask` as numpy even when the data is not) and a `dtype=xp.bool, device=...` placeholder for missing masks, then `xp.stack`;
- replaces the remaining `dtype=bool` with `dtype=xp.bool`;
- puts scaling arrays on the device of the data, reshapes them for broadcasting with a single `xp.reshape`, and indexes individual images with a trailing ellipsis (strict rejects both repeated `[:, newaxis]` on an N-D array and flat `arr[i]` indexing);
- adds a regression test that builds a `Combiner` from data on the testing device with a native-namespace mask, no mask, and a numpy mask, and checks the namespace, device, dtype, and values of `.data`, `.mask`, and `.scaling`.

Test matrix (local, `ccdp-dev-py312`):

| backend | result |
|---|---|
| numpy | 381 passed, 5 skipped |
| jax | 367 passed, 10 skipped, 2 xfailed, 7 xpassed (all pre-existing XPASSes on main) |
| dask | 371 passed, 15 skipped |
| dask + `CCDPROC_ENFORCE_ESCAPE_BASELINE=1` | 371 passed, 15 skipped; no escapes outside the baseline |
| array-api-strict, `test_combiner.py` | before: 75 failed, 12 passed; after: 68 failed, 19 passed, 1 skipped |

The remaining strict failures in `test_combiner.py` are unrelated to this issue: 35 `No NaN-aware ... function available` (bottleneck fallbacks), 6 `dtype=bool` in the tests themselves, 6 `IrreducibleUnit` passed to `xp.asarray`, 5 `numpy.ndarray has no attribute _array`, 3 flat-indexing `IndexError`s in test code, 3 `Fancy indexing __setitem__`, and a handful of others.

Fixes #965

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -179,10 +179,14 @@ class Combiner:
         # new_shape = (len(ccd_list),) + default_shape
         # Stack the individual images rather than passing a nested list to
         # xp.asarray: the array API does not allow nested sequences of arrays.
-        device = array_api_compat.device(ccd_list[0].data)
-        self._data_arr = xp.astype(
-            xp.stack([xp.asarray(ccd.data, device=device) for ccd in ccd_list]),
-            dtype,
+        # Keep the stack on the device of the input data, but only when the
+        # data already belong to ``xp``; a device object from a different
+        # namespace (e.g. numpy's 'cpu' for a jax namespace) is meaningless
+        # to ``xp``, so let ``xp`` use its default device instead.
+        data_xp = array_api_compat.array_namespace(ccd_list[0].data)
+        device = array_api_compat.device(ccd_list[0].data) if data_xp is xp else None
+        self._data_arr = xp.stack(
+            [xp.asarray(ccd.data, dtype=dtype, device=device) for ccd in ccd_list]
         )
 
         # populate self._data_arr_mask. The mask of a CCDData may be a numpy
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -282,8 +282,16 @@ class Combiner:
             n_images = self._data_arr.shape[0]
             device = array_api_compat.device(self._data_arr)
             if callable(value):
-                self._scaling = [value(self._data_arr[i, ...]) for i in range(n_images)]
-                self._scaling = xp.asarray(self._scaling, device=device)
+                # The callable may return a Python float or a 0-d array of
+                # the backend; stack per-element conversions rather than
+                # passing a list of arrays to asarray, which array-api-strict
+                # rejects as a nested sequence of arrays.
+                self._scaling = xp.stack(
+                    [
+                        xp.asarray(value(self._data_arr[i, ...]), device=device)
+                        for i in range(n_images)
+                    ]
+                )
             else:
                 try:
                     len(value)
@@ -1123,8 +1131,11 @@ def combine(
 
                 scalevalues.append(scale(imgccd.data))
 
-            to_set_in_combiner["scaling"] = xp.asarray(
-                scalevalues, device=array_api_compat.device(ccd.data)
+            # See Combiner.scaling: stack per-element conversions so that a
+            # callable returning 0-d backend arrays works on array-api-strict.
+            device = array_api_compat.device(ccd.data)
+            to_set_in_combiner["scaling"] = xp.stack(
+                [xp.asarray(value, device=device) for value in scalevalues]
             )
         else:
             to_set_in_combiner["scaling"] = scale
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -105,7 +105,8 @@ class Combiner:
     xp : array namespace, optional
         The array namespace to use for the data. If `None` or not provided, it will
         be inferred from the first `~astropy.nddata.CCDData` object in
-        ``ccd_iter``.
+        ``ccd_iter``. A plain module (e.g. ``numpy``) is accepted and is
+        converted to its array-API-compatible namespace.
         Default is `None`.
 
     Raises
@@ -165,8 +166,14 @@ class Combiner:
                 if not (default_unit == ccd.unit):
                     raise TypeError("CCDData objects don't have the same unit.")
 
-        # Set array namespace
-        xp = xp or array_api_compat.array_namespace(ccd_list[0].data)
+        # Set array namespace. A raw module such as ``numpy`` or ``dask.array``
+        # may lack array-API features that are used below (``xp.bool``, the
+        # ``device`` keyword), so normalise whatever the caller passed to the
+        # array-api-compat namespace of one of its arrays.
+        if xp is None:
+            xp = array_api_compat.array_namespace(ccd_list[0].data)
+        else:
+            xp = array_api_compat.array_namespace(xp.asarray(0))
         self._xp = xp
         if dtype is None:
             dtype = xp.float64
@@ -281,31 +288,41 @@ class Combiner:
         else:
             n_images = self._data_arr.shape[0]
             device = array_api_compat.device(self._data_arr)
+            dtype = self._data_arr.dtype
             if callable(value):
                 # The callable may return a Python float or a 0-d array of
                 # the backend; stack per-element conversions rather than
                 # passing a list of arrays to asarray, which array-api-strict
                 # rejects as a nested sequence of arrays.
+                # Cast to the data dtype so that scaling by, e.g., an integer
+                # does not require type promotion, which the array API does
+                # not guarantee between integer and floating dtypes.
                 self._scaling = xp.stack(
                     [
-                        xp.asarray(value(self._data_arr[i, ...]), device=device)
+                        xp.asarray(
+                            value(self._data_arr[i, ...]), dtype=dtype, device=device
+                        )
                         for i in range(n_images)
                     ]
                 )
             else:
+                # Array API arrays need not implement __len__, so use the
+                # shape where there is one and fall back to len() for lists
+                # and tuples.
                 try:
-                    len(value)
-                except TypeError as err:
+                    n_values = getattr(value, "shape", None)
+                    n_values = n_values[0] if n_values else len(value)
+                except (TypeError, IndexError) as err:
                     raise TypeError(
                         "scaling must be a function or an array "
                         "the same length as the number of images.",
                     ) from err
-                if len(value) != n_images:
+                if n_values != n_images:
                     raise ValueError(
                         "scaling must be a function or an array "
                         "the same length as the number of images."
                     )
-                self._scaling = xp.asarray(value, device=device)
+                self._scaling = xp.asarray(value, dtype=dtype, device=device)
             # reshape so that broadcasting occurs properly
             self._scaling = xp.reshape(
                 self._scaling, (n_images,) + (1,) * (self._data_arr.ndim - 1)
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -177,14 +177,26 @@ class Combiner:
 
         # set up the data array
         # new_shape = (len(ccd_list),) + default_shape
-        self._data_arr = xp.asarray([ccd.data for ccd in ccd_list], dtype=dtype)
+        # Stack the individual images rather than passing a nested list to
+        # xp.asarray: the array API does not allow nested sequences of arrays.
+        device = array_api_compat.device(ccd_list[0].data)
+        self._data_arr = xp.astype(
+            xp.stack([xp.asarray(ccd.data, device=device) for ccd in ccd_list]),
+            dtype,
+        )
 
-        # populate self._data_arr_mask
+        # populate self._data_arr_mask. The mask of a CCDData may be a numpy
+        # array even when its data is not, so coerce each mask into the data
+        # namespace and onto the data device before stacking.
         mask_list = [
-            ccd.mask if ccd.mask is not None else xp.zeros(default_shape)
+            (
+                xp.asarray(ccd.mask, dtype=xp.bool, device=device)
+                if ccd.mask is not None
+                else xp.zeros(default_shape, dtype=xp.bool, device=device)
+            )
             for ccd in ccd_list
         ]
-        self._data_arr_mask = xp.asarray(mask_list, dtype=bool)
+        self._data_arr_mask = xp.stack(mask_list)
 
         # Must be after self.data_arr is defined because it checks the
         # length of the data array.
@@ -264,9 +276,10 @@ class Combiner:
             self._scaling = value
         else:
             n_images = self._data_arr.shape[0]
+            device = array_api_compat.device(self._data_arr)
             if callable(value):
-                self._scaling = [value(self._data_arr[i]) for i in range(n_images)]
-                self._scaling = xp.asarray(self._scaling)
+                self._scaling = [value(self._data_arr[i, ...]) for i in range(n_images)]
+                self._scaling = xp.asarray(self._scaling, device=device)
             else:
                 try:
                     len(value)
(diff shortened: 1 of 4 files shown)


ccdproc/combiner.py
```diff
@@ -177,14 +177,26 @@ def __init__(self, ccd_iter, dtype=None, xp=None):
...
         # set up the data array
         # new_shape = (len(ccd_list),) + default_shape
-        self._data_arr = xp.asarray([ccd.data for ccd in ccd_list], dtype=dtype)
+        # Stack the individual images rather than passing a nested list to
+        # xp.asarray: the array API does not allow nested sequences of arrays.
+        device = array_api_compat.device(ccd_list[0].data)
+        self._data_arr = xp.astype(
+            xp.stack([xp.asarray(ccd.data, device=device) for ccd in ccd_list]),
+            dtype,
+        )
```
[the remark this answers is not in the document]
mwcraig: Both points are correct — thanks.

**Device:** verified. With numpy `CCDData` inputs and an explicit foreign namespace, `main` works but this branch regresses:

```python
Combiner([CCDData(np.ones((3, 3)), unit=u.adu)] * 2, xp=jax.numpy)
# main:   OK
# branch: TypeError (jax rejects device='cpu' from array_api_compat.device(numpy_array))
# with xp=array_api_strict: ValueError: Unsupported device 'cpu'
```

The device should only be pinned when it actually comes from `xp`'s own arrays; otherwise let `xp.asarray` place the data on its default device:

```python
data_xp = array_api_compat.array_namespace(ccd_list[0].data)
device = array_api_compat.device(ccd_list[0].data) if data_xp is xp else None
```

(`xp` is already set from `array_namespace(...)` when not given, so the `is` comparison holds in the inferred case, and `device=None` is spec-compliant in `asarray`/`zeros`.)

**Extra cast:** also correct — `xp.astype` copies even when the dtype already matches, so it's a second full-stack copy. Passing `dtype` per element avoids it and is valid under array-api-strict on a non-default device:

```python
self._data_arr = xp.stack(
    [xp.asarray(ccd.data, dtype=dtype, device=device) for ccd in ccd_list]
)
```

Will push both, plus a regression test for the explicit-`xp` case, to this branch.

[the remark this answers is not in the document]
mwcraig: Pushed 2cf8b2f with both changes and a regression test (`test_combiner_explicit_namespace_differs_from_data`: numpy inputs with an explicit `xp`), which fails on the previous commit under jax and passes on numpy/jax/dask/array-api-strict now. Full suites unchanged otherwise.

mwcraig:
Re Copilot's two suppressed comments (`Combiner.scaling` setter and the callable-`scale` branch of `combine()`): confirmed and fixed in b67e5fe. Under array-api-strict, a scaling callable returning a 0-d backend array (e.g. `xp.mean(img)`) made `xp.asarray([...])` raise the same "Nested Arrays are not allowed. Use `stack` instead." error as #965. Both sites now do `xp.stack([xp.asarray(v, device=device) for v in values])`, which handles both 0-d arrays and plain Python floats (`xp.stack` alone would break the float case the existing tests use). Added a strict-exercised check to `test_combiner_stacks_arrays_on_input_device` and a new `test_combine_scale_callable_returning_backend_scalar` (xfail on strict only because `combine()` still calls `.nbytes`, which is pre-existing and out of scope). Verified under numpy, dask, strict, and the dask escape ratchet.

mwcraig (commented):
Critical review of the diff (numpy-regression, array-API-semantics, and test-coverage passes, every claim re-run locally).

**Verdict: sound.** numpy/dask behaviour is bit-identical to `main` across ~70 combinations (int/float/mixed dtypes, numpy/int/float/`MaskedArray` masks, missing masks, `scaling` as list/tuple/ndarray/0-d/callable, 1-D and 4-D images, `combine()` with `mem_limit` chunking × scale × method). Peak memory is *lower* than `main` (10×1000² float64→float64: 170 MB → 100 MB, since `asarray(list)` built an intermediate). The device-pinning commit holds on strict (`device1`) and jax for every mixed-device / mixed-namespace ordering tried. jax suite has the same 6 pre-existing XPASSes as `main`; all 62 remaining strict failures in `test_combiner.py` are outside this diff (30 bottleneck-missing, 6 `dtype=bool` in the test wrapper, `.nbytes`, fancy-index `__setitem__`, …); no nested-`asarray` remains anywhere.

Inline: two small strict-only bugs the PR leaves reachable in the scaling setter (int scaling; backend array as `scaling`), one policy question about raw modules as `xp`, one CHANGES note, and two honesty notes on what the new tests actually guard.


ccdproc/combiner.py
```diff
@@ -264,9 +280,18 @@ def scaling(self, value):
...
             if callable(value):
-                self._scaling = [value(self._data_arr[i]) for i in range(n_images)]
-                self._scaling = xp.asarray(self._scaling)
+                # The callable may return a Python float or a 0-d array of
+                # the backend; stack per-element conversions rather than
+                # passing a list of arrays to asarray, which array-api-strict
+                # rejects as a nested sequence of arrays.
+                self._scaling = xp.stack(
+                    [
+                        xp.asarray(value(self._data_arr[i, ...]), device=device)
```
mwcraig: **Integer scaling breaks on array-api-strict.** The setter never casts to the data dtype, so `scaling=[1, 2]`, or a callable returning a Python `int`, yields an `int64` array and `self._data_arr * self.scaling` (line 491) raises

```
TypeError: array_api_strict.float64 and array_api_strict.int64 cannot be type promoted together
```

numpy/dask/jax promote silently so nobody sees it today, but it is exactly the class of bug the strict job exists for, and this PR is what makes the scaling path reachable there. One token fixes it here and on line 308:

```python
xp.asarray(value(self._data_arr[i, ...]), dtype=self._data_arr.dtype, device=device)
```

That also covers a callable that returns an `int` for some images and a `float` for others — `xp.stack` of mixed 0-d dtypes fails on strict too. Suggest adding `c.scaling = [1, 2]` and `c.scaling = lambda arr: 1` to `test_combiner_stacks_arrays_on_input_device`.

ccdproc/combiner.py
```diff
@@ -264,9 +280,18 @@ def scaling(self, value):
...
+                        for i in range(n_images)
+                    ]
+                )
             else:
                 try:
                     len(value)
```
mwcraig: **A backend array as `scaling` gives a misleading error on strict.** Standard arrays need not implement `__len__` (array-api-strict's doesn't), so

```python
c.scaling = xp.asarray([1.0, 0.5], device=device)
```

is caught by this `try/except` and re-raised as *"scaling must be a function or an array the same length as the number of images"* even though the length is right. Pre-existing, but newly reachable. `value.shape[0]` (falling back to `len()` for lists/tuples) is standard-compliant.

ccdproc/combiner.py
```diff
@@ -177,14 +177,30 @@ def __init__(self, ccd_iter, dtype=None, xp=None):
...
+        # Stack the individual images rather than passing a nested list to
+        # xp.asarray: the array API does not allow nested sequences of arrays.
+        # Keep the stack on the device of the input data, but only when the
+        # data already belong to ``xp``; a device object from a different
+        # namespace (e.g. numpy's 'cpu' for a jax namespace) is meaningless
+        # to ``xp``, so let ``xp`` use its default device instead.
+        data_xp = array_api_compat.array_namespace(ccd_list[0].data)
+        device = array_api_compat.device(ccd_list[0].data) if data_xp is xp else None
+        self._data_arr = xp.stack(
+            [xp.asarray(ccd.data, dtype=dtype, device=device) for ccd in ccd_list]
```
mwcraig, line 196: **Behaviour change for a raw module passed as `xp`.** `Combiner(ccds, xp=dask.array)` constructed fine on `main`; now `da.asarray(..., device=...)` raises `TypeError: from_array() got an unexpected keyword argument 'device'`. Same for raw `numpy` 1.26 (`xp.bool`, `device=`).

Mitigating: only the bare modules are affected (the `array_api_compat` wrappers work, `combine()` already normalises `array_package`), and on `main` the same raw-dask object died one call later in `clip_extrema` (`module 'dask.array' has no attribute 'argsort'`), while `core.py` on `main` already passes `device=` in five places. So this was never really supported — but it's worth deciding as policy. Cheapest hardening, mirroring what `combine()` does:

```python
xp = array_api_compat.array_namespace(xp.asarray(0)) if xp is not None else ...
```

or just say in the docstring that `xp` must be an array-API-compatible namespace.

ccdproc/combiner.py
```diff
@@ -280,10 +305,11 @@ def scaling(self, value):
                         "scaling must be a function or an array "
                         "the same length as the number of images."
                     )
-                self._scaling = xp.asarray(value)
+                self._scaling = xp.asarray(value, device=device)
             # reshape so that broadcasting occurs properly
-            for _ in range(len(self._data_arr.shape) - 1):
-                self._scaling = self.scaling[:, xp.newaxis]
+            self._scaling = xp.reshape(
```
mwcraig, line 327: Minor, worth a line in CHANGES: this replaces `main`'s "append one axis per data dim" loop, so a `(3, 1)` scaling array or a callable returning a 1-element array now yields `(3, 1, 1)` instead of `(3, 1, 1, 1)`, and `average_combine()` on `(3, 3)` images returns `(3, 3)` instead of `(3, 3, 3)`. That is a fix, but it is user-visible.

ccdproc/tests/test_combiner.py
```diff
@@ -165,6 +167,87 @@ def test_combiner_mask():
...
+    assert float(xp.max(xp.abs(c.scaling - (1.0 + 1 / 16)))) < 1e-6
+
+
+@pytest.mark.backend_xfail(
+    "array-api-strict",
+    reason="combine() sizes the image with .nbytes, which array-api-strict "
+    "does not provide",
+)
+def test_combine_scale_callable_returning_backend_scalar():
+    # Regression test for the review of #976: ``combine(scale=<callable>)``
```
mwcraig: Honesty note on "regression test": this currently cannot regress on any backend. On numpy/dask the pre-fix `xp.asarray(scalevalues)` already passed; on strict, `combine()` dies earlier at `ccd.data.nbytes` (`combiner.py:829`), which is why it is xfailed. The xfail reason is accurate, but until `.nbytes` is dealt with, the `combine()` half of the scaling fix (lines 1134–1139) is exercised by nothing. Suggest rewording the comment to say so.

ccdproc/tests/test_combiner.py
```diff
@@ -165,6 +167,87 @@ def test_combiner_mask():
...
+    ccds = [
+        CCDData(xp.full((4, 4), float(i), dtype=xp.float64), unit=u.adu)
+        for i in (1, 2, 4)
+    ]
+    result = combine(ccds, method="average", scale=lambda arr: 1 / xp.mean(arr))
+    assert_allclose(np.asarray(result.data), 1.0)
+
+
+def test_combiner_explicit_namespace_differs_from_data():
+    # Regression test for the review of #976: when the caller passes an ``xp``
```
mwcraig, line 255: This test guards the device logic **only** on array-api-strict. Traced per backend: on numpy, `array_namespace(np_arr) is xp` is true so the "different namespace" branch never runs; on dask, an unconditional `da.asarray(np_arr, device='cpu')` is accepted anyway so the old bug is undetectable; on strict it does catch it (`ValueError: Unsupported device 'cpu'`) — but on `main`-strict it fails at the unrelated `dtype=bool` mask line first. Not wrong, just weaker than the comment suggests, and the strict job is currently `continue-on-error`. Worth either noting in the comment or adding an assertion on the device that fails on numpy too (e.g. pass `xp=array_api_strict` explicitly in this test).

ccdproc/combiner.py
mwcraig: Do the hardening

ccdproc/tests/test_combiner.py
mwcraig: update the comment

ccdproc/combiner.py
mwcraig: Fixed in 1570989: both the callable and array branches now pass `dtype=self._data_arr.dtype`. Added `c.scaling = [1, 2, 3]` and `c.scaling = lambda arr: arr.shape[0]` to `test_combiner_stacks_arrays_on_input_device` with dtype assertions; passes on array-api-strict.

ccdproc/combiner.py
mwcraig: Fixed in 1570989: the length check now uses `value.shape[0]` when there is a `shape`, falling back to `len()`. Test adds a strict backend array as `scaling`, plus the wrong-length `ValueError` and scalar `TypeError` cases.

ccdproc/combiner.py
mwcraig: Done in 1570989: an explicit `xp` is normalised with `array_api_compat.array_namespace(xp.asarray(0))`, the docstring says a plain module is accepted, and `test_combiner_accepts_raw_module_as_namespace` passes `xp=np` and checks `_xp` is the compat namespace.

ccdproc/combiner.py
mwcraig: Added to the CHANGES entry in 1570989.

ccdproc/tests/test_combiner.py
mwcraig: Comment reworded in 1570989 to say the test cannot yet fail because of the scaling path on any backend, and why.

ccdproc/tests/test_combiner.py
mwcraig: Comment updated in 1570989 to state that only the array-api-strict job can fail this test, with the per-backend reasons.

PR #977, opened 2026-08-22
Fix background_deviation_box discarding the xpx.at result on immutable backends

### LINKED ISSUE #963: background_deviation_box discards the xpx.at result — silent no-op on immutable backends
**Behavior**

On the jax backend, `background_deviation_box` returns a flat array (every pixel = the global `xp.std(data)`) instead of per-box sigmas — silently wrong results, no error (verified empirically 2026-07-27). On numpy it works, but only by side effect.

**Root cause**

The `xpx.at(...).set(...)` result is never assigned:

https://github.com/astropy/ccdproc/blob/736555defed18e88ae0f107e64957be858ea1a1b/ccdproc/core.py#L1341-L1346

On numpy, `xpx.at` takes its in-place fast path so the discarded return value doesn't matter; on immutable backends (jax) it returns a new array that is thrown away, leaving `barr` at its initialized value. This is the same discarded-update bug class fixed for `flat_correct`/`ccdmask` in #956, and (as far as I can find) the last live instance on main.

**Note on #931**

The "Suggested fix" in #931 cites this exact line as the *correct-pattern exemplar* to copy — please don't: the pattern is only correct with the result rebound. I'll drop a note on #931 pointing here.

**Suggested fix**

One-line rebind:

```python
barr = xpx.at(barr)[y1:y2, x1:x2].set(float(sigma_func(data[y1:y2, x1:x2])))
```

plus a jax-lane regression test asserting per-box values differ from the global std (the numpy lane can't catch this — the in-place fast path masks it).

Follow-up to #909 / #956. *This issue was written by Claude (via Claude Code); Matt reviewed and approved its content before filing.*

mwcraig:
`background_deviation_box` called `xpx.at(barr)[y1:y2, x1:x2].set(...)` without rebinding the result. `xpx.at(...).set()` returns a new array, so on immutable backends (JAX, array-api-strict) every box was silently left at the global standard deviation. This rebinds the result.

The existing `test_background_deviation_box` only checked the mean of the output, which happens to pass on JAX even when the per-box values are never written. A new regression test (`test_background_deviation_box_per_box_values`) uses different noise scales in different boxes and checks the per-box values; it fails on JAX before the fix and passes after. The existing test now uses `xp.mean()` so it also runs on array-api-strict.

Test matrix (full suite unless noted):

| backend | result |
| --- | --- |
| numpy | 381 passed, 5 skipped |
| jax | 367 passed, 10 skipped, 2 xfailed, 7 xpassed (all 7 XPASS are pre-existing combiner / image_collection entries) |
| dask | 371 passed, 15 skipped |
| array-api-strict, `test_cosmicray.py` only | before: 1 failed, 7 passed, 33 xfailed; after: 9 passed, 33 xfailed |

Fixes #963

### DIFF
--- a/ccdproc/tests/test_cosmicray.py
+++ b/ccdproc/tests/test_cosmicray.py
@@ -2,6 +2,7 @@
 
 import array_api_compat
 import array_api_extra as xpx
+import numpy as np
 import pytest
 from astropy import units as u
 from astropy.nddata import (
@@ -518,7 +519,30 @@ def test_background_deviation_box():
     scale = 5.3
     cd = xp.asarray(default_rng(seed=123).normal(loc=0, size=(100, 100), scale=scale))
     bd = background_deviation_box(cd, 25)
-    assert abs(bd.mean() - scale) < 0.10
+    assert abs(float(xp.mean(bd)) - scale) < 0.10
+
+
+def test_background_deviation_box_per_box_values():
+    # Regression test for #963: the per-box deviation must actually be
+    # written into the result. A pure-mean check passes even if every box is
+    # silently left at the global standard deviation.
+    rng = default_rng(seed=123)
+    left = rng.normal(loc=0, size=(100, 50), scale=1.0)
+    right = rng.normal(loc=0, size=(100, 50), scale=10.0)
+    cd = xp.asarray(np.hstack([left, right]))
+    bd = background_deviation_box(cd, 50)
+    global_std = float(xp.std(cd))
+    left_val = float(bd[25, 25])
+    right_val = float(bd[25, 75])
+    assert abs(left_val - 1.0) < 0.1
+    assert abs(right_val - 10.0) < 1.0
+    assert left_val != right_val
+    assert abs(left_val - global_std) > 1.0
+    assert abs(right_val - global_std) > 1.0
+    # every pixel within a box shares that box's value (the top-left and
+    # top-right boxes cover rows 0-50 and columns 0-50 / 50-99)
+    assert bool(xp.all(bd[:50, :50] == bd[0, 0]))
+    assert bool(xp.all(bd[:50, 50:99] == bd[0, 50]))
 
 
 def test_background_deviation_box_fail():
--- a/ccdproc/tests/test_cosmicray.py
+++ b/ccdproc/tests/test_cosmicray.py
@@ -534,13 +534,19 @@ def test_background_deviation_box_per_box_values():
     global_std = float(xp.std(cd))
     left_val = float(bd[25, 25])
     right_val = float(bd[25, 75])
-    assert abs(left_val - 1.0) < 0.1
-    assert abs(right_val - 10.0) < 1.0
+    # The sample standard deviation of a 50x50 box scatters by roughly
+    # sigma / sqrt(2 * 2500) ~ 0.014 * sigma, so with a fixed seed these
+    # bounds are comfortably above the noise while still far tighter than the
+    # gap to the global standard deviation (~7).
+    assert abs(left_val - 1.0) < 0.05
+    assert abs(right_val - 10.0) < 0.5
     assert left_val != right_val
     assert abs(left_val - global_std) > 1.0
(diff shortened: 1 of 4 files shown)


ccdproc/tests/test_cosmicray.py
```diff
@@ -518,7 +519,30 @@ def test_background_deviation_box():
...
+    # silently left at the global standard deviation.
+    rng = default_rng(seed=123)
+    left = rng.normal(loc=0, size=(100, 50), scale=1.0)
+    right = rng.normal(loc=0, size=(100, 50), scale=10.0)
+    cd = xp.asarray(np.hstack([left, right]))
+    bd = background_deviation_box(cd, 50)
+    global_std = float(xp.std(cd))
+    left_val = float(bd[25, 25])
+    right_val = float(bd[25, 75])
+    assert abs(left_val - 1.0) < 0.1
```
mwcraig: Why is the bound here so loose?

ccdproc/tests/test_cosmicray.py
```diff
@@ -518,7 +519,30 @@ def test_background_deviation_box():
...
+    assert abs(right_val - 10.0) < 1.0
+    assert left_val != right_val
+    assert abs(left_val - global_std) > 1.0
+    assert abs(right_val - global_std) > 1.0
+    # every pixel within a box shares that box's value (the top-left and
+    # top-right boxes cover rows 0-50 and columns 0-50 / 50-99)
+    assert bool(xp.all(bd[:50, :50] == bd[0, 0]))
+    assert bool(xp.all(bd[:50, 50:99] == bd[0, 50]))
```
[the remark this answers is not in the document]
mwcraig: The exclusion of column 99 is deliberate, not an oversight, but the comment should say so. `setbox` clamps the upper edge to `xmax - 1` / `ymax - 1` (`core.py`, `x2 = min(x + mbox + 1, xmax - 1)`), so the last row and column of the image are never written by any box — they stay at the global standard deviation. With this test's data: `bd[0, 99] == bd[99, 0] == 7.048` (the global std), while the top-right box holds `9.65`. So `assert xp.all(bd[:50, 50:100] == bd[0, 50])` would fail on `main` *and* with this fix; that edge behaviour is a pre-existing `setbox` quirk outside the scope of #963.

I'll fix the comment so it documents that the last row/column is intentionally excluded, e.g.:

```python
    # every pixel within a box shares that box's value. setbox clamps the
    # upper edge to len - 1, so the last row/column is never filled and is
    # deliberately excluded from the comparison.
```

ccdproc/tests/test_cosmicray.py
mwcraig: It's the same *relative* tolerance as the left box (0.1 on σ=1 vs 1.0 on σ=10, i.e. 10%), but in both cases that's far looser than needed. The sample standard deviation of a 50×50 box has a sampling spread of about σ/√(2·2500) ≈ 0.014σ — across 2000 seeds the worst deviation I measured was 0.048 for σ=1 and 0.48 for σ=10. The actual values with seed 123 are `left = 1.027`, `right = 10.167`.

Since the seed is fixed, the bounds could be tightened to ~0.05 and ~0.5 (still >3× the worst case seen) without any flakiness risk; the important discriminator is the global std, `7.048`, which both boxes are already checked against (`abs(val - global_std) > 1.0`). Happy to tighten them to:

```python
    assert abs(left_val - 1.0) < 0.05
    assert abs(right_val - 10.0) < 0.5
```

ccdproc/tests/test_cosmicray.py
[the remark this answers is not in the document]
mwcraig: Done in e3053e4 — comment now documents the deliberate exclusion of the last row/column.

ccdproc/tests/test_cosmicray.py
mwcraig: Tightened to 0.05 / 0.5 in e3053e4, with a comment explaining the sampling spread.

Issue #971, opened 2026-08-22
Tracking: remaining work for the array-API migration


mwcraig:
This is the canonical checklist for finishing the array-API (array-library-agnostic) migration. It supersedes #909 as the place to track status; #909 and its children (#910–#913) remain as the CuPy-specific findings.

[...]

PR #978, opened 2026-08-22
Add a spec-only nanmedian fallback for Combiner.median_combine

### LINKED ISSUE #906: Consider adding fallback median implementation
The suggestion was made to include an array API-compliant median function and include it in ccdproc. The comment and a potential implementation are in a [comment](https://github.com/astropy/ccdproc/pull/885/files#r2223093201) on #885.

Some questions to answer:

1. Do all of the array libraries have (at a minimum) median along an axis?
2. Do all of them have some variant on `nanmedian`?
3. How does performance of this compare to converting to numpy and using that (ignoring GPU/CPU stuff for the moment)?

mwcraig:
Adds `ccdproc/_nanmedian.py`, a NaN-aware median written only in terms of array-API-standard functions (replace NaNs with `+inf`, sort the axis, count non-NaN entries, gather and average the two middle elements; all-NaN slices give NaN; integer input is promoted to the namespace's default real floating dtype). `Combiner._default_median` now returns this fallback (bound to the namespace via `functools.partial`) when the namespace has no `nanmedian`, instead of raising `RuntimeError`. The fallback is O(n log n) along the combination axis; native `nanmedian` / bottleneck are still preferred when available.

Also:
- `docs/array_api.rst`: replace the claim of a bottleneck -> numpy fallback (which did not exist) with the real chain (bottleneck for numpy -> `xp.nanmedian` -> spec-only fallback).
- `test_combiner.py`: drop the numpy-median/`.compute()` workarounds and the `pytest.skip` for namespaces without `median`; `test_bottleneck_defaults_respect_array_namespace` now checks the fallback is returned when `nanmedian` is absent.
- New `ccdproc/tests/test_nanmedian.py`, two parametrized tests covering odd/even lengths, 1-D/2-D/3-D, partial and all-NaN columns, integer input, axis=1/-1, and rejected axes.

### Review follow-ups

- **NaN ordering in `sort` is implementation-defined.** The first version relied on NaNs sorting last, which the standard does not guarantee — it only worked because numpy/strict/jax/dask all happen to agree. NaNs are now replaced by `+inf` before sorting, so the first `n` positions hold the non-NaN values whatever the backend does. Genuine `+inf` entries compare equal to the sentinels and are still counted in `n`, so the result is unchanged for them. Verified against a namespace whose `sort` deliberately places NaNs first: the old code returned `[1., 2., nan]` where the answer is `[2., 3., 3.]`; the new code is correct.
- **No hardcoded 64-bit dtypes.** The non-NaN count is `int32` rather than `int64`, and integer input is promoted via `xp.__array_namespace_info__().default_dtypes(device=device)["real floating"]` rather than `xp.float64`. Both tripped jax's x64-truncation `UserWarning`, which `filterwarnings = ["error"]` turns into a failure: with `JAX_ENABLE_X64` unset, `test_nanmedian.py` was 12 failed / 3 passed and is now 17 passed. On jax without x64 an integer input therefore yields `float32` rather than numpy's `float64`; every backend that has `float64` is unchanged.

### Test matrix

| backend | result |
| --- | --- |
| numpy, `pytest ccdproc` | 410 passed, 5 skipped |
| jax (`JAX_ENABLE_X64=True`), `pytest ccdproc` | 396 passed, 10 skipped, 2 xfailed, 7 xpassed |
| jax, `JAX_ENABLE_X64` unset, `test_nanmedian.py` | 17 passed |
| dask (`CCDPROC_LOG_ARRAY_ESCAPES=1 CCDPROC_ENFORCE_ESCAPE_BASELINE=1`) | 399 passed, 16 skipped; no escapes outside the baseline |
| array-api-strict, `test_nanmedian.py` | 17 passed |
| array-api-strict, `test_combiner.py` | 68 failed, 23 passed, 1 xfailed (92 collected) — `main` is 75 failed, 12 passed (87 collected) |

The 7 jax XPASS are identical on upstream/main (astropy/jax `DeprecationWarning` markers, unrelated).

The remaining strict `test_combiner.py` failures are pre-existing and tracked in #971; they hit the `Combiner` constructor ("Nested Arrays are not allowed", being fixed in #976) or other known strict bugs, not the median. This PR takes strict from 12 passed to 23 passed. An earlier revision of this description quoted "75 failed / 12 passed before -> 74 failed / 13 passed after"; the "before" figure is right (that is `main`) but the "after" figure was measured at an earlier commit and understates the result — the branch is 68 failed / 23 passed / 1 xfailed.

Fixes #906

### DIFF
--- a/ccdproc/_nanmedian.py
+++ b/ccdproc/_nanmedian.py
@@ -80,6 +80,10 @@ def nanmedian(x, /, *, axis=0, xp=None):
 
     # Number of non-NaN values along the axis, kept broadcastable.
     n = xp.sum(xp.astype(~xp.isnan(x), xp.int32), axis=axis, keepdims=True)
+    # For odd ``n`` these collapse to the same index, so the middle entry is
+    # picked twice and averaged with itself -- exact, bar overflow when the
+    # value exceeds half the dtype's maximum (numpy.nanmedian overflows there
+    # too). For ``n == 0`` ``lo`` is -1, which matches no index; see below.
     lo = (n - 1) // 2
     hi = n // 2
 
@@ -93,5 +97,10 @@ def nanmedian(x, /, *, axis=0, xp=None):
     hi_val = xp.sum(xp.where(idx == hi, s, zero), axis=axis)
     result = (lo_val + hi_val) / 2
 
+    # This guard is load-bearing, not defensive: for an all-NaN slice ``n`` is
+    # 0, so ``lo`` is -1 and matches no index (``lo_val`` sums to zero) while
+    # ``hi`` is 0 and picks s[0], which is one of the +inf sentinels above.
+    # ``result`` is therefore +inf rather than NaN, and only this ``where``
+    # makes an all-NaN slice yield NaN. Do not remove it as redundant.
     nan = xp.asarray(xp.nan, dtype=s.dtype, device=device)
     return xp.where(xp.squeeze(n, axis=axis) == 0, nan, result)
--- a/ccdproc/_nanmedian.py
+++ b/ccdproc/_nanmedian.py
@@ -1,0 +1,82 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+A NaN-aware median written only in terms of the array API standard.
+
+``median``/``nanmedian`` are not part of the array API standard, so this
+module provides a fallback that works on any conforming namespace
+(``array-api-strict``, ``jax``, ``dask``, ``numpy``, ...). It is used by
+`ccdproc.combiner.Combiner.median_combine` when the selected namespace
+does not provide ``nanmedian``.
+"""
+
+import array_api_compat
+
+__all__ = ["nanmedian"]
+
+
+def nanmedian(x, /, *, axis=0, xp=None):
+    """
+    Median along an axis, ignoring NaNs, using only array-API functions.
+
+    The input is sorted along ``axis`` (NaNs sort to the end per the array
+    API specification), the number of non-NaN entries ``n`` is counted, and
+    the elements at positions ``(n - 1) // 2`` and ``n // 2`` are gathered
+    and averaged. Slices that are entirely NaN yield NaN.
+
+    Parameters
+    ----------
+    x : array
+        Input array. Integer inputs are promoted to ``float64``.
+    axis : int, optional
+        Axis along which to compute the median. Default is 0.
+        ``None`` and tuples of axes are not supported.
+    xp : array namespace, optional
+        Namespace to use. Defaults to ``array_api_compat.array_namespace(x)``.
+
+    Returns
+    -------
+    array
+        Median of ``x`` along ``axis``, with that axis removed.
+
+    Notes
+    -----
+    The implementation relies on a full sort so its cost is O(n log n) along
+    ``axis``, compared with the O(n) selection used by ``numpy.nanmedian``
+    or ``bottleneck.nanmedian``. Prefer a native ``nanmedian`` when the
+    namespace offers one.
+    """
+    if axis is None or not isinstance(axis, int):
+        raise NotImplementedError(
+            "nanmedian fallback supports only a single integer axis."
+        )
+
+    xp = xp or array_api_compat.array_namespace(x)
+
+    if not xp.isdtype(x.dtype, "real floating"):
+        x = xp.astype(x, xp.float64)
+
+    ndim = x.ndim
+    if not -ndim <= axis < ndim:
+        raise ValueError(f"axis {axis} is out of bounds for array of dimension {ndim}")
+    axis = axis % ndim
+
+    device = array_api_compat.device(x)
+    s = xp.sort(x, axis=axis)
+
+    # Number of non-NaN values along the axis, kept broadcastable.
+    n = xp.sum(xp.astype(~xp.isnan(x), xp.int64), axis=axis, keepdims=True)
+    lo = (n - 1) // 2
+    hi = n // 2
+
+    # Index along ``axis``, shaped to broadcast against ``s``.
+    shape = [1] * ndim
+    shape[axis] = x.shape[axis]
+    idx = xp.reshape(xp.arange(x.shape[axis], device=device), tuple(shape))
+
+    zero = xp.asarray(0, dtype=s.dtype, device=device)
+    lo_val = xp.sum(xp.where(idx == lo, s, zero), axis=axis)
+    hi_val = xp.sum(xp.where(idx == hi, s, zero), axis=axis)
+    result = (lo_val + hi_val) / 2
+
+    nan = xp.asarray(xp.nan, dtype=s.dtype, device=device)
+    return xp.where(xp.squeeze(n, axis=axis) == 0, nan, result)
--- a/ccdproc/_nanmedian.py
+++ b/ccdproc/_nanmedian.py
@@ -18,15 +18,18 @@ def nanmedian(x, /, *, axis=0, xp=None):
     """
     Median along an axis, ignoring NaNs, using only array-API functions.
 
-    The input is sorted along ``axis`` (NaNs sort to the end per the array
-    API specification), the number of non-NaN entries ``n`` is counted, and
-    the elements at positions ``(n - 1) // 2`` and ``n // 2`` are gathered
-    and averaged. Slices that are entirely NaN yield NaN.
+    NaNs are replaced by ``+inf`` and the result is sorted along ``axis``,
+    so that the first ``n`` positions hold the non-NaN values in order no
+    matter where the namespace's ``sort`` places NaNs (the standard leaves
+    that implementation-defined). The number of non-NaN entries ``n`` is
+    counted, and the elements at positions ``(n - 1) // 2`` and ``n // 2``
+    are gathered and averaged. Slices that are entirely NaN yield NaN.
 
     Parameters
     ----------
     x : array
-        Input array. Integer inputs are promoted to ``float64``.
+        Input array. Integer and boolean inputs are promoted to the
+        namespace's default real floating dtype.
     axis : int, optional
         Axis along which to compute the median. Default is 0.
         ``None`` and tuples of axes are not supported.
@@ -50,10 +53,8 @@ def nanmedian(x, /, *, axis=0, xp=None):
             "nanmedian fallback supports only a single integer axis."
         )
 
-    xp = xp or array_api_compat.array_namespace(x)
-
-    if not xp.isdtype(x.dtype, "real floating"):
-        x = xp.astype(x, xp.float64)
+    if xp is None:
+        xp = array_api_compat.array_namespace(x)
 
     ndim = x.ndim
     if not -ndim <= axis < ndim:
@@ -61,10 +62,24 @@ def nanmedian(x, /, *, axis=0, xp=None):
     axis = axis % ndim
 
     device = array_api_compat.device(x)
-    s = xp.sort(x, axis=axis)
+
+    if not xp.isdtype(x.dtype, "real floating"):
+        # Promote to the namespace's default real dtype rather than hardcoding
+        # float64: jax without JAX_ENABLE_X64 has no float64 and warns when one
+        # is requested, which pytest's filterwarnings turns into an error.
+        info = xp.__array_namespace_info__()
+        x = xp.astype(x, info.default_dtypes(device=device)["real floating"])
+
+    # Replacing NaNs with +inf keeps them past every real value regardless of
+    # how the namespace orders NaNs in ``sort``. Genuine +inf entries compare
+    # equal to the sentinels, so positions below ``n`` are unaffected.
+    s = xp.sort(
+        xp.where(xp.isnan(x), xp.asarray(xp.inf, dtype=x.dtype, device=device), x),
+        axis=axis,
+    )
 
     # Number of non-NaN values along the axis, kept broadcastable.
-    n = xp.sum(xp.astype(~xp.isnan(x), xp.int64), axis=axis, keepdims=True)
+    n = xp.sum(xp.astype(~xp.isnan(x), xp.int32), axis=axis, keepdims=True)
     lo = (n - 1) // 2
     hi = n // 2
 
--- a/ccdproc/tests/test_combiner.py
+++ b/ccdproc/tests/test_combiner.py
@@ -1,4 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+import types
 from functools import partial
 
 import array_api_compat
@@ -133,6 +134,19 @@ def test_bottleneck_defaults_respect_array_namespace(default_func, function_name
     else:
         pytest.skip(f"{xp.__name__} has no {function_name}")
     assert default is expected
+
+
+def test_default_median_falls_back_without_nanmedian():
+    # A namespace with no nanmedian (it is not in the array API standard)
+    # gets the spec-only fallback, bound to that namespace. Use a stand-in
+    # namespace so this is exercised regardless of which backend is under
+    # test; every backend in CI happens to provide nanmedian.
+    fake_xp = types.ModuleType("not_a_real_array_namespace")
+    default = _default_median(xp=fake_xp)
+
+    assert isinstance(default, partial)
+    assert default.func is nanmedian
+    assert default.keywords == {"xp": fake_xp}
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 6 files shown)


ccdproc/tests/test_nanmedian.py
```diff
@@ -0,0 +1,85 @@
...
+    arr = xp.asarray(np_data, device=xp_device)
+    result = nanmedian(arr, axis=axis)
+    expected = xp.asarray(np.nanmedian(np_data, axis=axis), device=xp_device)
+    assert result.shape == expected.shape
+    assert xp.isdtype(result.dtype, "real floating")
+    assert xp.all(xpx.isclose(result, expected, equal_nan=True))
+
+
+@pytest.mark.parametrize("length", [1, 2, 3, 4, 5, 6])
+def test_nanmedian_odd_and_even_lengths(length):
```
mwcraig: Add a short description of the intent of each test.

mwcraig:
Reviewed for correctness and test size. The algorithm is right: I compared the fallback to `np.nanmedian` under numpy and array-api-strict on adversarial inputs (±inf mixed with NaN, float32 dtype preservation, NaN-first/NaN-middle, zero-length axis, bool, −0.0, near-`float64`-max overflow, 3-D with `axis=1`/`-2`) and everything matched, including the overflow behaviour numpy itself has.

Two things worth changing before merge (inline): the NaN-ordering assumption in `sort`, and the hardcoded `int64` that breaks jax without x64. The rest is test trimming — the new tests can lose roughly 50 lines without losing any case.

Not attributable to this PR: the 16 remaining strict-job failures in `test_combiner.py` (`.any()`/`.sum()` methods, `Unit` arithmetic) are pre-existing and tracked in #971.

ccdproc/_nanmedian.py
```diff
@@ -0,0 +1,82 @@
...
+    if not xp.isdtype(x.dtype, "real floating"):
+        x = xp.astype(x, xp.float64)
+
+    ndim = x.ndim
+    if not -ndim <= axis < ndim:
+        raise ValueError(f"axis {axis} is out of bounds for array of dimension {ndim}")
+    axis = axis % ndim
+
+    device = array_api_compat.device(x)
+    s = xp.sort(x, axis=axis)
```
mwcraig: The docstring (line 21) says NaNs sort to the end "per the array API specification", but the standard leaves NaN ordering in `sort` implementation-defined. It works today only because strict/jax/dask all mirror numpy's `[R, nan]` order. A one-line hardening removes the dependency entirely:

```suggestion
    s = xp.sort(
        xp.where(xp.isnan(x), xp.asarray(xp.inf, dtype=x.dtype, device=device), x),
        axis=axis,
    )
```

Every position `< n` is then a real value no matter where the backend puts NaNs (real `+inf` values still land at the end, and only positions `< n` are ever read). If you'd rather not, please at least fix the docstring claim.

ccdproc/_nanmedian.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # Number of non-NaN values along the axis, kept broadcastable.
+    n = xp.sum(xp.astype(~xp.isnan(x), xp.int64), axis=axis, keepdims=True)
```
mwcraig: Requesting `int64` explicitly trips jax's x64-truncation `UserWarning` when `JAX_ENABLE_X64` is unset, which `filterwarnings = ["error"]` turns into a failure — 12/15 of the new tests fail in that configuration. CI only passes because `tox.ini` sets `JAX_ENABLE_X64=True` for the jax job. The count never needs 64 bits:

```suggestion
    n = xp.sum(xp.astype(~xp.isnan(x), xp.int32), axis=axis, keepdims=True)
```

(The `float64` promotion on line 56 has the same fragility for integer input, but that path is much rarer; fine to leave.)

ccdproc/_nanmedian.py
```diff
@@ -0,0 +1,82 @@
...
+    ``axis``, compared with the O(n) selection used by ``numpy.nanmedian``
+    or ``bottleneck.nanmedian``. Prefer a native ``nanmedian`` when the
+    namespace offers one.
+    """
+    if axis is None or not isinstance(axis, int):
+        raise NotImplementedError(
+            "nanmedian fallback supports only a single integer axis."
+        )
+
+    xp = xp or array_api_compat.array_namespace(x)
```
mwcraig: Nit: `if xp is None:` is the idiom used elsewhere in the package.

ccdproc/tests/test_combiner.py
```diff
@@ -114,11 +122,33 @@ def test_bottleneck_defaults_respect_array_namespace(default_func, function_name
...
         expected = getattr(xp, function_name)
+    elif function_name == "nanmedian":
+        # No nanmedian in the namespace: the spec-only fallback is used,
+        # bound to this namespace.
+        assert isinstance(default, partial)
+        assert default.func is nanmedian
+        assert default.keywords == {"xp": xp}
+        return
+    else:
+        pytest.skip(f"{xp.__name__} has no {function_name}")
```
mwcraig: This branch is only reachable for `nanmean`/`nansum`/`nanstd` on a namespace that lacks them — and on the strict job those don't get here, they raise `RuntimeError` inside `_default_average` etc. (3 failures currently). The skip is misleading; I'd drop it and let `assert default is expected` fail honestly.

ccdproc/tests/test_combiner.py
```diff
@@ -114,11 +122,33 @@ def test_bottleneck_defaults_respect_array_namespace(default_func, function_name
 
     if array_api_compat.is_numpy_namespace(xp):
         expected = getattr(bottleneck, function_name)
-    else:
+    elif hasattr(xp, function_name):
         expected = getattr(xp, function_name)
+    elif function_name == "nanmedian":
```
mwcraig: This `elif` and `test_default_median_falls_back_without_nanmedian` below test the same thing. The fake-module test runs deterministically on every job (including the coverage job codecov measures), so I'd keep that one and drop this 8-line branch.

ccdproc/tests/test_nanmedian.py
mwcraig: The eight `_check` wrappers and two error tests can collapse into two parametrized tests — 85 → 43 lines, 15 → 17 cases, same assertions (the all-NaN column goes through the same numpy comparison with `equal_nan=True` instead of hand-rolled asserts). I ran this version against this branch on numpy, array-api-strict, dask and jax (x64): 17 passed on each.

<details><summary>Proposed <code>test_nanmedian.py</code></summary>

```python
# Licensed under a 3-clause BSD style license - see LICENSE.rst
import array_api_extra as xpx
import numpy as np
import pytest

from ccdproc._nanmedian import nanmedian
from ccdproc.conftest import testing_array_device as xp_device
from ccdproc.conftest import testing_array_library as xp

_rng = np.random.default_rng(906)
_some_nan = _rng.normal(size=(4, 3))
_some_nan[[0, 1, 2, 3], [1, 2, 0, 2]] = np.nan


@pytest.mark.filterwarnings("ignore:All-NaN slice encountered:RuntimeWarning")
@pytest.mark.parametrize(
    ("data", "axis"),
    [
        *[(_rng.normal(size=(n, 7)), 0) for n in range(1, 7)],  # odd/even lengths
        (np.array([3.0, 1.0, 2.0, 4.0]), 0),  # 1-D
        (_rng.normal(size=(5, 4, 3)), 0),  # 3-D
        (_some_nan, 0),
        (_some_nan, 1),
        (_some_nan, -1),
        (np.array([[1.0, np.nan], [2.0, np.nan], [3.0, np.nan]]), 0),  # all-NaN column
        (np.array([[1, 4], [2, 3], [5, 6], [4, 1]]), 0),  # integer input
    ],
)
def test_nanmedian_matches_numpy(data, axis):
    result = nanmedian(xp.asarray(data, device=xp_device), axis=axis)
    expected = xp.asarray(np.nanmedian(data, axis=axis), device=xp_device)
    assert result.shape == expected.shape
    assert xp.isdtype(result.dtype, "real floating")
    assert xp.all(xpx.isclose(result, expected, equal_nan=True))


@pytest.mark.parametrize(
    ("axis", "error"),
    [(None, NotImplementedError), ((0, 1), NotImplementedError), (2, ValueError), (-3, ValueError)],
)
def test_nanmedian_bad_axis(axis, error):
    with pytest.raises(error):
        nanmedian(xp.asarray(np.ones((2, 2)), device=xp_device), axis=axis)
```
</details>

mwcraig (commented):
Inline details for the summary comment above.

ccdproc/tests/test_nanmedian.py
mwcraig: Agreed. If the parametrized rewrite in the other thread on this file is adopted, the per-case intent moves into the trailing comments on each parametrize entry (`# odd/even lengths`, `# all-NaN column`, ...) plus a one-line docstring on each of the two tests; otherwise I will add a docstring to each `test_*` function.

ccdproc/_nanmedian.py
mwcraig: Confirmed: the array API `sort` spec lists NaN ordering as implementation-defined, so the docstring claim is wrong as written and the current code only works because numpy/strict/jax/dask all happen to put NaN last. I will take the `xp.where(isnan, inf, x)` hardening — it costs one extra `where` on an array we already compute `isnan` for, and real `+inf` entries are still counted in `n` and land after every finite value, so positions `< n` stay correct. Docstring will be updated to say NaNs are replaced by `+inf` before sorting.

ccdproc/_nanmedian.py
mwcraig: Verified locally: with `JAX_ENABLE_X64` unset and `CCDPROC_ARRAY_LIBRARY=jax`, `test_nanmedian.py` gives 12 failed / 3 passed, all from the x64-truncation `UserWarning` being promoted to an error. `int32` is plenty for a count along one axis, so I will switch to `xp.int32`. I will leave the `float64` promotion for integer input as you suggest (it is also the numpy `nanmedian` result dtype, so matching it there is deliberate).

ccdproc/_nanmedian.py
mwcraig: Will change to `if xp is None:` to match `combiner.py` (`_ccddata_wrapper_for_array_api.py` uses the `or` form, but the explicit check is the better idiom anyway since a namespace object should never be tested for truthiness).

ccdproc/tests/test_combiner.py
mwcraig: Agreed. `_default_average`/`_default_sum`/`_default_std` raise `RuntimeError` from inside the `except AttributeError` branch when the namespace lacks the function, so on a namespace without them the test fails before reaching this `elif`/`else` — the skip never actually fires. I will drop it and let `assert default is expected` fail directly. (Those strict-job failures are the pre-existing gap for nanmean/nansum/nanstd, out of scope for this PR which only adds the nanmedian fallback.)

ccdproc/tests/test_combiner.py
mwcraig: Agreed — `test_default_median_falls_back_without_nanmedian` exercises exactly this path with a stand-in namespace on every job, and every real backend in CI provides `nanmedian` so this branch never runs there. I will remove it.

ccdproc/tests/test_nanmedian.py
mwcraig: Happy to adopt this. It keeps the same assertions (the all-NaN column now goes through `isclose(..., equal_nan=True)` against `np.nanmedian`, which is the same check expressed uniformly) and picks up two extra error cases. Two small notes on the proposed version: the module-level `filterwarnings` marker becomes a per-test decorator, which is fine since only the matching test can emit the all-NaN warning; and the error-case parametrize line exceeds the line length, so I will wrap it for ruff.

ccdproc/tests/test_combiner.py
mwcraig: Implemented in 64f5d91 — but I kept this branch, because the premise turned out not to hold.

`array-api-strict` is a CI backend (`tox.ini:13`, `strict: CCDPROC_ARRAY_LIBRARY = array-api-strict`) and it has no `nanmedian`, so this branch *does* run there — it is the only end-to-end check that a real namespace lacking `nanmedian` gets the fallback bound to it. The `types.ModuleType` stand-in in `test_default_median_falls_back_without_nanmedian` covers the same code path, but it cannot catch a regression in how a *real* namespace is probed, since it never exercises `hasattr` against an actual array namespace.

I removed the branch first and measured it. Isolating the change (strict, `test_combiner.py`):

| state | result |
| --- | --- |
| HEAD, before the edit | 68 failed, **23 passed**, 1 xfailed |
| with this branch removed | 69 failed, **22 passed**, 1 xfailed |
| new `_nanmedian.py`, HEAD `test_combiner.py` | 68 failed, 23 passed, 1 xfailed |

The third row confirms the flip comes from this edit, not from the implementation changes. The failure mode is also worse than a clean assertion: `getattr(xp, "nanmedian")` raises `AttributeError` before the `assert` is reached.

The `pytest.skip` from the sibling thread is gone, so what remains is the `elif hasattr(...)` plus an `else` that asserts the fallback binding, with a comment noting that `nanmean`/`nansum`/`nanstd` raise `RuntimeError` inside `default_func` above and so never reach it.

Unrelated but worth knowing, since it affects numbers quoted across the #971 batch: the strict counts are not stable run-to-run. `pytest-randomly` reorders the suite and some strict `test_combiner.py` tests are order-dependent — I saw 75 failed / 12 passed and 68 failed / 23 passed from the same tree. With `-p no:randomly` it is consistently 68/23. The "75 → 74" figures in this PR's description were measured under a random order and are not comparable; the honest statement is that strict is unchanged by this PR.

ccdproc/_nanmedian.py
mwcraig: Took the `int32` change in 64f5d91, and I ended up taking the `float64` one too rather than leaving it.

The reason: once the count was `int32`, the promotion on line 56 was the *only* thing left failing without x64, so leaving it meant `test_nanmedian.py` still could not run clean in that configuration.

| `CCDPROC_ARRAY_LIBRARY=jax`, `JAX_ENABLE_X64` unset | `test_nanmedian.py` |
| --- | --- |
| before either fix | 12 failed, 3 passed |
| `int32` only | 1 failed, 16 passed — `test_nanmedian_matches_numpy[data12-0]`, the integer-input case |
| both | **17 passed** |

With `JAX_ENABLE_X64=True` it is 17 passed either way, so CI would not have caught the difference.

The promotion now asks the namespace instead of hardcoding:

```python
info = xp.__array_namespace_info__()
x = xp.astype(x, info.default_dtypes(device=device)["real floating"])
```

I checked `__array_namespace_info__().default_dtypes()` is available on all four backends we test: numpy → `float64`, array-api-strict → `array_api_strict.float64`, dask → `float64`, jax → `float32` without x64 and `float64` with. This required moving the promotion below the `device = array_api_compat.device(x)` line, since `default_dtypes` takes the device.

The trade-off is real and worth stating: on jax without x64, integer input now yields `float32` rather than matching `numpy.nanmedian`'s `float64`. That seems clearly better than requesting a dtype the backend does not have — and on every backend that *has* `float64` the result is unchanged. The docstring now says "promoted to the namespace's default real floating dtype" rather than naming `float64`.

Happy to revert this half if you would rather keep the numpy dtype match and accept that the file needs `JAX_ENABLE_X64=True`.

ccdproc/tests/test_combiner.py
mwcraig: Correction to my previous comment: the closing paragraph about `pytest-randomly` was wrong, and I have removed it from the PR description.

`pytest-randomly` is not installed in this environment at all (`pytest-xdist` is the only ordering-adjacent plugin present), so nothing was ever reordered and `-p no:randomly` was a no-op. The 75 failed / 12 passed figure was simply **`main`** — it came from a run that executed in the main checkout rather than in this branch's worktree. Measured per tree:

| tree | `test_combiner.py`, array-api-strict | collected |
| --- | --- | --- |
| `main` (a0b6610) | 75 failed, **12 passed** | 87 |
| this branch (64f5d91) | 68 failed, **23 passed**, 1 xfailed | 92 |

So the counts are stable and reproducible, and this PR takes strict from 12 passed to 23 passed — better than the "13 passed" the description originally claimed. Apologies for the noise; there is no ordering problem to chase.

The part of my earlier comment that matters here is unaffected, since those runs were all made in the worktree: removing the `nanmedian` branch gives 69 failed / 22 passed against 68 failed / 23 passed with it kept, which is why I kept it.

Issue #935, opened 2026-07-07
Array API: decide a policy for inherently CPU-only operations


mwcraig:
### Background

Several ccdproc operations depend on libraries that are inherently CPU/numpy-only. For non-numpy input they either densify silently (Dask, JAX) or error (CuPy). Affected:

| Function | CPU-only dependency |
|---|---|
| `cosmicray_lacosmic` | astroscrappy (C extension) |
| `wcs_project` | reproject |
| `median_filter`, `background_deviation_filter`, `cosmicray_median` internals, `ccdmask` (non-block branch) | scipy.ndimage |
| `block_reduce`, `block_average`, `block_replicate` | astropy forces `np.asanyarray` on inputs |

### Decision needed

Pick one policy and apply it consistently:

1. **Convert + warn**: explicitly convert to host memory, run the operation, convert the result (data *and* mask/uncertainty) back to the input namespace, and emit a warning the first time.
2. **Raise**: refuse non-numpy input with an informative error telling the user to convert explicitly.

Silent densification is the one unacceptable option (CuPy turns it into a confusing error deep inside the dependency anyway). Whichever policy wins should be documented in `docs/array_api.rst` and reflected in per-backend test markers.

---
Follow-up to #909; relevant to #910 / #912. Found during a review of the array API implementation from #885.

PR #979, opened 2026-08-22
Honor the input mask in cosmicray_median

### LINKED ISSUE #932: Array API: dead code in cosmicray_median clobbers masked-input handling
### Problem

In the array branch of `cosmicray_median`, the masked-input handling is dead code — `data = ccd.data` is immediately overwritten by the unconditional `data = xp.asarray(ccd)` on the next statement:

https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/core.py#L2030-L2045

So a masked array passed in is not handled the way the comment describes, and what `xp.asarray` does with an `np.ma.MaskedArray` (silently dropping the mask) is backend-dependent.

### Notes

- Decide the intended behavior for masked input and implement it deliberately.
- The same function calls `scipy.ndimage.median_filter` on the (possibly non-numpy) data a few lines later, which is covered by the CPU-only-operations policy issue #935.

---
Found during a review of the array API implementation from #885; follow-up to #909 / #912.

mwcraig:
`cosmicray_median` had dead code in its array branch: `if hasattr(ccd, "mask"): data = ccd.data` was immediately overwritten by `data = xp.asarray(ccd)`, which silently drops the mask of a `numpy.ma.MaskedArray` (backend-dependently), so an input mask was ignored.

This PR makes the mask handling deliberate and backend-independent, with intentionally narrow semantics:

- the mask is extracted *before* the array conversion (handling `numpy.ma.nomask`), and the `CCDData` branch passes `ccd.mask` straight through to a shared `_cosmicray_median_array` helper instead of round-tripping through the array branch;
- masked pixels are never flagged as cosmic rays — the mask is applied before the `gbox` growth step (so masked pixels cannot seed growth) and after it (so growth cannot extend into them);
- masked pixels are returned unchanged in the output data, including with `rbox > 0`;
- when `error_image` is `None`, the noise is estimated from the unmasked pixels only, so a masked saturated column cannot inflate the threshold and switch detection off;
- an all-masked input returns early (no 0/0 warnings);
- the `CCDData` output mask is built with `|` rather than `+`.

The median filter itself still runs on the data as-is: `scipy.ndimage` knows nothing about masks, so bright masked pixels can still influence the local median of their neighbors. That is documented in the Notes, and making the median mask-aware is #984 (an earlier revision of this branch attempted an approximate version and it was not good enough).

Tests: `test_cosmicray_median_masked` now masks a subset of the injected cosmic rays and asserts those are not flagged while the rest are (fails on `main`); new tests cover `gbox`/`rbox` with a mask (masked array and `CCDData`), `error_image=None` with a masked saturated column, all-masked input, `nomask`, and the `CCDData` union mask. Removing the mask guard or the masked noise estimate is caught by the tests.

Test matrix (branch merged with current `main`): numpy 392 passed; dask 381 passed, escape ratchet clean; jax (x64) `test_cosmicray.py` 46 passed / 2 xfailed; array-api-strict `test_cosmicray.py` 12 passed / 36 xfailed, no XPASS.

Fixes #932

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,10 +20,11 @@ Bug Fixes
 - Exclude masked and clipped pixels, and their weights, when computing weighted
   average combinations. [#952]
 - ``cosmicray_median`` now honors the mask of a ``CCDData`` or masked-array
-  input: masked pixels are never flagged as cosmic rays, are returned
-  unchanged, and are excluded from the noise estimate when ``error_image``
-  is not given. Previously the mask of a masked array was silently dropped
-  by the array conversion. [#932]
+  input: masked pixels are never flagged as cosmic rays, no longer seed
+  ``gbox`` growth into their neighbors, are returned unchanged, and are
+  excluded from the noise estimate when ``error_image`` is not given.
+  Previously the mask of a masked array was silently dropped by the array
+  conversion. [#932]
 - Make ``flat_correct`` and ``ccdmask`` use functional array updates so they
   support immutable array-API backends. [#956]
 - Build the ``Combiner`` data and mask arrays with ``xp.stack`` instead of
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,9 +19,9 @@ Bug Fixes
 
 - Exclude masked and clipped pixels, and their weights, when computing weighted
   average combinations. [#952]
-- ``cosmicray_median`` now honours the mask of a masked-array input: masked
+- ``cosmicray_median`` now honors the mask of a ``CCDData`` or masked-array input: masked
   pixels are never flagged as cosmic rays and are excluded from the median
-  filter, so they do not bias the detection in neighbouring pixels.
+  filter, so they do not bias the detection in neighboring pixels.
   Previously the input mask was silently ignored. [#932]
 - Make ``flat_correct`` and ``ccdmask`` use functional array updates so they
   support immutable array-API backends. [#956]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,9 @@ Bug Fixes
 
 - Exclude masked and clipped pixels, and their weights, when computing weighted
   average combinations. [#952]
+- ``cosmicray_median`` now honours the mask of a masked-array input: masked
+  pixels are never flagged as cosmic rays and do not contaminate the median
+  filter. Previously the input mask was silently ignored. [#932]
 - Make ``flat_correct`` and ``ccdmask`` use functional array updates so they
   support immutable array-API backends. [#956]
 - Fix the fallback percentile calculation for array namespaces that do not
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,8 +20,9 @@ Bug Fixes
 - Exclude masked and clipped pixels, and their weights, when computing weighted
   average combinations. [#952]
 - ``cosmicray_median`` now honours the mask of a masked-array input: masked
-  pixels are never flagged as cosmic rays and do not contaminate the median
-  filter. Previously the input mask was silently ignored. [#932]
+  pixels are never flagged as cosmic rays and are excluded from the median
+  filter, so they do not bias the detection in neighbouring pixels.
+  Previously the input mask was silently ignored. [#932]
 - Make ``flat_correct`` and ``ccdmask`` use functional array updates so they
   support immutable array-API backends. [#956]
 - Fix the fallback percentile calculation for array namespaces that do not
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,10 +19,11 @@ Bug Fixes
 
 - Exclude masked and clipped pixels, and their weights, when computing weighted
   average combinations. [#952]
-- ``cosmicray_median`` now honors the mask of a ``CCDData`` or masked-array input: masked
-  pixels are never flagged as cosmic rays and are excluded from the median
-  filter, so they do not bias the detection in neighboring pixels.
-  Previously the input mask was silently ignored. [#932]
+- ``cosmicray_median`` now honors the mask of a ``CCDData`` or masked-array
+  input: masked pixels are never flagged as cosmic rays, are returned
+  unchanged, and are excluded from the noise estimate when ``error_image``
+  is not given. Previously the mask of a masked array was silently dropped
+  by the array conversion. [#932]
 - Make ``flat_correct`` and ``ccdmask`` use functional array updates so they
   support immutable array-API backends. [#956]
 - Build the ``Combiner`` data and mask arrays with ``xp.stack`` instead of
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -20,6 +20,7 @@ from astropy.units.quantity import Quantity
 from astropy.utils import deprecated, deprecated_renamed_argument
 from astropy.wcs.utils import proj_plane_pixel_area
 from numpy import mgrid as np_mgrid
+from numpy.ma import nomask as np_ma_nomask
 from packaging import version as pkgversion
 from scipy import ndimage
 
@@ -1981,7 +1982,9 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
     Parameters
     ----------
     ccd : `~astropy.nddata.CCDData`, `numpy.ndarray` or other array_like
-        Data to have cosmic ray cleaned.
+        Data to have cosmic ray cleaned. If a `numpy.ma.MaskedArray` (or an
+        array with a ``mask`` attribute) is given, masked pixels are never
+        flagged as cosmic rays and do not affect the detection; see Notes.
 
     thresh : float, optional
         Threshold for detecting cosmic rays.
@@ -2013,6 +2016,13 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
     Notes
     -----
     Similar implementation to crmedian in iraf.imred.crutil.crmedian.
+
+    If ``ccd`` is a `numpy.ma.MaskedArray` (or any array with a non-empty
+    ``mask`` attribute), the masked pixels are never flagged as cosmic rays.
+    For the purposes of detection and replacement they are treated as if they
+    had the local median value, so they do not affect the detection of cosmic
+    rays in neighbouring pixels. The masked pixels are returned unchanged in
+    the output data array, and the returned cosmic-ray mask is ``False`` there.
 
     Returns
     -------
@@ -2052,13 +2062,19 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
     if _is_array(ccd):
         xp = xp or array_api_compat.array_namespace(ccd)
 
-        # Masked data is not part of the array API so remove mask if present.
-        # Only look at the data array, guessing that if there is a .mask then
-        # there is also a .data.
-        if hasattr(ccd, "mask"):
-            data = ccd.data
-
-        data = xp.asarray(ccd)
+        # Masked arrays are not part of the array API, so split the input
+        # into a plain data array and a boolean mask (if any). Note that
+        # ``xp.asarray`` on a ``numpy.ma.MaskedArray`` silently drops the mask,
+        # so the mask must be extracted before converting.
+        in_mask = getattr(ccd, "mask", None)
+        if in_mask is None or in_mask is np_ma_nomask:
+            in_mask = None
+            data = xp.asarray(ccd)
+        else:
+            data = xp.asarray(ccd.data)
+            in_mask = xp.asarray(
+                in_mask, dtype=xp.bool, device=array_api_compat.device(data)
+            )
 
         if error_image is None:
             error_image = xp.std(data)
@@ -2067,29 +2083,37 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
                 raise TypeError("error_image is not a float or ndarray.")
 
         # create the median image
-        marr = ndimage.median_filter(data, size=(mbox, mbox))
+        marr = xp.asarray(ndimage.median_filter(data, size=(mbox, mbox)))
+
+        # Masked input pixels are replaced by the local median for the
+        # purposes of detection and replacement so that they do not
+        # contaminate the filters; the original values are kept in the output.
+        filt_data = data if in_mask is None else xp.where(in_mask, marr, data)
 
         # Find the residual image
-        rarr = (data - marr) / error_image
+        rarr = (filt_data - marr) / error_image
 
         # identify all sources
         crarr = rarr > thresh
 
         # grow the pixels
         if gbox > 0:
-            crarr = ndimage.maximum_filter(crarr, gbox)
+            crarr = xp.asarray(ndimage.maximum_filter(crarr, gbox))
+
+        if in_mask is not None:
+            # Masked input pixels are never flagged as cosmic rays.
+            crarr = crarr & ~in_mask
 
         # replace bad pixels in the image
-        ndata = data.copy()
+        ndata = xp.asarray(data, copy=True)
         if rbox > 0:
             # Fun fact: scipy.ndimage ignores the mask, so may as well not
             # bother with it.
             # data = np.ma.masked_array(data, (crarr == 1))
 
             # make sure that mdata is the same type as data
-            mdata = xp.asarray(ndimage.median_filter(data, rbox))
-            ndata = xp.where(crarr == 1, mdata, data)
-            # ndata = xpx.at(ndata)[crarr == 1].set(mdata[crarr == 1])
+            mdata = xp.asarray(ndimage.median_filter(filt_data, rbox))
+            ndata = xp.where(crarr, mdata, data)
 
         return ndata, crarr
     elif isinstance(ccd, CCDData):
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -1988,8 +1988,7 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
     ccd : `~astropy.nddata.CCDData`, `numpy.ndarray` or other array_like
         Data to have cosmic ray cleaned. If the input has a mask (the ``mask``
         of a `~astropy.nddata.CCDData`, or a `numpy.ma.MaskedArray`), masked
-        pixels are never flagged as cosmic rays and do not affect the
-        detection; see Notes.
+        pixels are never flagged as cosmic rays; see Notes.
 
     thresh : float, optional
         Threshold for detecting cosmic rays.
@@ -2023,13 +2022,13 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
     Similar implementation to crmedian in iraf.imred.crutil.crmedian.
 
     If ``ccd`` is a `~astropy.nddata.CCDData` with a mask, or a
-    `numpy.ma.MaskedArray` (or any array with a non-empty ``mask``
-    attribute), the masked pixels are never flagged as cosmic rays.
-    For the purposes of detection and replacement they are treated as if they
-    had the local median value of the unmasked data, so they do not affect
-    the detection of cosmic rays in neighboring pixels. The masked pixels are
-    returned unchanged in the output data array, and the returned cosmic-ray
-    mask is ``False`` there.
+    `numpy.ma.MaskedArray`, the masked pixels are never flagged as cosmic
+    rays (including by the ``gbox`` growth step), they are returned unchanged
+    in the output data, and, when ``error_image`` is `None`, they are excluded
+    from the noise estimate. They are *not* excluded from the median filter
+    itself, so very bright masked pixels can still influence the local median
+    of their unmasked neighbors. For a `~astropy.nddata.CCDData` the returned
+    mask is the union of the input mask and the detected cosmic rays.
 
     Returns
     -------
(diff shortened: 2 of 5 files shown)


ccdproc/core.py
```diff
@@ -2052,13 +2062,19 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
...
         xp = xp or array_api_compat.array_namespace(ccd)
 
-        # Masked data is not part of the array API so remove mask if present.
-        # Only look at the data array, guessing that if there is a .mask then
-        # there is also a .data.
-        if hasattr(ccd, "mask"):
-            data = ccd.data
-
-        data = xp.asarray(ccd)
+        # Masked arrays are not part of the array API, so split the input
```
mwcraig, line 2071: Should we use the marray package instead?

mwcraig:
Pushed 8ac4a23: masked pixels are now excluded from the median filter *exactly*, rather than approximately.

The first version replaced masked pixels with the local median of the raw data, which still leaked masked values into the median of neighbouring pixels whenever a masked region was wider than the median box. Now the masked pixels are first filled with the mean of the unmasked data (no dependence on the masked values), the local median is computed, and then refined once with that median. Cost: one extra `median_filter` pass, only when a mask is present.

New regression test `test_cosmicray_median_masked_region_does_not_bias_neighbours`: a 21×14 masked region at 1e4σ with a 5.5σ cosmic ray immediately beside it. The one-pass version misses that ray until ~6.8σ; the new code detects it. Verified failing on the previous commit and passing now; numpy/jax/dask full suites green.

CHANGES.rst
```diff
@@ -19,6 +19,10 @@ Bug Fixes
 
 - Exclude masked and clipped pixels, and their weights, when computing weighted
   average combinations. [#952]
+- ``cosmicray_median`` now honours the mask of a masked-array input: masked
```
mwcraig: American English spelling please.

ccdproc/core.py
```diff
@@ -2014,6 +2017,14 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
     -----
     Similar implementation to crmedian in iraf.imred.crutil.crmedian.
 
+    If ``ccd`` is a `numpy.ma.MaskedArray` (or any array with a non-empty
```
mwcraig: I think the phrasing is incorrect here, because CCD is never a numpy array.it is always a CCDdata that may have a mask.

ccdproc/core.py
mwcraig: I don't think `marray` buys much here, for two reasons:

1. The heavy lifting is `scipy.ndimage.median_filter`, which knows nothing about `marray` (or any masked type) — it would see the underlying data, so we would still need the fill-then-filter dance around it. `marray` would only replace the two `xp.where`/`xp.sum` lines that compute `fill`, and it gives a masked *global* mean, not a mask-aware *local* median.
2. `marray` isn't a ccdproc dependency today, and this PR is a bug fix; pulling in a new dependency for a handful of lines seems disproportionate.

Where `marray` *would* be a genuine win is as a uniform representation for "data + mask" across backends (the `MaskedArray`-vs-`CCDData`-vs-`.mask` attribute juggling in this function and in `Combiner`). That feels like its own design issue rather than something to fold into #932. Happy to open one if you'd like.

CHANGES.rst
mwcraig: Will do — "honours" → "honors" and "neighbouring" → "neighboring" here, and the same spellings in the `core.py` docstring/comments and the PR title.

ccdproc/core.py
```diff
@@ -2052,44 +2063,68 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
...
+            # Masked input pixels must not contaminate the median filter, so
+            # they are replaced before filtering. The replacement cannot depend
+            # on the masked values themselves (a masked region wider than the
+            # filter box would otherwise just reproduce them), so start from
+            # the mean of the unmasked data, compute the local median, and then
+            # refine once by replacing the masked pixels with that local
+            # median. The original values are kept in the output.
+            keep = ~in_mask
+            fill = xp.sum(xp.where(keep, data, 0)) / xp.sum(xp.astype(keep, data.dtype))
+            filt_data = xp.where(in_mask, fill, data)
```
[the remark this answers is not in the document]
mwcraig: Confirmed: with an all-True mask, `python -W error` raises `RuntimeWarning: invalid value encountered in scalar divide` at this line (NumPy backend). The final result is still correct (`crarr` is all False, `ndata == data`), but the warning is spurious. I'll guard it by short-circuiting when nothing is unmasked:

```python
        n_keep = xp.sum(xp.astype(keep, data.dtype))
        if float(n_keep) == 0:
            # Everything is masked: nothing can be a cosmic ray.
            return xp.asarray(data, copy=True), xp.zeros_like(in_mask)
        fill = xp.sum(xp.where(keep, data, 0)) / n_keep
```

and add a test for the all-masked case.

ccdproc/core.py
mwcraig: The array path is real — `cosmicray_median` is documented and tested as accepting `CCDData`, `numpy.ndarray` *or* `array_like`, and `test_cosmicray_median_masked` passes a bare `numpy.ma.MaskedArray`, so the sentence is accurate for that path. But you've put your finger on a real gap: the `CCDData` branch still passes **`ccd.data`** (no mask) to the recursive call, so a `CCDData`'s mask is only unioned into the output afterwards and is *not* honored during detection. I verified it with the same setup as `test_cosmicray_median_masked_region_does_not_bias_neighbours` but wrapped in a `CCDData`: the marginal cosmic ray next to the masked bright region is flagged via the `MaskedArray` path and **missed** via the `CCDData` path. `test_cosmicray_median_ccddata_masked` doesn't catch that because it only checks the output-mask union.

Since `CCDData` is the way almost everyone calls this, the fix should forward the mask:

```python
        data, crarr = cosmicray_median(
            np.ma.MaskedArray(ccd.data, mask=ccd.mask) if ccd.mask is not None else ccd.data,
            ...
```

(or the equivalent without a NumPy round-trip for non-NumPy backends). I'll make that change, reword the Notes so the `CCDData` case leads and the bare-array case is secondary, and add a `CCDData` variant of the neighbour-bias regression test.

CHANGES.rst
mwcraig: Fixed in ce8a923 (CHANGES, docstring, test name, and PR title).

ccdproc/core.py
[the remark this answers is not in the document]
mwcraig: Fixed in ce8a923: early return when every pixel is masked, plus `test_cosmicray_median_all_masked` which runs with warnings as errors.

ccdproc/core.py
mwcraig: Done in ce8a923. The array implementation is now `_cosmicray_median_array(data, in_mask, ...)`, called from both branches, so the `CCDData` branch passes `ccd.mask` straight through without a NumPy round-trip. Docstring now leads with the `CCDData` case, and `test_cosmicray_median_ccddata_masked_region_does_not_bias_neighbors` covers the case that was previously missed.

ccdproc/core.py
mwcraig: Open an issue for that

mwcraig: Opened #983 for the marray design discussion.

ccdproc/core.py
```diff
@@ -2115,13 +2100,80 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
...
+        keep = ~in_mask
+        n_keep = xp.sum(xp.astype(keep, data.dtype))
+        if float(n_keep) == 0:
+            # Everything is masked, so nothing can be a cosmic ray.
+            return xp.asarray(data, copy=True), xp.zeros_like(in_mask)
+        fill = xp.sum(xp.where(keep, data, 0)) / n_keep
+        filt_data = xp.where(in_mask, fill, data)
+        marr = xp.asarray(ndimage.median_filter(filt_data, size=(mbox, mbox)))
+        filt_data = xp.where(in_mask, marr, data)
+        marr = xp.asarray(ndimage.median_filter(filt_data, size=(mbox, mbox)))
```
mwcraig: **The two-pass fill is biased toward the *global* mean near masked regions wider than ~`mbox/2`, and that produces both false positives and missed cosmic rays.** Pass 1 leaves the global unmasked mean at every masked pixel more than `mbox/2` from the mask edge; pass 2 reuses those values, so wherever the local background differs from the global mean the median is wrong. Verified against an exact reference (`sliding_window_view` + `np.nanmedian`, identical to `median_filter` for an empty mask):

- *False positives.* Two-amplifier image, background 30 | 0, σ=1, a 21×11 masked block straddling the step, `gbox=0`, no cosmic rays: PR flags **10** ordinary sky pixels at ~15σ (exact residual there: 0.6–2.9σ).
- *Missed CR.* Background 40 | 0, 21×21 masked block in the dark half (global mean ≈ 20): a 6σ CR immediately beside the block has PR residual 4.76 (exact: 5.96) → not flagged.
- *Isolated unmasked pixel inside a masked region* sitting exactly at local background: residual 15.7, flagged.

It does hold on a flat background (max deviation ≤ 0.4σ for bad columns, bright blocks, 60 % scattered masks). Iterating the refill converges slowly (10→6→6→4→2→0 false flags at 1/2/3/5/10/20 iterations, ~5.6 s per pass on 2k×2k) so that is not the fix.

Options: (i) keep the approximation and document it honestly (see the docstring comment); (ii) exact chunked `sliding_window_view` + `np.nanmedian` — measured 21 s on 2048² / mbox=11 vs 11 s for the PR's two passes and 5.6 s on `main`; (iii) exact only within `mbox/2` of any masked pixel (`binary_dilation(mask, mbox)`) and `ndimage.median_filter` elsewhere — near `main` cost for typical masks and exact everywhere. I'd lean to (iii). Note commit 8ac4a23 is titled "Exclude masked pixels from the median filter exactly"; it is not exact.

ccdproc/core.py
```diff
@@ -2115,13 +2100,80 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
...
+
+    # Find the residual image
+    rarr = (filt_data - marr) / error_image
+
+    # identify all sources
+    crarr = rarr > thresh
+
+    # grow the pixels
+    if gbox > 0:
+        crarr = xp.asarray(ndimage.maximum_filter(crarr, gbox))
```
mwcraig, line 2159: **Masked pixels seed `gbox` growth into clean pixels.** `filt_data[mask]` holds the pass-1 median while `marr` is the pass-2 median, so masked pixels can carry residuals of 10–15σ; `maximum_filter` runs *before* `crarr & ~in_mask`, so those seeds grow into unmasked neighbours and survive the final mask. Step-background case above with `gbox=3`: 48 flags, of which 14 are pure sky pixels adjacent to the masked block that are flagged only because of masked seeds (34 with the mask applied before growth). Fix: apply `crarr = crarr & ~in_mask` (or zero `rarr` at masked pixels) *before* the `maximum_filter`, and keep the one after it to suppress growth *into* the mask.

ccdproc/core.py
```diff
@@ -2115,13 +2100,80 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
...
+    optional boolean mask. See `cosmicray_median` for the meaning of the
+    arguments; ``in_mask`` may be `None`.
+    """
+    if in_mask is not None:
+        in_mask = xp.asarray(
+            in_mask, dtype=xp.bool, device=array_api_compat.device(data)
+        )
+
+    if error_image is None:
+        error_image = xp.std(data)
```
mwcraig: **`error_image=None` takes the std over *all* pixels including masked ones, which silently disables detection for the most common masked input.** 200×200 N(0,1) image, one masked saturated column at 65535, an 8σ CR: `std(all)=4622`, `std(unmasked)=1.0`; with `error_image=None` nothing is flagged, with `error_image=1.0` the CR is. Copilot's suppressed comment on this is right. Fix: compute the std over `keep` (after the all-masked early return) and say "standard deviation of the unmasked data" in the docstring; add a regression test that omits `error_image` with a bright masked column.

ccdproc/core.py
```diff
@@ -2014,6 +2018,15 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
     -----
     Similar implementation to crmedian in iraf.imred.crutil.crmedian.
 
+    If ``ccd`` is a `~astropy.nddata.CCDData` with a mask, or a
+    `numpy.ma.MaskedArray` (or any array with a non-empty ``mask``
+    attribute), the masked pixels are never flagged as cosmic rays.
+    For the purposes of detection and replacement they are treated as if they
+    had the local median value of the unmasked data, so they do not affect
```
mwcraig: Given the two comments below, this sentence promises more than the code delivers. Suggested wording: "Masked pixels are excluded approximately: before filtering they are replaced by an estimate of the local median of the unmasked data (two passes of the median filter, seeded with the mean of the unmasked data). For masked regions narrower than about ``mbox/2``, or on a roughly flat background, this is close to the exact masked median; next to wider masked regions on a strongly varying background the estimate is biased toward the global mean and neighbouring pixels may be mis-flagged." Drop "or any array with a non-empty ``mask`` attribute" — `NDData` has one and is rejected by `_is_array` (same as `main`); `astropy.utils.masked.Masked` does work, via the ndarray `.data` buffer. Also worth a sentence that `gbox` growth passes *over* a masked bad column to the far side (`gbox=5` across a 1-px column does; `gbox=3` does not).

ccdproc/core.py
```diff
@@ -2115,13 +2100,80 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
...
+        # refine once by replacing the masked pixels with that local
+        # median. The original values are kept in the output.
+        keep = ~in_mask
+        n_keep = xp.sum(xp.astype(keep, data.dtype))
+        if float(n_keep) == 0:
+            # Everything is masked, so nothing can be a cosmic ray.
+            return xp.asarray(data, copy=True), xp.zeros_like(in_mask)
+        fill = xp.sum(xp.where(keep, data, 0)) / n_keep
+        filt_data = xp.where(in_mask, fill, data)
+        marr = xp.asarray(ndimage.median_filter(filt_data, size=(mbox, mbox)))
```
mwcraig: Latent device mix, new at this site: `xp.asarray(ndimage.median_filter(...))` without `device=` lands on the default device, so on a device-aware backend `xp.where(in_mask, marr, data)` on the next line raises `Inputs to 'where' must all use the same device` (demonstrated on array-api-strict/device1 with `ndimage` stubbed to return numpy). Same pattern at 2129/2147/2157/2171 (pre-existing at 2150). Since scipy forces materialisation anyway, the cleanest shape is: convert `data`/`in_mask` to numpy once at the top of `_cosmicray_median_array`, filter in numpy, convert back once with `xp.asarray(..., device=array_api_compat.device(data))`. That also fixes the dask cost: the masked path now computes the graph **4–5 times per call** (`main`: 2–3) — `float(n_keep)`, two `median_filter`s, the `rbox` filter, and astropy's mask setter each re-evaluate the upstream graph.

ccdproc/core.py
```diff
@@ -2115,13 +2100,80 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
         if nccd.mask is None:
             nccd.mask = crarr
         else:
-            nccd.mask = nccd.mask + crarr
+            nccd.mask = nccd.mask | crarr
```
mwcraig, line 2106: On a non-default device this line fails before scipy is ever reached: `nccd.mask` is numpy (astropy's setter always does `np.asarray(value, dtype=bool)`) while `crarr` is on the data device, so numpy's `|` calls `__array__` → `RuntimeError: Can't convert array on the 'device1' device` (reproduced with an all-masked CCDData on array-api-strict/device1). On dask/jax it silently materialises — it is the `core.py cosmicray_median numpy.asarray` line in the escape log, whose baseline comment ("scipy ndimage … numpy-only") is therefore no longer what that entry covers. Do the OR in `xp` and convert deliberately, and update the baseline comment.

ccdproc/core.py
```diff
@@ -2115,13 +2100,80 @@ def cosmicray_median(ccd, error_image=None, thresh=5, mbox=11, gbox=0, rbox=0, x
...
+
+    # replace bad pixels in the image
+    ndata = xp.asarray(data, copy=True)
+    if rbox > 0:
+        # Fun fact: scipy.ndimage ignores the mask, so may as well not
+        # bother with it.
+        # data = np.ma.masked_array(data, (crarr == 1))
+
+        # make sure that mdata is the same type as data
+        mdata = xp.asarray(ndimage.median_filter(filt_data, rbox))
```
mwcraig: `rbox > 0` together with a mask is untested. Mutation check: replacing `filt_data` with `data` here survives the whole test file, although the behaviours differ (20σ CR beside a bright masked block, `rbox=11`: PR replaces it with −0.008, the mutant with 1.44, and ~1e4 for a wider box). A test: region fixture + `rbox=11`, assert `|ndata[cr]| < 5σ` and that `ndata[mask] == data[mask]`.

ccdproc/tests/test_cosmicray.py
```diff
@@ -446,16 +453,163 @@ def test_cosmicray_median_ccddata():
...
+    assert not crarr[np_mask].any()
+    # ...and all of the unmasked cosmic rays are still detected
+    for y, x in crrays[n_masked:]:
+        assert crarr[y, x]
+    assert crarr.sum() == NCRAYS - n_masked_unique
+    # masked pixels are returned unchanged
+    assert_allclose(np_array(ndata)[np_mask], np_data[np_mask])
+
+
+def test_cosmicray_median_masked_region_does_not_bias_neighbors():
```
mwcraig: **Mutation testing: the new tests catch almost none of the fix.** Seven mutants of `_cosmicray_median_array` (drop `crarr & ~in_mask`; single-pass fill; `fill=0`; raw `data` in the residual; pass `None` for the CCDData mask; `data` instead of `filt_data` in the `rbox` filter; `|`→`+`) run against the full file on numpy and dask: **only the CCDData-mask mutant is caught**, by one test. The root cause for this test and its CCDData twin is that the data has zero mean, so `fill=0` *is* the mean fill and a single pass gives the same residual (5.51 for all four fill variants at the CR pixel; 4.06 on `main`). Adding a sky level (`loc=1000`) still does not separate them; a sky level **plus** a U-shaped notch (a box > 50 % masked) does — `fill=0` then gives 28 false positives at |r| ≈ 1005 vs 1 for the PR. Suggest: add a sky offset to both region tests, assert `crarr.sum() == 1`, and add the notch geometry. `test_cosmicray_median_ccddata_masked` passes on `main` as well (it only checks the union mask, which `main` already produced) so it guards nothing new; `..._all_masked` does fail on the parent of ce8a923, so that one is real.

ccdproc/tests/test_cosmicray.py
```diff
@@ -50,6 +56,7 @@ def add_cosmicrays(data, scale, threshold, ncrays=NCRAYS):
         y, x = crrays[i]
         data_as_np[y, x] = crflux[i]
     data.data = xp.asarray(data_as_np)
+    return crrays
```
mwcraig, line 61: **None of the strict xfails fail for the stated reason, and the masked code path has zero non-default-device coverage.** With `--runxfail` all eight `cosmicray_median` xfails (five pre-existing, three new) die in this helper, at `np_array(data.data)` five lines up, with `Can't convert array on the 'device1' device to a NumPy array` — `cosmicray_median` is never called. Meanwhile `test_cosmicray_median_masked_region_does_not_bias_neighbors` (no marker) and `..._all_masked` (marked) build `np.ma.array` inputs, so `array_namespace` resolves to numpy and the backend is never exercised; `..._ccddata_masked_region_...` does `xp.asarray(np_data)` without `device=`, so it lands on the CPU device and scipy works. Net: two new markers **XPASS** on strict already, and the other tests aren't testing the backend. Fix: make `add_cosmicrays` device-aware (`xpx.at`, or build on numpy then `xp.asarray(..., device=testing_array_device)`), use `device=testing_array_device` in the CCDData test, build the masked-array tests from `xp` like the rest of the file, and drop the two XPASSing markers per `docs/array_api.rst`.

mwcraig (commented):
Critical review of the diff (algorithm-semantics, numpy-regression, multi-backend, and test-quality passes, every number below re-run locally on the PR head 49cf28e).

**What holds.** No regression for unmasked inputs: 209 input combinations (dtypes × thresh/mbox/gbox/rbox × error_image forms, CCDData with/without uncertainty, memmap, non-contiguous, `nomask`, `Quantity`) are bit-identical to `main` except four intentional mask-honouring differences and a harmless C→K memory-order change. The headline fix for #932 works: masked bright values (bad column at 65535, bright blocks, bright edge rows) no longer leak into the median on a flat background (≤ 0.4σ from an exact masked median, zero mis-flags), masked pixels are never flagged and come back unchanged even with `rbox>0`, all-masked input is clean, NaN under the mask is harmless, jax/dask/numpy give identical CR counts, and the dask escape ratchet is green.

**What doesn't, inline.** Three real algorithm problems: (1) the two-pass fill is biased toward the global mean next to masked regions wider than ~`mbox/2` on a non-flat background — measured false positives at ~15σ and a missed 6σ CR; (2) masked pixels seed `gbox` growth into clean pixels because the mask is applied after `maximum_filter`; (3) `error_image=None` still takes the std over masked pixels, so one masked saturated column turns detection off entirely. Then test coverage: 6 of 7 mutants of the new function survive the whole file, and none of the strict xfails fails for the stated reason (two already XPASS). Plus device/dask notes and a docstring that over-promises.

Housekeeping: the branch is based on a0b6610 and needs a rebase (the strict `test_background_deviation_box` failure is #977's, already fixed on `main`).

mwcraig:
Scoped back down in dc27fbb, per @mwcraig: this PR was meant to be the array-API fix for #932, and the approximate masked median had turned it into an algorithm change. The global-mean fill and second median pass are gone; the median filter runs on the data exactly as on `main`. What remains is the deliberate mask extraction, masked pixels never flagged (now applied before *and* after `gbox` growth, fixing the seeding problem from the review), returned unchanged, and excluded from the `error_image=None` noise estimate. The "does not bias neighbors" tests are replaced by tests of those semantics, which fail if either guard is removed; the two XPASSing strict markers are dropped. Mask-aware median filtering is tracked in #984 with the measurements from the review. The other inline threads are resolved by the removal of the code they refer to; the device/dask notes at the remaining `xp.asarray(ndimage...)` sites are pre-existing on `main` and covered by #935. PR description updated.

PR #987, opened 2026-08-23
Add array-API NaN-aware sum, mean and std fallbacks

### LINKED ISSUE #986: Combiner: no array-API fallback for NaN-aware mean/sum/std (only median has one)
Follow-up to #906 / PR #978, which gave `Combiner.median_combine` a spec-only NaN-aware median (`ccdproc/_nanmedian.py`) for namespaces that do not provide `nanmedian`. Its three siblings did not get the same treatment.

`_default_average`, `_default_sum` and `_default_std` (`ccdproc/combiner.py:44-84`) all still end in

```python
    try:
        return xp.nanmean      # or xp.nansum / xp.nanstd
    except AttributeError as e:
        raise RuntimeError(
            "No NaN-aware mean function available. Please install bottleneck."
        ) from e
```

`nanmean`/`nansum`/`nanstd` are not in the array API standard, so on any conforming-but-minimal namespace the average and sum paths raise instead of computing. The error message is also misleading in that situation: installing bottleneck does not help, because `_default_*` only reaches for bottleneck when the namespace is numpy.

### Impact

On `main` (d85524a) the `array-api-strict` job reports 99 failures. **23 of them are this bug** — the single largest remaining cluster, all in `ccdproc/tests/test_combiner.py`:

`RuntimeError: No NaN-aware mean function available` (14)

- `test_combiner_average`, `test_combiner_mask_average`, `test_combiner_dtype`, `test_combiner_3d`
- `test_1Dweights`, `test_combiner_weighted_average_with_mask_by_pixel`, `test_combiner_weighted_average_with_clipping`, `test_combiner_weighted_average_preserves_custom_scale_func`
- `test_average_combine_uncertainty`, `test_combiner_uncertainty_average`
- `test_combine_result_uncertainty_and_mask[average_combine-True/False]`
- `test_writeable_after_combine[average_combine]`, `test_user_supplied_combine_func_that_relies_on_masks[average_combine]`

`RuntimeError: No NaN-aware sum function available` (9)

- `test_combiner_sum`, `test_combiner_mask_sum`, `test_combiner_sum_weighted`, `test_combiner_sum_weighted_by_pixel`, `test_combiner_sum_weighted_with_mask`
- `test_sum_combine_uncertainty`, `test_combine_result_uncertainty_and_mask[sum_combine-True/False]`, `test_writeable_after_combine[sum_combine]`

No failure is currently attributed to `_default_std`, but that is only because `average_combine` calls `_default_average` (line 687) and `sum_combine` calls `_default_sum` (line 771) *before* either reaches `_default_std` (lines 693 / 774). Fixing mean and sum without std would just move the same `RuntimeError` a few lines down, so **all three need to land together**.

### What is needed

Three fallbacks in the spirit of `_nanmedian.nanmedian`, but much simpler — no sort is required, only masking and counting:

- `nansum(x, axis=0)` — `xp.sum(xp.where(xp.isnan(x), 0, x), axis=axis)`
- `nanmean(x, axis=0)` — the above divided by the count of non-NaN entries
- `nanstd(x, axis=0)` — the usual two-pass or sum-of-squares form over the same count, `ddof=0`

Points to get right, mirroring what `nanmedian` already had to handle:

- **All-NaN slices.** `numpy.nansum` returns `0.0`, while `numpy.nanmean` and `numpy.nanstd` return `NaN` (verified locally; both emit a `RuntimeWarning`, which the fallbacks should *not* do since pytest turns warnings into errors here). Confirm bottleneck agrees before relying on it.
- **Dtype promotion.** Integer/boolean input should be promoted via `xp.__array_namespace_info__().default_dtypes(device=...)["real floating"]` rather than a hardcoded `float64` — jax without `JAX_ENABLE_X64` has no `float64` and warns when asked for one.
- **Device.** Every scalar built inside the fallback needs `device=array_api_compat.device(x)`; the device-mismatch failures in #946 are exactly what happens when that is missed.
- **Axis.** All call sites use `axis=0` only (`combiner.py:687-705`, `771-790`), so matching `nanmedian`'s "single integer axis, `NotImplementedError` otherwise" signature is sufficient.

The error message in the remaining `raise` paths, if any survive, should stop recommending bottleneck for non-numpy namespaces.

### Tests

`ccdproc/tests/test_nanmedian.py` is the model: exercise the fallbacks directly against the numpy reference across dtypes, all-NaN slices, partially-masked slices and non-default devices, independently of whether the active backend happens to provide native versions.

### Notes

Once this lands, the strict job should drop from 99 to roughly 76, and the `Combiner` failures that remain are the `.any()`/`.nbytes` method calls at `combiner.py:517` and `combiner.py:844-859` plus the `astropy.stats` densification tracked in #929.

mwcraig:
Closes #986.

`nansum`/`nanmean`/`nanstd` are not in the array API standard, so on a conforming-but-minimal namespace `_default_average`, `_default_sum` and `_default_std` raised `RuntimeError` instead of computing — and told the user to install bottleneck, which cannot help there, since those helpers only reach for bottleneck when the namespace is numpy. `_default_median` got a spec-only fallback in #906/#978; its three siblings did not.

## What is here

**`ccdproc/_nanfuncs.py`** — spec-only `nansum`, `nanmean` and `nanstd`, and the three helpers wired to them exactly as `_default_median` is wired to `_nanmedian.nanmedian`. All three land together because `average_combine` and `sum_combine` reach mean/sum *before* std, so fixing either alone would only move the error a few lines down.

The awkward parts, all of which `nanmedian` had to handle too:

- **All-NaN slices** match numpy in value — `nansum` gives 0.0, `nanmean` and `nanstd` give NaN — but do it *silently*. `numpy.nanmean` and `numpy.nanstd` emit a `RuntimeWarning` there, and this project's pytest configuration turns warnings into errors, so a warning would fail every `Combiner` test with a fully masked pixel. Zero denominators are swapped for ones *before* the division rather than patched up after, so no `0/0` is ever evaluated.
- **Dtype promotion** goes through `xp.__array_namespace_info__().default_dtypes(device=...)["real floating"]` rather than a hardcoded float64, since jax without `JAX_ENABLE_X64` has no float64 and warns when asked for one. This is where `nansum` deliberately parts company with `numpy.nansum`, which preserves an integer dtype.
- **Device**: every scalar built inside the fallbacks carries `device=array_api_compat.device(x)`.
- **Axis**: a single integer, `NotImplementedError` otherwise — the same contract `nanmedian` offers, and all call sites use `axis=0`.
- `nanstd` is the two-pass form (`ddof=0`, matching `numpy.nanstd` and `bottleneck.nanstd`). A test with values `1e8 + {0,1,2,3}` pins it there: the single-pass `sum(x**2) - sum(x)**2/n` form loses every significant digit at CCD-count scale.

**`ccdproc/tests/test_nanfuncs.py`** — modelled on `test_nanmedian.py`: all three fallbacks against the numpy reference across dtypes (float, int, bool), 1-D/2-D/3-D, all-NaN and single-value slices, negative and non-zero axes, and the non-default device, independently of whether the active backend happens to provide native versions.

`test_bottleneck_defaults_respect_array_namespace` is parametrized off a `_DEFAULT_FUNCS` table naming each helper's fallback, and a new `test_defaults_fall_back_without_native_nan_function` covers all four against a stand-in namespace.

**`tox.ini`** — a `strict` environment so the array-api-strict suite can be run locally. It pins only the interpreter and inherits `setenv`, `deps`, `extras` and `commands` through the existing `strict` factor, so it stays in lockstep with CI by construction; `tox -e strict --showconfig` is identical to `py313-strict` apart from the env name. The pin is the point — without a `py3xx` factor in the name, tox builds the env against whatever interpreter tox itself is installed under, which is often too old.

## Verification

- numpy: 492 passed. dask: 481 passed. No regressions.
- array-api-strict (`tox -e strict`): no `No NaN-aware ...` errors remain; 266 → 330 passed.

## This does not, by itself, lower the strict failure count

It stays at exactly 99. #986 predicted 99 → ~76, but all 23 of the tests it named now get *past* the reduction and fail one frame later:

| next failure | count |
|---|---|
| `'Array' object has no attribute 'any'` (`combiner.py:517`, `_get_nan_substituted_data`) | 21 |
| `'Array' object has no attribute 'sum'` | 2 |
| `dtype must be one of the supported dtypes, got <class 'bool'>` | 1 |

That `.any()` wall already blocked six `median_combine` tests on `main`, so it is independent of this change — #986's estimate double-counted a blocker it had itself listed under "what remains". The fallbacks are still the necessary fix; they bring average/sum/std to parity with median. Fixing `combiner.py:517`/`:536` is now the highest-leverage strict change, with roughly 23 tests queued behind it.


ccdproc/_nanfuncs.py
```diff
@@ -0,0 +1,191 @@
...
+    without triggering a promotion the namespace might not allow.
+    """
+    isnan = xp.isnan(x)
+    zero = xp.asarray(0, dtype=x.dtype, device=device)
+    total = xp.sum(xp.where(isnan, zero, x), axis=axis, keepdims=keepdims)
+    count = xp.sum(xp.astype(~isnan, x.dtype), axis=axis, keepdims=keepdims)
+    return total, count
+
+
+def _safe_divide(total, count, xp, device):
```
mwcraig, line 129: I understand what this does, but why not simply filter out the warning that we know can occur when there is a divide by zero?

ccdproc/_nanfuncs.py
```diff
@@ -0,0 +1,191 @@
...
+        raise ValueError(f"axis {axis} is out of bounds for array of dimension {ndim}")
+    axis = axis % ndim
+
+    device = array_api_compat.device(x)
+
+    if not xp.isdtype(x.dtype, "real floating"):
+        # Promote to the namespace's default real dtype rather than hardcoding
+        # float64: jax without JAX_ENABLE_X64 has no float64 and warns when one
+        # is requested, which pytest's filterwarnings turns into an error.
+        info = xp.__array_namespace_info__()
```
mwcraig, line 88: I do not like this....is this part of the actual array API or did you come up with it by reading its source code?

ccdproc/_nanfuncs.py
```diff
@@ -0,0 +1,191 @@
...
+`ccdproc._nanmedian.nanmedian`.
+"""
+
+import array_api_compat
+
+__all__ = ["nanmean", "nanstd", "nansum"]
+
+
+def _setup(x, axis, xp):
+    """
```
mwcraig, line 27: Make this a real docstring (numpy-style)

ccdproc/_nanfuncs.py
```diff
@@ -0,0 +1,191 @@
...
+        # float64: jax without JAX_ENABLE_X64 has no float64 and warns when one
+        # is requested, which pytest's filterwarnings turns into an error.
+        info = xp.__array_namespace_info__()
+        x = xp.astype(x, info.default_dtypes(device=device)["real floating"])
+
+    return x, axis, xp, device
+
+
+def _sum_and_count(x, axis, xp, device, *, keepdims):
+    """
```
mwcraig, line 95: make this a real docstring

ccdproc/_nanfuncs.py
```diff
@@ -0,0 +1,191 @@
...
+    isnan = xp.isnan(x)
+    zero = xp.asarray(0, dtype=x.dtype, device=device)
+    total = xp.sum(xp.where(isnan, zero, x), axis=axis, keepdims=keepdims)
+    count = xp.sum(xp.astype(~isnan, x.dtype), axis=axis, keepdims=keepdims)
+    return total, count
+
+
+def _safe_divide(total, count, xp, device):
+    """
```
mwcraig, line 130: Make this a real docstring if the function stays around

ccdproc/_nanfuncs.py
```diff
@@ -0,0 +1,191 @@
...
+        Axis along which to average. Default is 0. ``None`` and tuples of
+        axes are not supported.
+    xp : array namespace, optional
+        Namespace to use. Defaults to ``array_api_compat.array_namespace(x)``.
+
+    Returns
+    -------
+    array
+        Mean of ``x`` along ``axis``, with that axis removed. Slices that are
+        entirely NaN yield NaN, matching `numpy.nanmean` -- but silently,
```
mwcraig: Why is silence a design goal?

ccdproc/_nanfuncs.py
```diff
@@ -0,0 +1,191 @@
...
+        but silently, where numpy warns. A slice with a single non-NaN entry
+        yields zero.
+
+    Notes
+    -----
+    This is the two-pass form: the mean is computed first and then subtracted,
+    rather than accumulating ``sum(x**2) - sum(x)**2 / n``. The extra pass
+    costs one more reduction but avoids the catastrophic cancellation the
+    single-pass form suffers when the values are large relative to their
+    spread, which is the normal situation for CCD counts.
```
mwcraig: Not sure I would call it normal but certainly not unusual.

CHANGES.rst
```diff
@@ -6,6 +6,9 @@ New Features
 
 - Add ``array-api-strict`` (on a non-default device) as a test backend, a
   CPU-only proxy for CuPy device behavior. [#942]
+- Add a ``strict`` tox environment for running the ``array-api-strict`` test
```
mwcraig: This is not user facing -- why add it?

CHANGES.rst
```diff
@@ -28,6 +31,12 @@ Bug Fixes
 - Add a NaN-aware median written purely in terms of the array API standard
   and use it in ``Combiner.median_combine`` when the selected array namespace
   has no ``nanmedian``, instead of raising ``RuntimeError``. [#906]
+- Add NaN-aware sum, mean and standard deviation written purely in terms of
+  the array API standard and use them in ``Combiner.average_combine`` and
+  ``Combiner.sum_combine`` when the selected array namespace has no
+  ``nansum``/``nanmean``/``nanstd``, instead of raising a ``RuntimeError``
```
mwcraig: The part starting with "instead" is unnecessary

ccdproc/_nanfuncs.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def _setup(x, axis, xp):
+    """
+    Validate ``axis``, resolve the namespace and device, promote to float.
+
+    Returns ``(x, axis, xp, device)`` with ``axis`` normalised to a
+    non-negative integer and ``x`` guaranteed to have a real floating dtype.
+    """
+    if axis is None or not isinstance(axis, int):
```
mwcraig: `isinstance(axis, int)` has two quirks, both inherited verbatim from `_nanmedian`: it rejects `np.int64(0)` with `NotImplementedError` (numpy integers are not `int`), yet silently accepts `axis=True` as axis 1, because `bool` subclasses `int`. All ccdproc call sites pass a literal `0`, and consistency with `nanmedian` is a fair defense — but if you touch this, `operator.index(axis)` plus an explicit `bool` rejection fixes both, ideally in both modules.

ccdproc/_nanfuncs.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    NaN-free sum along ``axis`` and the number of non-NaN entries in it.
+
+    ``count`` has the same dtype as ``x`` so that it can divide the sum
+    without triggering a promotion the namespace might not allow.
+    """
+    isnan = xp.isnan(x)
+    zero = xp.asarray(0, dtype=x.dtype, device=device)
+    total = xp.sum(xp.where(isnan, zero, x), axis=axis, keepdims=keepdims)
+    count = xp.sum(xp.astype(~isnan, x.dtype), axis=axis, keepdims=keepdims)
```
mwcraig, line 125: Worth a one-line comment, not a change: because `count` is accumulated in `x.dtype`, a float32 count is exact only up to 2^24 non-NaN elements per slice. The combiner reduces over the image axis so this is unreachable in practice, but the constraint is invisible in the code.

ccdproc/tests/test_nanfuncs.py
```diff
@@ -0,0 +1,135 @@
...
+from ccdproc.conftest import testing_array_library as xp
+
+_rng = np.random.default_rng(986)
+_some_nan = _rng.normal(size=(4, 3))
+_some_nan[[0, 1, 2, 3], [1, 2, 0, 2]] = np.nan
+
+# Values large compared with their spread: the single-pass
+# ``sum(x**2) - sum(x)**2 / n`` form of the variance loses every significant
+# digit here, so this pins ``nanstd`` to the two-pass form it actually uses.
+_ill_conditioned = np.array([1e8, 1e8 + 1.0, 1e8 + 2.0, 1e8 + 3.0])
```
mwcraig, line 19: This row assumes float64. Under a float32-default backend (jax without `JAX_ENABLE_X64`), `1e8 + 1` collapses to `1e8` on conversion, the fallback correctly returns std 0, and the float64 numpy reference says ~1.118 — a hard test failure, not a precision degradation. CI sets `JAX_ENABLE_X64 = True` so it's latent, but a comment or a skip-guard on the backend's default real dtype would save the next person a confusing local failure.

ccdproc/tests/test_nanfuncs.py
```diff
@@ -0,0 +1,135 @@
...
+
+_FUNCS = [
```
mwcraig, line 21: The bigger consolidation lever: add `pytest.param(nanmedian, np.nanmedian, id="nanmedian")` here, take the union of this `_DATA` table with the one in `test_nanmedian.py` (its `range(1, 7)` odd/even rows subsume the `range(1, 4)` rows here), and delete `test_nanmedian.py`. The two `bad_axis` tests are already line-for-line identical modulo the function. Net: one ~85-line file covering all four fallbacks instead of 185 lines across two. Touching a merged test file may be follow-up material, but "modelled on test_nanmedian.py" is the argument for doing it now.

ccdproc/tests/test_nanfuncs.py
```diff
@@ -0,0 +1,135 @@
...
+    (np.array([[1.0, np.nan], [np.nan, np.nan]]), 0),  # single non-NaN in a slice
+    (np.array([np.nan, np.nan, np.nan]), 0),  # every value NaN
+    (_ill_conditioned, 0),
+    (np.array([[1, 4], [2, 3], [5, 6], [4, 1]]), 0),  # integer input
+    (np.array([[True, False], [False, True], [True, True]]), 0),  # boolean input
+]
+
+
+@contextmanager
+def _numpy_reference():
```
mwcraig: This 17-line contextmanager is used once; `test_nanmedian.py` handles the same problem with a one-line `@pytest.mark.filterwarnings("ignore:...")` on the test. Here that's two ignore marks (`"ignore:Mean of empty slice"`, `"ignore:Degrees of freedom <= 0"`) on `test_matches_numpy`, saving ~14 lines. The marks would also mask hypothetical warnings from the fallback call in that one test — which is exactly why `test_no_warning_on_all_nan_slice` is the extra test worth *keeping*: it becomes the sole, explicit owner of the silent-on-all-NaN property.

ccdproc/tests/test_nanfuncs.py
```diff
@@ -0,0 +1,135 @@
...
+    data = xp.asarray(
+        np.array([[1.0, np.nan], [2.0, np.nan], [3.0, np.nan]]), device=xp_device
+    )
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        func(data, axis=0)
+
+
+def test_nansum_all_nan_slice_is_zero():
```
mwcraig, line 95: This test and `test_mean_and_std_all_nan_slice_is_nan` below (~20 lines together) are strictly implied by `test_matches_numpy`: `_DATA` already contains an all-NaN column, an all-NaN 1-D array, and a single-survivor slice, and `xpx.isclose(..., equal_nan=True)` verifies NaN *positions* as well as neighboring values. If `nansum` returned NaN instead of 0, or `nanmean` corrupted a neighbor, `test_matches_numpy` fails on those exact rows. Suggest dropping both.

ccdproc/tests/test_nanfuncs.py
```diff
@@ -0,0 +1,135 @@
...
+    """An all-NaN slice yields NaN, and leaves its neighbours alone."""
+    data = xp.asarray(np.array([[1.0, np.nan], [2.0, np.nan]]), device=xp_device)
+
+    result = func(data, axis=0)
+
+    assert not xp.any(xp.isnan(result[0:1]))
+    assert xp.all(xp.isnan(result[1:2]))
+
+
+def test_nanstd_is_population_deviation():
```
mwcraig: Also implied by `test_matches_numpy`: `numpy.nanstd`'s default is `ddof=0`, so any other ddof fails the differential test on every multi-element float case. The docstring's point (matches numpy and bottleneck) could move to a comment on the `nanstd` row of `_FUNCS`. Another ~7 lines.

ccdproc/tests/test_combiner.py
```diff
@@ -125,27 +132,30 @@ def test_bottleneck_defaults_respect_array_namespace(default_func, function_name
...
     else:
-        # The namespace has no such function -- array-api-strict has no
-        # nanmedian, for one. Only the median has a spec-only fallback; the
-        # others raise RuntimeError inside default_func above, so in practice
-        # only nanmedian reaches here.
+        # The namespace has no such function -- array-api-strict has none of
+        # the four -- so the spec-only fallback stands in, bound to it.
         assert isinstance(default, partial)
-        assert default.func is nanmedian
+        assert default.func is fallback
```
mwcraig, line 139: This extended fallback branch is unreachable in CI: the test starts with `pytest.importorskip("bottleneck")`, and the only CI backend lacking the native functions is array-api-strict — whose env deliberately omits bottleneck, so the whole test skips there (confirmed: 4 skipped in a strict run). The branch only runs if someone hand-installs bottleneck into a strict env. Since `test_defaults_fall_back_without_native_nan_function` already covers fallback selection on every backend via the stand-in namespace, either drop the importorskip (compute `expected` from bottleneck availability) or shrink this branch to a comment pointing at the stand-in test.

ccdproc/tests/test_combiner.py
```diff
@@ -125,27 +132,30 @@ def test_bottleneck_defaults_respect_array_namespace(default_func, function_name
...
 
 
-def test_default_median_falls_back_without_nanmedian():
-    # A namespace with no nanmedian (it is not in the array API standard)
-    # gets the spec-only fallback, bound to that namespace. Use a stand-in
-    # namespace so this is exercised regardless of which backend is under
-    # test; every backend in CI happens to provide nanmedian.
+@pytest.mark.parametrize(
+    ("default_func", "fallback"),
+    [(func, fallback) for func, _, fallback in _DEFAULT_FUNCS],
```
mwcraig: Parametrizing directly over the triple with an unused middle arg drops four lines:

```python
@pytest.mark.parametrize(("default_func", "_name", "fallback"), _DEFAULT_FUNCS)
def test_defaults_fall_back_without_native_nan_function(default_func, _name, fallback):
```

mwcraig (commented):
Critical review focused on (a) correctness of the new fallbacks and (b) test size.

**Verdict: the fallbacks are correct.** A differential fuzz against numpy (27,000 cases: float32/float64, 1-D through 4-D, negative axes, NaN densities 0/0.3/1.0) found zero mismatches, and the new tests pass on numpy, dask, and array-api-strict (65 passed each). The combiner failures in a full strict run are the pre-existing `.any()` wall at `combiner.py:518` that the PR description already documents; this change does not touch them.

The main opportunity is test size: roughly 50 of the 135 lines in `test_nanfuncs.py` restate coverage that `test_matches_numpy` already provides, and merging with `test_nanmedian.py` would save ~95 lines total. Details inline.

**Non-issues I checked deliberately:**
- All combiner call sites pass `axis=0` by keyword, so the keyword-only fallback signature is compatible with every path (`average_combine`, `sum_combine`, `_weighted_sum`, uncertainty).
- `_safe_divide` genuinely never evaluates 0/0: with `warnings.simplefilter("error")` active, none of the three fallbacks warn on an all-NaN slice.
- Slices containing `inf` match `numpy.nanstd` (both give NaN).
- float32 input stays float32; only int/bool promote.
- The int64-precision divergence from `numpy.nansum` (e.g. `2**53 + 1` loses a bit on promotion to float64) is real but exactly what the module docstring discloses.

One nit not worth an inline thread: the 18-line comment on `[testenv:strict]` in `tox.ini` could lose its last paragraph — the `changedir` comment it cites already explains the no-`test`-factor rule.

ccdproc/_nanfuncs.py
mwcraig: A filter can't be scoped correctly on every backend, so it would only fix numpy. Two verified data points from this branch's own envs:

- **dask**: the division is lazy, so wrapping it in `warnings.catch_warnings()` + ignore does nothing — I ran exactly that in `.tox/py312-alldeps-dask`, and the `RuntimeWarning: invalid value encountered in divide` fired later, at `.compute()` time, *outside* the block, where this project's `filterwarnings = error` turns it into an error in whatever test happens to trigger the compute.
- **jax**: `0/0` is already silent (returns NaN, no warning), so the filter is dead code there.

So a message-matched filter helps only eager numpy-backed namespaces, while `catch_warnings` also mutates process-global state (not thread-safe). Never evaluating `0/0` at all is the one variant that behaves identically on every conforming namespace, and it costs two `xp.where` calls.

ccdproc/_nanfuncs.py
mwcraig: It is part of the standard, not something scraped from an implementation: the inspection API (`xp.__array_namespace_info__()` and its `default_dtypes(device=...)` method, returning a dict keyed by `"real floating"`, `"complex floating"`, `"integral"`, `"indexing"`) was added in the **2023.12** revision:

- https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.__array_namespace_info__.html
- https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.default_dtypes.html

Verified locally that `array-api-strict` implements it (`{'real floating': float64, ...}`), and `array_api_compat` backfills it for the wrapped namespaces. It's the standard's only portable answer to "what does float mean here" — hardcoding float64 breaks jax without `JAX_ENABLE_X64`.

ccdproc/_nanfuncs.py
mwcraig: Agreed — will expand `_setup`'s docstring to numpy style (Parameters/Returns/Raises) in the next push.

ccdproc/_nanfuncs.py
mwcraig: Agreed — will convert `_sum_and_count`'s docstring to numpy style in the next push.

ccdproc/_nanfuncs.py
mwcraig: It stays (see the thread at line 72 for why filtering the warning instead doesn't survive dask's laziness), so agreed — numpy-style docstring in the next push.

ccdproc/_nanfuncs.py
mwcraig: Silence isn't the goal — consistency with every other path through the combiner is. Verified: `bottleneck.nanmean`/`nanstd` (what the numpy default actually uses) return NaN on an all-NaN slice with no warning, jax's native `nanmean` is silent, and the `_nanmedian` fallback is silent. numpy's nan-reductions are the odd ones out. An all-NaN slice is also a routine state here — it's just a fully masked pixel — not an anomaly worth a per-call warning; and with pytest's warnings-as-errors, a warning would fail every `Combiner` test containing one.

Suggest rewording the docstring so it carries that justification instead of just asserting the behavior, e.g.: "Slices that are entirely NaN yield NaN silently, matching ``bottleneck.nanmean`` (the numpy-backend default); ``numpy.nanmean`` warns here, but a fully masked pixel is a routine input for the combiner, not an anomaly." Happy to apply.

ccdproc/_nanfuncs.py
mwcraig: Fair — will change "which is the normal situation for CCD counts" to "which is not unusual for CCD counts" in the next push.

CHANGES.rst
mwcraig: Agreed it's contributor-facing, not user-facing — will drop it. For the record, the adjacent triage-tooling entry [#942] is in the same category; astropy core would put both under "Other Changes and Additions" rather than "New Features". If you'd rather keep a trace of infra changes, moving both under that heading is the alternative; otherwise the tox.ini comment block documents the env well enough on its own.

CHANGES.rst
mwcraig: Half agree. The sibling #906 entry just above ends "instead of raising ``RuntimeError``" — a bug-fix entry is stronger for saying what the broken behavior was, and those five words do it. The rest ("that recommended installing ``bottleneck`` even where ``bottleneck`` could not have helped") is elaboration that belongs in the PR description, not the changelog. Proposal: trim to exactly the #906 form —

```rst
  ``nansum``/``nanmean``/``nanstd``, instead of raising ``RuntimeError``. [#986]
```

Will apply that unless you want the clause gone entirely.

ccdproc/_nanfuncs.py
mwcraig: Both quirks confirmed locally (`isinstance(np.int64(0), int)` is `False`; `True % 2 == 1`, so `axis=True` silently reduces axis 1). Since the bool half is silent wrongness rather than a loud refusal, worth fixing now — will apply the `operator.index(axis)` + explicit `bool` rejection to `_setup` here **and** to `_nanmedian` in the same commit so the two modules keep the identical contract:

```python
if axis is None or isinstance(axis, bool):
    raise NotImplementedError(...)
try:
    axis = operator.index(axis)
except TypeError:
    raise NotImplementedError(...)
```

ccdproc/_nanfuncs.py
mwcraig: Agreed — will add the one-line comment (float32 count is exact only up to 2^24 non-NaN entries per slice, unreachable when reducing over the image axis) in the next push.

ccdproc/tests/test_nanfuncs.py
mwcraig: Confirmed latent (float32 spacing at 1e8 is 8, so `1e8 + {1,2,3}` all collapse to `1e8` on conversion and the fallback's correct answer is 0 while the float64 reference says ~1.118). A comment doesn't stop the confusing failure, so will add a runtime skip-guard instead: in `test_matches_numpy`, skip the row when the converted values can't be told apart, e.g.

```python
converted = xp.asarray(data, device=xp_device)
if data is _ill_conditioned and bool(xp.all(converted == converted[0])):
    pytest.skip("backend's default real dtype cannot resolve these values")
```

so a float32-default backend skips with an explanation rather than failing.

ccdproc/tests/test_nanfuncs.py
mwcraig: The union math checks out (verified: `test_nanmedian.py`'s `range(1, 7)` rows subsume the `range(1, 4)` rows here, and its remaining rows are all present in this `_DATA` already), so the merge is mechanically safe. Recommending against it anyway: the test files map 1:1 onto the modules (`test_nanmedian.py` ↔ `_nanmedian.py`, this file ↔ `_nanfuncs.py`), and a merged file couples `_nanmedian`'s coverage to this module's scaffolding for a ~95-line saving. The genuinely duplicated part is only the two `bad_axis` tests plus table scaffolding, ~25 lines. Happy to do the merge if you weigh the line count higher than the file↔module mapping — say the word.

ccdproc/tests/test_nanfuncs.py
mwcraig: Agreed — and it's also the more consistent choice: `test_nanmedian.py` already handles the same problem with a `@pytest.mark.filterwarnings("ignore:All-NaN slice encountered...")` mark. Will replace the contextmanager with

```python
@pytest.mark.filterwarnings("ignore:Mean of empty slice:RuntimeWarning")
@pytest.mark.filterwarnings("ignore:Degrees of freedom <= 0:RuntimeWarning")
```

on `test_matches_numpy`, with `test_no_warning_on_all_nan_slice` remaining the explicit owner of the silent-on-all-NaN property.

ccdproc/tests/test_nanfuncs.py
mwcraig: Agree on `test_mean_and_std_all_nan_slice_is_nan` — fully implied (the all-NaN column row plus `equal_nan=True` closeness checks both the NaN position and the untouched neighbour), will drop it. Arguing to keep `test_nansum_all_nan_slice_is_zero`, though: it's the one genuinely counterintuitive value in the module (sum of nothing is 0, not NaN), and the differential test pins it only *relative to numpy* — nothing in `test_matches_numpy`'s name or failure output states the contract, and an edit to `_DATA` could silently drop the covering row. Seven lines is cheap for an absolute pin of the surprising case. Net: −13 of the ~20 lines.

ccdproc/tests/test_nanfuncs.py
mwcraig: Agreed — every multi-element float row in the differential test already fails for any ddof other than numpy's default 0 (and verified `bottleneck.nanstd` defaults to ddof=0 too, so the docstring's claim survives as a comment). Will drop the test and move the note onto the `nanstd` row of `_FUNCS`.

ccdproc/tests/test_combiner.py
mwcraig: Confirmed dead in CI as written. Going with the first option, in its minimal form: move the `importorskip` into the only branch that needs bottleneck —

```python
if array_api_compat.is_numpy_namespace(xp):
    bottleneck = pytest.importorskip("bottleneck")
    expected = getattr(bottleneck, function_name)
```

so on the strict job the fallback branch actually executes instead of skipping, and the test still skips only where its premise (numpy + bottleneck) is unavailable. (numpy-without-bottleneck would return `xp.nanmean` etc. from the hooks — a real case, but a different test's job.)

ccdproc/tests/test_combiner.py
mwcraig: Agreed — will parametrize over the triple with an unused `_name` middle argument as suggested.

mwcraig:
Responding to the [review summary](https://github.com/astropy/ccdproc/pull/987#pullrequestreview-5003388709):

On the one summary-level nit — the last paragraph of the `[testenv:strict]` comment in `tox.ini`: half agree. It is a cross-reference, not a duplication — the reader it protects is the one about to rename the env, and without a pointer *at the rename site* the explanation in `[testenv]` never reaches them (adding a `test` factor would silently move the run into `.tmp/{envname}` and break the `__file__`-anchored escapes baseline). But it doesn't need three lines; will compress to one:

```ini
# No "test" factor in the name on purpose -- see the changedir comment in [testenv].
```

On the test-size theme: agreed in the inline threads to drop `test_mean_and_std_all_nan_slice_is_nan` and `test_nanstd_is_population_deviation`, replace the 17-line contextmanager with two `filterwarnings` marks, and tighten the `test_combiner.py` parametrization; pushed back on merging with `test_nanmedian.py` (file↔module mapping) and on dropping `test_nansum_all_nan_slice_is_zero` (the one counterintuitive absolute value deserves its own pin). Details in each thread.

ccdproc/tests/test_nanfuncs.py
mwcraig: Why aren't the nanfuncs all in one file, anyway? That would make more sense to me.

ccdproc/_nanfuncs.py
mwcraig: ok, keep it

CHANGES.rst
mwcraig: Move both there

ccdproc/tests/test_nanfuncs.py
mwcraig: Sequencing, not design: `_nanmedian.py` landed first as a self-contained module in #906, and this PR added the three siblings without touching it to keep the diff reviewable. There's no technical reason for the split — and one concrete reason to end it: `nanmedian` duplicates the axis-validation/promotion preamble that `_setup` now owns (~25 lines that just had to be fixed in both places for the `operator.index` change). Merging `nanmedian` into `_nanfuncs.py` would let it reuse `_setup`, and the test files would then merge naturally too — which resolves the consolidation question above in the direction you're leaning. Happy to do the module+test merge in this PR or as a follow-up; holding off until you pick.

mwcraig: Let's do it here

Issue #946, opened 2026-07-14
array-api-strict job: device-mismatch noise from inline array creation in test bodies


mwcraig:
Follow-up from PR #942 (review item D3, https://github.com/astropy/ccdproc/pull/942#pullrequestreview-4695447403).

The strict CI job (`ubuntu-py313-strict`) creates fixture data on array-api-strict's non-default device (`device1`, set up in `ccdproc/tests/pytest_fixtures.py`) to emulate CuPy-on-GPU behavior. But many test bodies create arrays inline with `xp.ones(...)` / `xp.asarray(...)` and no `device=` argument, so those arrays land on the *default* device. array-api-strict raises on any cross-device operation, so a chunk of the strict job's failure list is harness artifacts (fixture-on-device1 vs inline-array-on-default), not real ccdproc bugs.

Fix sketch: provide a fixture-supplied `make_array` helper (or similar) that pins the testing device, then migrate test bodies to use it instead of raw `xp.*` constructors.

Worth doing *before* burning down the strict failure list (#941), so the remaining failures are all genuine array-API escapes.

mwcraig:
Closing, as recorded in #971: after the section-1 `device=` fixes (#993, #995) and the section-2 test-body pass (#998), none of the 15 strict failures left on `main` (`4d7d21b`) is a test-body device mismatch. The remaining device errors all sit in astropy setters/`Quantity` (#936, #983) or `astropy.stats` (#929), which have their own issues.

— _Written by Claude at @mwcraig's direction._

Issue #929, opened 2026-07-07
Array API: Combiner.sigma_clipping and the default sigma_func densify via astropy.stats


mwcraig:
### Problem

Two closely related densification points in the combining pipeline:

1. `Combiner.sigma_clipping` delegates to `astropy.stats.sigma_clip`, which internally does `np.asanyarray` and works with `np.ma` masked arrays — both densify non-numpy input:

   https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/combiner.py#L449

2. `ccdproc.sigma_func` — the **default** deviation function for `median_combine` uncertainty — is built on `astropy.stats.median_absolute_deviation`, which is likewise numpy-only:

   https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/core.py#L1233

These likely account for a substantial share of the `Implicit conversion to a NumPy array is not allowed` failures in #911.

### Options

- Provide xp-native implementations in ccdproc: an iterative sigma-clip using `xp.where`, and a MAD built on the fallback median discussed in #906.
- Or wait for / contribute array-API support in `astropy.stats` (tracked in #940).

---
Found during a review of the array API implementation from #885; follow-up to #909 / #911. Related: #906.

PR #988, opened 2026-08-23
Make combiner internals array-API compatible

mwcraig:
Part of #971. This addresses three unnumbered checklist items from the array-API tracker, all in the combiner internals, plus the immediately adjacent numpy-isms those fixes exposed in the same code paths.

### The fixes

1. **`Combiner._get_nan_substituted_data` used `mask.any()`** — a numpy-only array method. Now `xp.any(...)` in the combiner's namespace. Fixing this exposed the next layer in the same functions, which is also fixed here: `.sum(axis=0)` method calls (now `xp.count_nonzero`, since the array API also disallows summing booleans), `len(<array>)` (now `.shape[0]`; `__len__` is not required by the standard), a list passed to `xp.reshape` (now a tuple), and integer counts fed to `xp.sqrt`/uncertainty scaling (now cast to `float64`, matching the promotion numpy does implicitly).
2. **Builtin `bool` passed as a dtype** in the `_CCDDataWrapperForArrayAPI` mask setters and in test bodies that build masks in the testing namespace — strict namespaces only accept their own dtype objects. Now `xp.bool`.
3. **`combine()` memory sizing used `.nbytes`**, which is not in the array API standard. A new `_array_size_in_bytes()` helper computes size from `array_api_compat.size(arr)` and the dtype bit width from `xp.finfo`/`xp.iinfo` (booleans counted as 1 byte, complex as two components). The stale `backend_xfail` on `test_combine_scale_callable_returning_backend_scalar` is removed and a direct unit test of `_calculate_size_of_image` is added.

### Strict suite: 99 failed / 348 passed → 85 failed / 364 passed

(`CCDPROC_ARRAY_LIBRARY=array-api-strict`, on top of #987.)

Causes eliminated: the 28 `.any()` failures, the 9 builtin-`bool` dtype failures, the 2 `.nbytes` failures, and the intermediate layers behind them (`.sum` method, `len()`, list reshape, int→float promotion). One previously xfailed test now passes and its marker is removed; three `test_cosmicray` parametrizations now xpass as a side effect of the wrapper mask-setter fix (their marker still guards other parametrizations, so it stays).

Remaining failures are known separate issues, left for follow-up: device handling (#946, now the largest bucket at 31 since more tests reach the device checks), `xp.median` in `core.py` (4), fancy-index `__setitem__` in `clip_extrema` (4), sigma clipping via `astropy.stats` (#929), `combine()` file-input handling, CCDData arithmetic converting arrays inside `astropy.nddata` (the "Could not convert Array" bucket, hit by the scaling tests), and numpy-isms in test bodies (`.mean()` method calls, mixed int/float test arithmetic, dtype comparisons).

`test_combiner.py` passes fully on numpy (89 passed) and dask (93 passed); the full numpy suite is green (501 passed).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,13 +39,9 @@ Bug Fixes
   the array API standard and use them in ``Combiner.average_combine`` and
   ``Combiner.sum_combine`` when the selected array namespace has no
   ``nansum``/``nanmean``/``nanstd``, instead of raising ``RuntimeError``. [#986]
-- Make the ``Combiner`` combination internals array-API compatible: masked
-  entries are detected and counted with ``xp.any``/``xp.count_nonzero``
-  instead of numpy-only array methods, image counts no longer rely on arrays
-  implementing ``len()`` and are promoted to floating point before entering
-  the uncertainty calculation, mask dtypes use the namespace's ``bool``, and
-  ``combine()`` sizes images for ``mem_limit`` from the element count and
-  dtype width instead of the non-standard ``nbytes``. [#988]
+- Make the ``Combiner`` combination internals array-API compatible and size
+  images for ``mem_limit`` from the element count and dtype width instead of
+  the non-standard ``nbytes``. [#988]
 - Make ``flat_correct`` and ``ccdmask`` use functional array updates so they
   support immutable array-API backends. [#956]
 - Build the ``Combiner`` data and mask arrays with ``xp.stack`` instead of
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,6 +39,13 @@ Bug Fixes
   the array API standard and use them in ``Combiner.average_combine`` and
   ``Combiner.sum_combine`` when the selected array namespace has no
   ``nansum``/``nanmean``/``nanstd``, instead of raising ``RuntimeError``. [#986]
+- Make the ``Combiner`` combination internals array-API compatible: masked
+  entries are detected and counted with ``xp.any``/``xp.count_nonzero``
+  instead of numpy-only array methods, image counts no longer rely on arrays
+  implementing ``len()`` and are promoted to floating point before entering
+  the uncertainty calculation, mask dtypes use the namespace's ``bool``, and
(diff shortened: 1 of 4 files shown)


ccdproc/combiner.py
```diff
@@ -836,28 +842,45 @@ def _calculate_step_sizes(x_size, y_size, num_chunks):
     return xstep, ystep
 
 
+def _array_size_in_bytes(arr):
```
mwcraig, line 845: Does array_api_compat offer an alternative to this?

CHANGES.rst
```diff
@@ -39,6 +39,13 @@ Bug Fixes
   the array API standard and use them in ``Combiner.average_combine`` and
   ``Combiner.sum_combine`` when the selected array namespace has no
   ``nansum``/``nanmean``/``nanstd``, instead of raising ``RuntimeError``. [#986]
+- Make the ``Combiner`` combination internals array-API compatible: masked
```
mwcraig: Shorten this

ccdproc/combiner.py
mwcraig: Checked: no. `array_api_compat` 1.15.0 offers only `size()` — the element count, which this helper already uses — and nothing for byte size; on `array-api-strict` the array has no `.nbytes` and the dtype has no `.itemsize`, so there is nothing portable to delegate to. The `finfo(...)/iinfo(...).bits` attributes are the only dtype-width source the standard guarantees, which is why the helper dispatches on dtype kind. If `array_api_compat` grows an `nbytes` helper later, this function reduces to one line.

CHANGES.rst
mwcraig: Agreed — the mechanism list belongs in the PR, not the changelog. Proposed replacement for the whole entry:

```rst
- Make the ``Combiner`` combination internals array-API compatible and size
  images for ``mem_limit`` from the element count and dtype width instead of
  the non-standard ``nbytes``. [#988]
```

Happy to push that.

mwcraig: Pushed in 3e644d8.

PR #989, opened 2026-08-24
Add an array-API median fallback for subtract_overscan

mwcraig:
## What

`subtract_overscan` calls `xp.median(overscan.data, axis=overscan_axis)`
(`ccdproc/core.py`), but `median` is not part of the array API standard, so
it raises `AttributeError` on backends such as `array-api-strict` that don't
provide it (numpy, jax and dask all have a native `median` and are
unaffected).

This adds `ccdproc._nanfuncs.median`, a small wrapper around the existing
`nanmedian` fallback, and uses it from a new `_median_fallback` helper in
`core.py` that tries the namespace's own `median` first and only falls back
to the array-API-only implementation when it's missing -- mirroring the
existing `_percentile_fallback` pattern already used for `percentile`.

## Why

Part of #971 (array-API migration tracker). `median` joins `nansum` /
`nanmean` / `nanstd` / `nanmedian` (#906, #986) as another reduction that
isn't part of the array API standard and needs a pure-array-API fallback.

## NaN semantics

`numpy.median` propagates NaN: if any value in the reduced slice is NaN, the
result is NaN. `nanmedian` does the opposite -- it ignores NaNs. Since
`subtract_overscan`'s `median` fallback needs to match `numpy.median`
(not `nanmedian`), the new `median` function wraps `nanmedian` and then
restores NaN-propagation with a final `xp.where(xp.any(xp.isnan(x), ...))`.
On NaN-free input `median` is bit-for-bit identical to `nanmedian`, since it
reuses the exact same sort-and-average algorithm.

## Testing

- `ccdproc/tests/test_nanfuncs.py`: added `median` (against `np.median`) to
  the differential `_FUNCS` table, which already exercises NaN-scattered,
  all-NaN, single-non-NaN, integer and boolean inputs across a range of
  shapes and axes -- this is what pins the NaN-propagating semantics. Also
  added `median` to `test_no_warning_on_all_nan_slice` and `test_bad_axis`
  (verified `numpy.median`, unlike `numpy.nanmedian`, does not warn on
  all-NaN input).
- `array-api-strict`: all "module 'array_api_strict' has no attribute
  'median'" errors are eliminated -- confirmed zero occurrences anywhere in
  the suite after this change.
- numpy backend: `ccdproc` suite is unaffected (0 failures, native `xp.median`
  path taken, byte-identical to before).
- dask backend: `test_nanfuncs.py` + `test_ccdproc.py` pass.

**Note on the strict-suite failure count:** the 4 tests originally cited as
failing due to this bug (`test_ccd_process[*]`,
`test_ccd_process_gain_corrected`) do *not* go green with this fix alone.
Diffing the full `array-api-strict` failure list before/after shows the same
85 tests fail both times -- but for 3 of these 4 tests the *reason* changes,
from `AttributeError: ... has no attribute 'median'` to two separate,
pre-existing bugs that this fix now uncovers (they were previously masked
because `subtract_overscan` raised first):
- `ccd_process`'s bad-pixel-mask handling uses `xp.asarray(bad_pixel_mask,
  dtype=bool)` with the Python builtin `bool` instead of the namespace's
  `xp.bool` (`core.py`, in `ccd_process`), which `array-api-strict` rejects.
- `gain_correct` builds its gain array with `xp.asarray(gain_value)` without
  specifying `device=`, so it ends up on the default device while the image
  data is on `array-api-strict`'s non-default device, and the subsequent
  multiply fails with "Arrays from two different devices".

Both are real, independent array-API incompatibilities, out of scope for
this PR, and worth filing as separate follow-up issues against #971.

### DIFF
--- a/ccdproc/_nanfuncs.py
+++ b/ccdproc/_nanfuncs.py
@@ -1,26 +1,29 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """
-NaN-aware sum/mean/standard deviation/median written only in terms of the
-array API.
+NaN-aware sum/mean/standard deviation/median, and a NaN-propagating median,
+written only in terms of the array API.
 
 ``nansum``/``nanmean``/``nanstd``/``nanmedian`` are not part of the array
 API standard, so this module provides fallbacks that work on any conforming
 namespace (``array-api-strict``, ``jax``, ``dask``, ``numpy``, ...). They
 are used by `ccdproc.combiner.Combiner` when the selected namespace does
-not provide the native versions.
+not provide the native versions. ``median`` is not part of the standard
+either, and this module also provides a fallback for it, built on
+``nanmedian``, used by `ccdproc.core.subtract_overscan` when the selected
+namespace does not provide the native version.
 
-All four functions promote integer and boolean input to the namespace's
+All five functions promote integer and boolean input to the namespace's
 default real floating dtype, which is where they part company with
 ``numpy.nansum``: numpy preserves an integer dtype, these do not. Every
 caller in `ccdproc` combines floating point image data, and the promotion
-keeps the four functions consistent with each other.
+keeps the five functions consistent with each other.
 """
 
 import operator
 
 import array_api_compat
 
-__all__ = ["nanmean", "nanmedian", "nanstd", "nansum"]
+__all__ = ["median", "nanmean", "nanmedian", "nanstd", "nansum"]
 
 
 def _setup(x, axis, xp):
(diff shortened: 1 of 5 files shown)


ccdproc/_nanfuncs.py
```diff
@@ -347,3 +350,39 @@ def nanmedian(x, /, *, axis=0, xp=None):
     # makes an all-NaN slice yield NaN. Do not remove it as redundant.
     nan = xp.asarray(xp.nan, dtype=s.dtype, device=device)
     return xp.where(xp.squeeze(n, axis=axis) == 0, nan, result)
+
+
+def median(x, /, *, axis=0, xp=None):
+    """
+    Median along an axis, using only array-API functions.
+
+    On input with no NaNs this is exactly `nanmedian` -- the same sorting
```
mwcraig: This fires in notes

mwcraig: Agreed — reading this as "this fits in Notes": the paragraph is implementation detail, and the other three functions in this module (`nansum`, `nanmean`, `nanmedian`) already keep that material in a `Notes` section, so `median` is the odd one out. Proposed shape:

```python
    """
    Median along an axis, using only array-API functions.

    Parameters
    ----------
    ...

    Returns
    -------
    array
        Median of ``x`` along ``axis``, with that axis removed. Slices that
        contain any NaN yield NaN, matching `numpy.median`; this is the
        difference from `nanmedian`, which ignores NaNs entirely.

    Notes
    -----
    On input with no NaNs this is exactly `nanmedian` -- the same sorting
    and index-picking algorithm is used, so the values agree bit for bit.
    The two differ only in how a NaN in the reduced slice is handled: this
    function propagates it to the result, matching `numpy.median`, while
    `nanmedian` ignores it. That NaN-propagating behaviour is restored here
    with a final `where` over whether any NaN is present along ``axis``,
    since `nanmedian` alone would silently drop NaNs instead.
    """
```

Happy to push that if it matches what you meant.

mwcraig: Done in 5399fc2: the paragraph now lives in a `Notes` section after `Returns`, matching the other functions in the module; no other text changed.

Issue #970, opened 2026-08-01
Test suite: raw np.zeros_like on backend-aware arrays in test_ccdproc_logging.py and test_rebin.py


mwcraig:
Running the array-api-strict backend surfaces a test-suite defect: raw `np.zeros_like` called on backend-aware fixture arrays.

```
CCDPROC_ARRAY_LIBRARY=array-api-strict pytest \
    ccdproc/tests/test_ccdproc_logging.py ccdproc/tests/test_image_collection.py -q
# 1 failed, 85 passed

ccdproc/tests/test_ccdproc_logging.py:75:
    bias = CCDData(np.zeros_like(ccd_data.data), unit="adu")
RuntimeError: Can't convert array on the 'array_api_strict.Device('device1')'
device to a NumPy array.
```

**Root cause**: `test_implicit_logging` builds its `CCDData` via the shared `ccd_data_func()` fixture (`ccdproc/tests/pytest_fixtures.py:66`), which already constructs data through the configured backend (`xp.asarray(..., device=xp_device)`). The test then calls plain `np.zeros_like(ccd_data.data)` instead of the backend-aware `xp.zeros_like` pattern used consistently elsewhere (e.g. `test_combiner.py:678`, `test_ccdproc.py:378`). A second, identical instance exists at `ccdproc/tests/test_rebin.py:70`.

**This is test-only, and narrower than the #941 table row.** That row ("test_image_collection.py, test_ccdproc_logging.py:75, misc singletons", ~4 failures, "an ImageFileCollection gap") was verified at branch tip d329ac5: only the `test_ccdproc_logging.py:75` failure reproduces. `test_image_collection.py` passes 77/77 under array-api-strict — `ImageFileCollection` defaults to numpy and only converts via `xp.asarray` when a caller explicitly passes `array_package`, so no library-side gap was found. The ImageFileCollection claim was either fixed since the 94ca9ec run or a false positive; no library issue is warranted.

**Suggested fix**: swap `np.zeros_like` for `xp.zeros_like` (using the per-test `xp` fixture as other test modules do) at `ccdproc/tests/test_ccdproc_logging.py:75` and `ccdproc/tests/test_rebin.py:70`.

Part of the array-API triage in #941; found with the array-api-strict tooling from #937/#939.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #969, opened 2026-08-01
Test suite: test_subtract_overscan calls numpy-only .copy() on array-API arrays


mwcraig:
Under `CCDPROC_ARRAY_LIBRARY=array-api-strict`, all 6 parametrized variants of `test_subtract_overscan` fail with the same error:

```
ccdproc/tests/test_ccdproc.py:174: in test_subtract_overscan
    science_data = ccd_data.data[science_region].copy()
E   AttributeError: 'Array' object has no attribute 'copy'
```

**Offending code** — `ccdproc/tests/test_ccdproc.py:174-175` (branch `array-api-strict-triage-tooling`):

```python
science_data = ccd_data.data[science_region].copy()
overscan_data = 0 * ccd_data.data[oscan_region].copy() + oscan
```

**Root cause**: the Array API standard does not include a `.copy()` method on arrays; the portable spelling is `xp.asarray(..., copy=True)`.

**This is test-only.** The library code is not implicated: `core.py`'s `subtract_overscan` only calls `.copy()` on a `CCDData` object, which is unaffected. The fix belongs in the test suite:

```python
science_data = xp.asarray(ccd_data.data[science_region], copy=True)
overscan_data = 0 * xp.asarray(ccd_data.data[oscan_region], copy=True) + oscan
```

(the `xp` namespace is already available in the test function).

Part of the array-API triage in #941; found with the array-api-strict tooling from #937/#939. Verified at branch tip d329ac5 (2026-08-01). Related test-suite cleanup of the same kind: raw `np.zeros_like` calls, tracked separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #990, opened 2026-08-24
Fix numpy-isms in strict-backend test bodies (test hygiene)

mwcraig:
## Summary

Test-only hygiene batch for the array-API migration (#971). Many
`array-api-strict` test failures were caused by numpy-only idioms in the
**test bodies**, not by bugs in ccdproc source. This PR fixes only files
under `ccdproc/tests/`.

**Strict suite: 85 failed, 366 passed → 65 failed, 386 passed** (36
skipped / 46 xfailed / 5 xpassed unchanged). Numpy backend stays at 0
failed (503 passed); dask backend on the four changed test files also
stays at 0 failed.

No CHANGES.rst entry — this is a test-only change.

## Fixed causes (itemized)

1. **#969** — `.copy()` is not an array-API method. Replaced
   `arr.copy()` with `xp.asarray(arr, copy=True)` in
   `test_ccdproc.py::test_subtract_overscan`. Closes all 6 parametrized
   failures for that test as numpy-isms; two of the six
   (`[True-False-False]`, `[True-False-True]`, the `median=True` cases)
   still fail afterward because they hit the separate, already-tracked
   `xp.median` gap at `core.py:629` (another agent is fixing that on a
   different branch) — the `.copy()` fix itself is still correct and is
   kept per the issue's scope.
2. **#970** — raw `np.zeros_like` mixed with backend arrays.
   `test_ccdproc_logging.py::test_implicit_logging` now builds its bias
   frame with `xp.zeros_like` (now passes). `test_rebin.py::test_rebin_ccddata[True-True]`
   was updated the same way, but it still fails afterward: astropy's
   `CCDData.mask` setter (`astropy/nddata/compat.py`) forces
   `np.asarray(value, dtype=np.bool_)`, which can't convert an array on
   a non-default strict device. That's an astropy-internals limitation
   outside a test-only fix, so it's kept per the issue's scope and noted
   as a residual failure.
3. `test__overscan_schange` used `xp.allclose`, which isn't in the
   array-API standard. Replaced with `xp.all(xpx.isclose(...))`.
4. `test_combiner.py` numpy-only method calls (`.mean()`, `.all()`) on
   backend arrays, replaced with `xp.mean(...)` / `xp.all(...)`:
   `test_combiner_minmax`, `test_combiner_minmax_max`,
   `test_combiner_minmax_min`.
5. Broader numpy-isms diagnosed from actual strict tracebacks:
   - Single-axis indexing (`arr[i]`) on arrays with `ndim > 1`, which
     the array-API standard requires an explicit ellipsis for
     (`arr[i, ...]`): `test_subtract_overscan_fails`,
     `test_user_supplied_combine_func_that_relies_on_masks[sum_combine]`.
   - int/float mixed arithmetic that strict's dtype promotion rejects:
     `test_trim_with_wcs_alters_wcs` (`xp.asarray(shape) / 2` on an int
     array), `test_pixelwise_weights`, `test_combiner_sum_weighted`,
     `test_combiner_sum_weighted_by_pixel`,
     `test_combiner_sum_weighted_with_mask`,
     `test_combiner_uncertainty_average`,
     `test_combiner_uncertainty_average_mask`,
     `test_combiner_uncertainty_median_mask`,
     `test_combiner_uncertainty_sum_mask` (int-literal arrays passed to
     `xp.std`/`xp.sqrt`, and an unwrapped `np.float64` from
     `astropy.stats.median_absolute_deviation` multiplied against a
     strict `Array`, following the existing `float(...)` cast pattern
     already used elsewhere in that file).
   - `xp.asarray(list_of_arrays)` ("Nested Arrays are not allowed"),
     replaced with `xp.stack(...)`:
     `test_user_supplied_combine_func_that_relies_on_masks[sum_combine]`.

## Failures examined and deliberately left (with why)

- `test_ccd_process[*]`, `test_ccd_process_gain_corrected`,
  `test_subtract_overscan[True-False-*]` (median path) — the known,
  already-tracked `xp.median` gap at `core.py:629`; another agent is
  fixing this on a separate branch, explicitly out of scope here.
- `test_flat_correct*`, `test_gain_correct*`, `test_gain.py::*` — source
  bugs in `gain_correct`/`flat_correct` (`core.py:934`, `core.py:1023`,
  `core.py:1037`): `xp.asarray(...)` called without `device=`, causing
  "Arrays from two different devices" errors from ccdproc source, not
  the test body.
- `test_flat_correct_data_uncertainty` — intentionally uses a raw numpy
  array as uncertainty; the test's own comment documents this as a
  regression test for astropy's `NDUncertainty` explicitly checking for
  `numpy.ndarray`, to be removed once that's fixed upstream.
- `test_flat_correct_deviation`, `test_sigma_func_for_ccddata`,
  `test_combiner_median`/`average`/`sum`/`dtype`,
  `test_average_combine_uncertainty`, `test_median_combine_uncertainty`,
  `test_sum_combine_uncertainty`, `test_combine_result_uncertainty_and_mask[*]` —
  fail inside `astropy.stats.median_absolute_deviation` or
  `CCDData.mask`/`Quantity` construction forcing numpy conversion of a
  non-default-device strict array; astropy-internals, not test-body.
- `test_combiner_sigmaclip_high/low/single_pix` — the test intentionally
  passes `astropy.stats.median_absolute_deviation` as `dev_func`, which
  forces a numpy conversion inside `combiner.py`'s `sigma_clipping`;
  same astropy-internals limitation.
- `test_combine_average_fitsimages`, `test_combine_numpyndarray`,
  `test_combiner_image_file_collection_input`,
  `test_combine_image_file_collection_input`,
  `test_combine_average_ccddata`, `test_combine_limitedmem_fitsimages`,
  `test_combine_limitedmem_scale_fitsimages`,
  `test_combine_ccd_with_uncertainty_and_mask_from_fits[*]` — explicitly
  out of scope per the task brief (`combine()` file-input handling
  returns plain numpy arrays).
- `test_clip_extrema_3d/alone/via_combine/with_other_rejection` —
  explicitly out of scope per the task brief (clip_extrema fancy
  indexing).
- `test_combiner_with_scaling`, `test_combiner_result_dtype`,
  `test_combine_overwrite_output`, `test_combiner_with_scaling_uncertainty[*]` —
  astropy `ndarithmetic`'s `np.result_type(ref, operand)` can't convert
  a strict `Array`'s dtype; astropy-internals ("Could not convert Array"
  pattern named as out of scope in the task brief).
- `test_writeable_after_combine[*]` — `CCDData.write`/mask conversion
  forcing numpy on a non-default device; source/astropy-internals.
- `test_3d_combiner_with_scaling`, `test_combiner_with_scaling` (helper
  `scale_by_mean`) — tried fixing the `.mean()` numpy-ism, but it then
  exposed a genuine ccdproc source bug: `Combiner.scaling`'s setter
  (`combiner.py:327`) calls `xp.asarray(list_of_arrays)` on a list of
  strict `Array`s ("Nested Arrays are not allowed. Use `stack` instead."),
  and for `test_combiner_with_scaling` the test is blocked even earlier
  by the astropy `ndarithmetic` issue above. Per the task's revert
  policy for non-numbered items, these test-body changes were reverted
  back to the pre-existing (still-numpy) form and left failing.
- `test_transform_image[True-True]` — the test's user-supplied `tran =
  lambda arr: 10 * arr` is applied generically to data, uncertainty, and
  mask by `transform_image`; ccdproc's array-API wrapper coerces the
  mask to `bool` before this call, so `10 * bool_array` hits strict's
  "Only numeric dtypes are allowed in `__rmul__`". Making the test's
  transform function dtype-aware would change what the test exercises
  rather than being a simple hygiene fix, so left as-is.
- `test_unit_mismatch_behaves_as_expected` — fails inside astropy's
  `Quantity.__new__` forcing a numpy conversion via
  `_arithmetic_data`/`_prepare_then_do_arithmetic`; astropy-internals.
- `test_image_collection.py::test_generator_ccds_without_unit` — fails
  only when run as part of the full suite (passes in isolation),
  indicating cross-test state leakage unrelated to array-API numpy-isms;
  left as a pre-existing, order-dependent issue outside this batch's
  scope.

## Test plan

- [x] `CCDPROC_ARRAY_LIBRARY=array-api-strict tox -e strict` (from the
      main checkout's tox env): 85 failed, 366 passed → 65 failed, 386
      passed.
- [x] `CCDPROC_ARRAY_LIBRARY=numpy` full suite: 503 passed, 0 failed (no
      regression).
- [x] `CCDPROC_ARRAY_LIBRARY=dask` on the four changed test files: 181
      passed, 0 failed.
- [x] `ruff check` / `ruff format --check` clean on changed files.

Part of #971

### DIFF
--- a/ccdproc/tests/test_combiner.py
+++ b/ccdproc/tests/test_combiner.py
@@ -358,9 +358,7 @@ def test_pixelwise_weights():
     ]
     combo = Combiner(ccd_list)
     combo.weights = xp.ones_like(combo._data_arr)
-    combo.weights = xpx.at(combo.weights)[:, 5, 5].set(
-        xp.asarray([1, 5, 10], dtype=xp.float64)
-    )
+    combo.weights = xpx.at(combo.weights)[:, 5, 5].set(xp.asarray([1.0, 5.0, 10.0]))
     ccd = combo.average_combine()
     assert xp.all(xpx.isclose(ccd.data[5, 5], 312.5))
     assert xp.all(xpx.isclose(ccd.data[0, 0], 0))
@@ -583,10 +581,10 @@ def test_combiner_sum():
 
 # test weighted sum
 def test_combiner_sum_weighted():
-    ccd_data = CCDData(data=xp.asarray([[0, 1], [2, 3]], dtype=xp.float64), unit="adu")
+    ccd_data = CCDData(data=xp.asarray([[0.0, 1.0], [2.0, 3.0]]), unit="adu")
     ccd_list = [ccd_data, ccd_data, ccd_data]
     c = Combiner(ccd_list)
-    c.weights = xp.asarray([1, 2, 3], dtype=xp.float64)
+    c.weights = xp.asarray([1.0, 2.0, 3.0])
     ccd = c.sum_combine()
     expected_result = sum(w * d.data for w, d in zip(c.weights, ccd_list, strict=True))
     assert xp.all(xpx.isclose(ccd.data, expected_result))
@@ -598,10 +596,10 @@ def test_combiner_sum_weighted_by_pixel():
     ccd_list = [ccd_data, ccd_data, ccd_data]
     c = Combiner(ccd_list)
     # Weights below are chosen so that every entry in
-    weights_pixel = [[8, 4], [2, 1]]
-    c.weights = xp.asarray([weights_pixel] * 3, dtype=xp.float64)
+    weights_pixel = [[8.0, 4.0], [2.0, 1.0]]
+    c.weights = xp.asarray([weights_pixel] * 3)
     ccd = c.sum_combine()
-    expected_result = xp.asarray([[24, 24], [24, 24]], dtype=xp.float64)
+    expected_result = xp.asarray([[24.0, 24.0], [24.0, 24.0]])
     assert xp.all(xpx.isclose(ccd.data, expected_result))
 
 
@@ -612,11 +610,11 @@ def test_combiner_sum_weighted_with_mask():
         CCDData(xp.asarray([[10, 20]]), unit=u.adu),
     ]
     combiner = Combiner(ccd_list)
-    combiner.weights = xp.asarray([1, 3], dtype=xp.float64)
+    combiner.weights = xp.asarray([1.0, 3.0])
 
     combined = combiner.sum_combine()
 
-    expected = xp.asarray([[30, 62]], dtype=xp.float64)
+    expected = xp.asarray([[30.0, 62.0]])
     assert xp.all(xpx.isclose(combined.data, expected))
 
 
@@ -1126,13 +1124,11 @@ def test_combiner_uncertainty_average_mask():
     c = Combiner(ccd_list)
     ccd = c.average_combine()
     # Just the standard deviation of ccd data.
-    ref_uncertainty = xp.ones((10, 10)) * xp.std(
-        xp.asarray([1, 2, 3], dtype=xp.float64)
-    )
+    ref_uncertainty = xp.ones((10, 10)) * xp.std(xp.asarray([1.0, 2.0, 3.0]))
     # Correction because we combined two images.
     ref_uncertainty /= xp.sqrt(xp.asarray(3.0))
     ref_uncertainty = xpx.at(ref_uncertainty)[5, 5].set(
-        xp.std(xp.asarray([2, 3], dtype=xp.float64)) / xp.sqrt(xp.asarray(2.0))
+        xp.std(xp.asarray([2.0, 3.0])) / xp.sqrt(xp.asarray(2.0))
     )
     assert xp.all(xpx.isclose(ccd.uncertainty.array, ref_uncertainty))
 
@@ -1176,12 +1172,10 @@ def test_combiner_uncertainty_sum_mask():
     c = Combiner(ccd_list)
     ccd = c.sum_combine()
     # Just the standard deviation of ccd data.
-    ref_uncertainty = xp.ones((10, 10)) * xp.std(
-        xp.asarray([1, 2, 3], dtype=xp.float64)
-    )
+    ref_uncertainty = xp.ones((10, 10)) * xp.std(xp.asarray([1.0, 2.0, 3.0]))
     ref_uncertainty *= xp.sqrt(xp.asarray(3.0))
     ref_uncertainty = xpx.at(ref_uncertainty)[5, 5].set(
-        xp.std(xp.asarray([2, 3], dtype=xp.float64)) * xp.sqrt(xp.asarray(2.0))
+        xp.std(xp.asarray([2.0, 3.0])) * xp.sqrt(xp.asarray(2.0))
     )
     assert xp.all(xpx.isclose(ccd.uncertainty.array, ref_uncertainty))
 
--- a/ccdproc/tests/test_combiner.py
+++ b/ccdproc/tests/test_combiner.py
@@ -358,7 +358,9 @@ def test_pixelwise_weights():
     ]
     combo = Combiner(ccd_list)
     combo.weights = xp.ones_like(combo._data_arr)
-    combo.weights = xpx.at(combo.weights)[:, 5, 5].set(xp.asarray([1, 5, 10]))
+    combo.weights = xpx.at(combo.weights)[:, 5, 5].set(
+        xp.asarray([1, 5, 10], dtype=xp.float64)
+    )
     ccd = combo.average_combine()
     assert xp.all(xpx.isclose(ccd.data[5, 5], 312.5))
     assert xp.all(xpx.isclose(ccd.data[0, 0], 0))
@@ -463,7 +465,7 @@ def test_combiner_minmax():
     c = Combiner(ccd_list)
     c.minmax_clipping(min_clip=-500, max_clip=500)
     ccd = c.median_combine()
-    assert ccd.data.mean() == 0
+    assert xp.mean(ccd.data) == 0
 
 
 def test_combiner_minmax_max():
@@ -475,7 +477,7 @@ def test_combiner_minmax_max():
 
     c = Combiner(ccd_list)
     c.minmax_clipping(min_clip=None, max_clip=500)
-    assert c._data_arr_mask[2].all()
+    assert xp.all(c._data_arr_mask[2, ...])
 
 
 def test_combiner_minmax_min():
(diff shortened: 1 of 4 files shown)


ccdproc/tests/test_combiner.py
```diff
@@ -475,7 +477,7 @@ def test_combiner_minmax_max():
 
     c = Combiner(ccd_list)
     c.minmax_clipping(min_clip=None, max_clip=500)
-    assert c._data_arr_mask[2].all()
+    assert xp.all(c._data_arr_mask[2, ...])
```
mwcraig, line 478: Why the ellipsis index?

ccdproc/tests/test_combiner.py
```diff
@@ -1147,14 +1151,14 @@ def test_combiner_uncertainty_median_mask():
...
-    # Correction because we combined two images.
-    ref_uncertainty /= xp.sqrt(3)  # 0.855980789955
     # It turns out that the expression below evaluates to a np.float64, which
     # introduces numpy into the array namespace, which raises an error
     # when arrat_api_compat tries to figure out the namespace. Casting
     # it to a regular float fixes that.
-    med_value = float(mad_to_sigma * mad([2, 3]) / xp.sqrt(2))
+    ref_uncertainty = xp.ones((10, 10)) * float(mad_to_sigma * mad([1, 2, 3]))
+    # Correction because we combined two images.
+    ref_uncertainty /= xp.sqrt(xp.asarray(3.0))  # 0.855980789955
```
mwcraig, line 1156: Why this change to the test? It does not seem API related

ccdproc/tests/test_combiner.py
```diff
@@ -1172,10 +1176,12 @@ def test_combiner_uncertainty_sum_mask():
...
     # Just the standard deviation of ccd data.
-    ref_uncertainty = xp.ones((10, 10)) * xp.std(xp.asarray([1, 2, 3]))
-    ref_uncertainty *= xp.sqrt(3)
+    ref_uncertainty = xp.ones((10, 10)) * xp.std(
+        xp.asarray([1, 2, 3], dtype=xp.float64)
+    )
+    ref_uncertainty *= xp.sqrt(xp.asarray(3.0))
     ref_uncertainty = xpx.at(ref_uncertainty)[5, 5].set(
-        xp.std(xp.asarray([2, 3])) * xp.sqrt(2)
+        xp.std(xp.asarray([2, 3], dtype=xp.float64)) * xp.sqrt(xp.asarray(2.0))
```
mwcraig: Why do we need the dtypes all over the place?

ccdproc/tests/test_combiner.py
mwcraig: The array API standard leaves indexing an `ndim > 1` array with fewer indices than dimensions unspecified, and array-api-strict enforces that. `_data_arr_mask` is 3-D, so with the old `c._data_arr_mask[2]` the strict lane fails with:

```
IndexError: self.ndim=3, but the multi-axes index only specifies 1 dimensions. If this
was intentional, add a trailing ellipsis (...) which expands into as many slices (:) as
necessary - this is what np.ndarray arrays implicitly do, but such flat indexing
behaviour is not specified in the Array API.
```

Verified by reverting just the ellipsis while keeping `xp.all(...)`: same `IndexError` on strict, passes on numpy — so the ellipsis is load-bearing on its own, independent of the `.all()` → `xp.all` fix. The same justification covers the other six ellipsis additions in this PR (line 492, four sites in `test_user_supplied_combine_func_that_relies_on_masks`, and `test_ccdproc.py:356` — partial *slices* trigger it too, e.g. `a[0:10]` on a 2-D strict array raises the same error).

ccdproc/tests/test_combiner.py
mwcraig: It is API-related — the diff just looks like a logic change because the pre-existing "np.float64" comment block moved up (it now applies to the `ref_uncertainty` line too). The math is unchanged; reverting this hunk passes on numpy but fails on strict. Two violations in the old code:

1. `xp.ones((10, 10)) * mad_to_sigma * mad([1, 2, 3])` — astropy's `mad()` returns `np.float64`, and strict rejects foreign scalars as operands: `TypeError: Expected Array or Python scalar; got <class 'numpy.float64'>` (this is the exact failure on revert). Hence the `float(...)` wraps.
2. `xp.sqrt(3)` — strict's elementwise functions take arrays only: `TypeError: Only Array objects are allowed; got <class 'int'>`. Hence `xp.asarray(...)`.

That said, `xp.sqrt(xp.asarray(3.0))` is clunkier than it needs to be: strict *does* allow Python scalars in arithmetic, so `ref_uncertainty /= math.sqrt(3)` (and `/ math.sqrt(2)` for `med_value`) passes both lanes — verified. That would simplify all 7 `xp.sqrt(xp.asarray(...))` sites in the PR; happy to push that change if you prefer it.

ccdproc/tests/test_combiner.py
mwcraig: Each `dtype=` is individually load-bearing on the strict lane (verified by reverting them one at a time: every revert fails strict, all pass numpy). Two root causes across the PR's 12 sites:

1. **`xp.std` requires a floating dtype** (4 sites, including this one): `xp.asarray([1, 2, 3])` is int64, and strict raises `TypeError: Only real floating-point dtypes are allowed in std`.
2. **int64/float64 mixing is forbidden** (7 sites, the weights/expected arrays in the `sum_weighted*` and `pixelwise_weights` tests): the combine functions return float64 even from int input, while the tests' int reference arrays stay int64, so the tests' own `xpx.isclose`/`xpx.at(...).set(...)` calls fail with `TypeError: array_api_strict.float64 and array_api_strict.int64 cannot be type promoted together`. numpy promotes silently; the standard leaves int↔float promotion undefined and strict forbids it. (Plus one `__truediv__`-on-int site at `test_ccdproc.py:405`.)

Note the failures are all in test-body reference construction, not in ccdproc source — the library output is float64 either way.

If the `dtype=xp.float64` noise bothers you, float literals do the same job more quietly: `xp.std(xp.asarray([1.0, 2.0, 3.0]))` — verified passing both lanes. That removes every `dtype=` except the `test_ccdproc.py:405` one (its input is `ccd_data.shape`, so there's no literal to float-ify). Happy to push that variant if you prefer.

mwcraig: Let's go with the float literals.

mwcraig: Done in 288d55a: all 11 literal-array sites in `test_combiner.py` now use float literals (e.g. `xp.asarray([1.0, 2.0, 3.0])`), with no `dtype=` argument. The only `dtype=xp.float64` left in the PR is the `crpix` line in `test_ccdproc.py`, since its input is `ccd_data.shape`. Verified `test_combiner.py` on numpy, dask, and jax (x64) all green, and the strict lane has the identical pre-existing failure set before and after the change.

PR #992, opened 2026-08-25
Keep combine result masks in the array namespace

mwcraig:
Part of #971 (section 1, "combine methods pass `mask=` to `CCDData(...)`" — the largest remaining bucket on the strict job).

`Combiner.average_combine`, `median_combine` and `sum_combine` built their result with `CCDData(..., mask=mask)`, and `combine()` assigned `ccd.mask = ...` at several points, so astropy's `CCDData.mask` setter (`np.asarray(value, dtype=np.bool_)`) ran on every result. On backends whose arrays cannot be converted to NumPy (a non-default `array-api-strict` device, a GPU) that raises; on jax/dask it silently handed back a NumPy mask on non-NumPy data.

### Changes

- The three `*_combine` methods assign `combined_image._mask = mask` after construction, the same workaround `ccd_process` already uses (`core.py:381`), with the same TODO.
- `combine()` does the same at its own mask sites, and coerces the template image's mask into the data's namespace and onto its device the way `Combiner.__init__` does — a `CCDData` that received its mask through the setter always carries a NumPy mask, so without this the tile loop fails with "Multiple namespaces for array inputs" once the tiles' masks are no longer NumPy.
- The tile loop's `ccd.mask.copy()` (NumPy-only; it only worked because the mask was always NumPy) is replaced by `xpx.at(...).set(..., copy=True)`.
- Test bodies that were previously never reached on strict are made array-API clean: positional `axis`, `xp.sqrt(3)` → `math.sqrt(3)`, `mask.sum()` → `xp.count_nonzero`.

### Results

- Strict job locally (`tox -e strict`): **63 → 54 failed**, 418 → 427 passed, no new failures.
- numpy full suite, and dask / jax (`JAX_ENABLE_X64=1`) `test_combiner.py`: all green.
- The rest of the original 12-test bucket now stops at the next layer: `median_combine` uses the default `sigma_func` → `astropy.stats.median_absolute_deviation` (#929, section 3), and `test_writeable_after_combine` writes a `device1` array to FITS, which belongs with the section-2 "write a NumPy copy" tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -91,6 +91,13 @@ Bug Fixes
   standard, built on the existing NaN-aware ``nanmedian``, and use it in
   ``subtract_overscan`` when the selected array namespace has no ``median``,
   instead of raising ``AttributeError``. [#989]
+- Keep the mask of the result of ``Combiner.average_combine``,
+  ``median_combine`` and ``sum_combine`` and of ``combine`` in the array
+  namespace and on the device of the data, instead of letting the
+  ``CCDData.mask`` setter convert it to NumPy, which fails for arrays that
+  cannot be converted (for example on a non-default device). ``combine``
+  now also coerces a NumPy mask on the template image into the data's
+  namespace and device. [#992]
 
 2.5.1 (2025-07-05)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -92,12 +92,8 @@ Bug Fixes
   ``subtract_overscan`` when the selected array namespace has no ``median``,
   instead of raising ``AttributeError``. [#989]
 - Keep the mask of the result of ``Combiner.average_combine``,
-  ``median_combine`` and ``sum_combine`` and of ``combine`` in the array
-  namespace and on the device of the data, instead of letting the
-  ``CCDData.mask`` setter convert it to NumPy, which fails for arrays that
-  cannot be converted (for example on a non-default device). ``combine``
-  now also coerces a NumPy mask on the template image into the data's
-  namespace and device. [#992]
+  ``median_combine``, ``sum_combine`` and ``combine`` in the array namespace
+  and on the device of the data instead of converting it to NumPy. [#992]
 
 2.5.1 (2025-07-05)
 ------------------
--- a/ccdproc/tests/test_combiner.py
+++ b/ccdproc/tests/test_combiner.py
@@ -1,4 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+import math
 import types
 from functools import partial
 
@@ -933,7 +934,7 @@ def test_average_combine_uncertainty():
     ccd_list = [ccd_data, ccd_data, ccd_data]
     c = Combiner(ccd_list)
     ccd = c.average_combine(uncertainty_func=xp.sum)
-    uncert_ref = xp.sum(c._data_arr, 0) / xp.sqrt(3)
+    uncert_ref = xp.sum(c._data_arr, axis=0) / math.sqrt(3)
     assert xp.all(xpx.isclose(ccd.uncertainty.array, uncert_ref))
 
     # Compare this also to the "combine" call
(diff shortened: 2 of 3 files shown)


ccdproc/tests/test_combiner.py
```diff
@@ -933,7 +934,7 @@ def test_average_combine_uncertainty():
     ccd_list = [ccd_data, ccd_data, ccd_data]
     c = Combiner(ccd_list)
     ccd = c.average_combine(uncertainty_func=xp.sum)
-    uncert_ref = xp.sum(c._data_arr, 0) / xp.sqrt(3)
+    uncert_ref = xp.sum(c._data_arr, axis=0) / math.sqrt(3)
```
mwcraig, line 937: What is the problem with using `xp.sqrt`?

ccdproc/tests/test_combiner.py
```diff
@@ -1054,9 +1055,9 @@ def test_combine_result_uncertainty_and_mask(comb_func, mask_point):
     # Check that the right point is masked, and only one point is
     # masked
     assert expected_result.mask[0, 0] == mask_point
-    assert expected_result.mask.sum() == mask_point
+    assert int(xp.count_nonzero(expected_result.mask)) == int(mask_point)
```
mwcraig, line 1058: Why the `int` casts here?

CHANGES.rst
```diff
@@ -91,6 +91,13 @@ Bug Fixes
   standard, built on the existing NaN-aware ``nanmedian``, and use it in
   ``subtract_overscan`` when the selected array namespace has no ``median``,
   instead of raising ``AttributeError``. [#989]
+- Keep the mask of the result of ``Combiner.average_combine``,
```
mwcraig, line 94: Shorten this up.

ccdproc/tests/test_combiner.py
mwcraig: Under `array_api_strict` it raises. The standard's elementwise functions take arrays only, so `xp.sqrt(3)` (a Python `int`) fails with

```
AttributeError: 'int' object has no attribute 'dtype'
```

NumPy, JAX and Dask all happily accept the scalar, which is why this was never seen before. `math.sqrt(3)` gives a plain Python `float`, and `array / python_float` is defined by the standard for every namespace, so it works everywhere without having to build a 0-d array first.

The `0` → `axis=0` change on the same line is for the same reason: in the standard `sum`'s `axis` is keyword-only, and strict enforces it (`TypeError: sum() takes 1 positional argument but 2 were given`).

ccdproc/tests/test_combiner.py
mwcraig: The one on the right is required under `array_api_strict`; the one on the left is optional.

Before this PR the result mask was a NumPy array (the `CCDData.mask` setter converted it), so `.mask.sum() == mask_point` was plain NumPy-vs-`bool`. Now the mask stays in the namespace, and under strict:

- `mask.sum()` doesn't exist (no array methods in the standard), hence `xp.count_nonzero`, which returns a 0-d `int64` array.
- comparing that to the parametrized `mask_point` (a Python `bool`) raises `TypeError: Python bool scalars can only be promoted with bool arrays` — the standard doesn't define bool/int promotion. So `int(mask_point)` is needed on the right.

The `int(...)` on the left isn't needed for correctness (`assert` on a 0-d bool array works via `__bool__`); it just turns the check into a Python `int == int` so a failure reads `assert 0 == 1` rather than `assert Array(False, dtype=array_api_strict.bool)`. Happy to drop it if you'd rather keep the line shorter:

```python
assert xp.count_nonzero(expected_result.mask) == int(mask_point)
```

CHANGES.rst
mwcraig: Proposed:

```rst
- Keep the mask of the result of ``Combiner.average_combine``,
  ``median_combine``, ``sum_combine`` and ``combine`` in the array namespace
  and on the device of the data instead of converting it to NumPy. [#992]
```

That drops the explanation of *why* the conversion failed and the note about `combine` coercing a NumPy mask on the template, which is an implementation detail of the same fix. I'll apply it on the branch if that reads right to you.

mwcraig: Applied in 280f32f.

PR #993, opened 2026-08-25
Fix gain_correct and flat_correct on a non-default device

mwcraig:
Part of #971 (the `gain_correct`/`flat_correct` `device=` bucket, plus the multiply/divide uncertainty-propagation bucket that sits directly behind it).

On the strict job this takes the failure count from **54 to 41** on top of `main` (`ba758f0`), with no new failures. The two strict `xfail` markers in `test_ccddata_wrapper_for_array_api.py` that covered the propagation leak started XPASSing and are removed; the 5 remaining XPASSes are the same stale `test_cosmicray.py` markers as on `main`.

Three commits:

1. **`gain_correct`/`flat_correct` create the gain and the flat normalization on the device of the data.** `xp.asarray(gain_value)` / `xp.asarray(flat_mean)` had no `device=`, so the scalar landed on the default device and the multiply/divide failed with "arrays from two different devices" when the image is elsewhere (`core.py:968`, `:1057`). Also, a plain-number gain is now made a `float` first: `gain_correct(ccd, gain=3)` produced an `int64` array and the array API standard does not promote that with `float64` data.

2. **The wrapper's uncertainty mixin owns `_propagate_multiply_divide`.** Astropy's version uses `np.abs`/`np.sqrt`, which convert the operands to numpy — silently on a CPU array (the result then fails at `xp.sqrt`), and with a `RuntimeError` on a non-default device. The override is astropy's formula with the numpy calls replaced by the array namespace's, and `_propagate_multiply`/`_propagate_divide` now call it. This is what makes `gain_correct` on an image *with* an uncertainty work on a non-default device.

3. **Test bodies.** The flats in `test_flat_correct`, `test_flat_correct_norm_value` and `test_flat_correct_deviation` were built on the default device while the image is on the testing device; `test_flat_correct_data_uncertainty` used a numpy uncertainty and `.all()` (not in the standard); `test_gain.py` multiplied a numpy `float64` from `Quantity.value` into the array and was missing the `assert`s on its two `xp.all(...)` checks.

The one gain/flat test still failing on strict, `test_ccd_process_gain_corrected`, stops at the `ccd_process` builtin-`bool` bucket (`core.py:383`), which is separate.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -94,6 +94,12 @@ Bug Fixes
 - Keep the mask of the result of ``Combiner.average_combine``,
   ``median_combine``, ``sum_combine`` and ``combine`` in the array namespace
   and on the device of the data instead of converting it to NumPy. [#992]
+- Fix ``gain_correct`` and ``flat_correct`` for images on a non-default
+  device: the gain and the flat normalization are now created on the device
+  of the data, an integer gain is converted to a float so that it can be
+  multiplied with floating-point data in any array namespace, and the
+  uncertainty propagation for multiplication and division no longer goes
+  through NumPy. [#993]
 
 2.5.1 (2025-07-05)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -95,11 +95,9 @@ Bug Fixes
   ``median_combine``, ``sum_combine`` and ``combine`` in the array namespace
   and on the device of the data instead of converting it to NumPy. [#992]
 - Fix ``gain_correct`` and ``flat_correct`` for images on a non-default
-  device: the gain and the flat normalization are now created on the device
-  of the data, an integer gain is converted to a float so that it can be
-  multiplied with floating-point data in any array namespace, and the
-  uncertainty propagation for multiplication and division no longer goes
-  through NumPy. [#993]
+  device: put the gain and flat normalization on the device of the data, cast
+  an integer gain to float, and keep the uncertainty propagation for
+  multiplication and division out of NumPy. [#993]
 
 2.5.1 (2025-07-05)
 ------------------
--- a/ccdproc/_ccddata_wrapper_for_array_api.py
+++ b/ccdproc/_ccddata_wrapper_for_array_api.py
@@ -336,19 +336,43 @@ class _ArrayAPIPropagationMixin:
         from_variance=lambda x: x,
     ):
         """
-        Error propagation for multiplication or division.
+        Propagate uncertainty for multiplication or division.
 
-        This is astropy's ``_VariancePropagationMixin._propagate_multiply_divide``
-        with the ``numpy`` calls replaced by their array namespace equivalents;
-        astropy's ``np.abs`` converts the operands to numpy arrays, which fails
-        for arrays that cannot be converted (for example on a non-default
-        device). See the astropy version for the derivation of the formulae.
+        This is astropy's
+        ``_VariancePropagationMixin._propagate_multiply_divide`` with the
+        NumPy calls replaced by their array-namespace equivalents; see the
+        astropy version for the derivation of the formulae. Unlike astropy's
+        version this does not convert the uncertainties between units,
+        because ``_CCDDataWrapperForArrayAPI._arithmetic_wrapper`` removes
+        the units from the uncertainties before doing the arithmetic.
 
-        Unlike astropy's version this does not convert the uncertainties
-        between units: ``_CCDDataWrapperForArrayAPI._arithmetic_wrapper``
-        removes the units from the uncertainties before doing the arithmetic,
-        so there is never a unit to convert. ``result_data`` is not used by the
-        formulae and is only accepted for compatibility with astropy.
+        Parameters
+        ----------
+        other_uncert : `~astropy.nddata.NDUncertainty`
+            The uncertainty of the other operand. Its ``array`` and
+            ``parent_nddata.data`` must be in the same array namespace as
+            ``self.array``.
+        result_data : array-like
+            Accepted only for signature compatibility with astropy; the
+            formulae do not use it.
+        correlation : float or array-like
+            Correlation coefficient between the two operands, ``0`` for
+            uncorrelated.
+        divide : bool, optional
+            ``True`` for division, ``False`` (default) for multiplication.
+        to_variance : callable, optional
+            Converts the stored uncertainty array to a variance. Defaults to
+            the identity, i.e. the uncertainty is already a variance.
+        from_variance : callable, optional
+            Converts a variance back to the stored uncertainty type. Defaults
+            to the identity.
+
+        Returns
+        -------
+        array-like
+            The propagated uncertainty array, in the same array namespace and
+            on the same device as the inputs, in the representation of
+            ``self`` (as determined by ``from_variance``).
         """
         del result_data  # accepted only for compatibility with astropy
         xp = array_api_compat.array_namespace(self.array, other_uncert.array)
--- a/ccdproc/_ccddata_wrapper_for_array_api.py
+++ b/ccdproc/_ccddata_wrapper_for_array_api.py
@@ -305,7 +305,7 @@ class _ArrayAPIPropagationMixin:
     def _propagate_multiply(self, other_uncert, result_data, correlation):
         xp = array_api_compat.array_namespace(self.array, other_uncert.array)
         to_variance, from_variance = self._variance_hooks(xp)
-        return super()._propagate_multiply_divide(
+        return self._propagate_multiply_divide(
             other_uncert,
             result_data,
             correlation,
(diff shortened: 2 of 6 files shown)


mwcraig:
Codecov's 5 missing patch lines (from its line-level report for `ffa9337`, not the summary) were all inside the new `_propagate_multiply_divide`:

| line | what | resolution in 0772348 |
|---|---|---|
| 351 | `result_data = result_data.value` | removed — `result_data` isn't used by the formulae at all; the parameter stays for signature compatibility with astropy |
| 364, 383 | the two uncertainty-unit conversion branches | removed — unreachable through the wrapper, which nulls the uncertainty units in `_arithmetic_wrapper` before doing arithmetic (and the branch also needs a parent unit, which the wrapper has removed too) |
| 393 | `left = 0` (no uncertainty on `self`) | new test `test_wrapped_arithmetic_uncertainty_only_on_operand` — 3 uncertainty types × 4 ops, against astropy's numpy result. This is the `flat_correct(ccd_without_uncertainty, flat_with_uncertainty)` path |
| 396 | correlated uncertainties | new test `test_wrapped_arithmetic_correlated_uncertainty` — 3 uncertainty types × multiply/divide with `uncertainty_correlation=0.5`, against astropy's numpy result |

Local coverage on numpy now reports no misses in that function. Full matrix on 0772348: numpy 546 passed, dask 539, jax 538 (7 xfailed); strict still 41 failed with the same set as before, 5 XPASS (the stale `test_cosmicray.py` markers).

One thing found while writing the correlation test, **not** fixed here: correlated `add`/`subtract` still leaks, because the wrapper delegates those to astropy's `_propagate_add_sub`, whose correlation term is `np.sqrt(this * other)`. Uncorrelated add/subtract is fine. Added as a bullet to #971.

ccdproc/_ccddata_wrapper_for_array_api.py
```diff
@@ -326,6 +326,69 @@ def _propagate_divide(self, other_uncert, result_data, correlation):
...
+        self,
+        other_uncert,
+        result_data,
+        correlation,
+        divide=False,
+        to_variance=lambda x: x,
+        from_variance=lambda x: x,
+    ):
+        """
+        Error propagation for multiplication or division.
```
mwcraig: Needs a proper numpy-style docstring

CHANGES.rst
```diff
@@ -94,6 +94,12 @@ Bug Fixes
 - Keep the mask of the result of ``Combiner.average_combine``,
   ``median_combine``, ``sum_combine`` and ``combine`` in the array namespace
   and on the device of the data instead of converting it to NumPy. [#992]
+- Fix ``gain_correct`` and ``flat_correct`` for images on a non-default
```
mwcraig, line 97: Shorten this.

ccdproc/_ccddata_wrapper_for_array_api.py
```diff
@@ -326,6 +326,69 @@ def _propagate_divide(self, other_uncert, result_data, correlation):
             from_variance=from_variance,
         )
 
+    def _propagate_multiply_divide(
```
mwcraig, line 329: Walk me through why we need a complete re-implementation here instead of, e.g. stripping out a unit and delegating most of it to `super`'s method.

ccdproc/_ccddata_wrapper_for_array_api.py
mwcraig: Agreed — the current docstring is a prose note with no ``Parameters``/``Returns``. Proposed replacement, following the numpy-style used by ``propagate`` above:

```python
        """
        Propagate uncertainty for multiplication or division.

        This is astropy's
        ``_VariancePropagationMixin._propagate_multiply_divide`` with the
        NumPy calls replaced by their array-namespace equivalents; see the
        astropy version for the derivation of the formulae. Unlike astropy's
        version this does not convert the uncertainties between units,
        because ``_CCDDataWrapperForArrayAPI._arithmetic_wrapper`` removes
        the units from the uncertainties before doing the arithmetic.

        Parameters
        ----------
        other_uncert : `~astropy.nddata.NDUncertainty`
            The uncertainty of the other operand. Its ``array`` and
            ``parent_nddata.data`` must be in the same array namespace as
            ``self.array``.
        result_data : array-like
            Accepted only for signature compatibility with astropy; the
            formulae do not use it.
        correlation : float or array-like
            Correlation coefficient between the two operands, ``0`` for
            uncorrelated.
        divide : bool, optional
            ``True`` for division, ``False`` (default) for multiplication.
        to_variance : callable, optional
            Converts the stored uncertainty array to a variance. Defaults to
            the identity, i.e. the uncertainty is already a variance.
        from_variance : callable, optional
            Converts a variance back to the stored uncertainty type. Defaults
            to the identity.

        Returns
        -------
        array-like
            The propagated uncertainty array, in the same array namespace and
            on the same device as the inputs, in the representation of
            ``self`` (as determined by ``from_variance``).
        """
```

Not applied on the branch yet; say the word and I'll push it.

CHANGES.rst
mwcraig: Agreed — at five lines it's the longest entry in the section. Proposed four-line version that keeps all three fixes (the integer-gain cast is real: `array-api-strict` refuses `int_array * float_array`, so `gain_correct(ccd, 2)` failed there before this PR):

```rst
- Fix ``gain_correct`` and ``flat_correct`` for images on a non-default
  device: put the gain and flat normalization on the device of the data, cast
  an integer gain to float, and keep the uncertainty propagation for
  multiplication and division out of NumPy. [#993]
```

Not applied on the branch yet.

ccdproc/_ccddata_wrapper_for_array_api.py
mwcraig: Short version: stripping the units is necessary but not sufficient. Astropy's `_propagate_multiply_divide` (astropy 8.0.1, `nduncertainty.py`) has two **unconditional** `np.abs` calls on the data arrays that `_propagate_add_sub` doesn't have, and those are what force the NumPy conversion.

Every `np.` call in the two astropy methods:

| method | line | call | when it runs |
|---|---|---|---|
| `_propagate_add_sub` | 741–742 | `isinstance(correlation, np.ndarray)`, `np.sqrt(this * other)` | only if `correlation != 0` |
| `_propagate_multiply_divide` | 813 | `right = np.abs(self.parent_nddata.data**2 * d_b)` | **always** (whenever `other` has an uncertainty) |
| `_propagate_multiply_divide` | 832 | `left = np.abs(other_uncert.parent_nddata.data**2 * d_a)` | **always** (whenever `self` has an uncertainty) |
| `_propagate_multiply_divide` | 864, 868 | `isinstance(correlation, np.ndarray)`, `np.sqrt(d_a * d_b)` | only if `correlation != 0` |

So add/sub can delegate to `super()` because its only NumPy call sits behind the correlation branch, which `_arithmetic_wrapper` never takes with the default `uncertainty_correlation=0`. Multiply/divide can't, because `np.abs` runs on every call — and `np.abs` on an array it can't coerce isn't a slow path, it's an exception.

Empirically, with units already stripped exactly as `_arithmetic_wrapper` does, on `array_api_strict.Device("device1")`:

```
astropy _propagate_multiply_divide  → RuntimeError: Can't convert array on the
    'array_api_strict.Device('device1')' device to a NumPy array.
    (nduncertainty.py:813  right = np.abs(self.parent_nddata.data**2 * d_b))
astropy _propagate_add_sub          → succeeds, result stays on device1
ccdproc _propagate_multiply         → succeeds, result stays on device1
```

Alternatives I considered and why not:

- **Monkeypatch `np.abs` in astropy's module and delegate.** Works for `correlation=0`, then fails at line 868 (`np.sqrt`) for nonzero correlation — so it's a global patch of astropy's namespace that still leaves a hole. Not something I'd want in library code.
- **Delegate and post-process.** There is nothing to post-process: the failure is inside the method, before it returns.
- **Extract only the two `abs` lines.** They're interleaved with the `d_a`/`d_b` computation that the correlation term reuses, so a "partial" override would have to duplicate everything except the final `from_variance` line anyway.

The unit-conversion branches (lines 800–811, 819–830) are a separate hazard — they call `to_variance()` on `Unit`/`Quantity`, which `xp.square` rejects, and on JAX they fail earlier still with `AttributeError: 'ArrayImpl' object has no attribute 'to'`. That's the part unit-stripping *does* fix, which is why the reimplementation omits those branches entirely rather than carrying dead code.

Repro script (uses `.tox/strict`): I can paste it into a test or a gist if you want it preserved.

ccdproc/_ccddata_wrapper_for_array_api.py
mwcraig: Applied in 8f36710.

CHANGES.rst
mwcraig: Applied in 8f36710.

PR #994, opened 2026-08-25
Compute Combiner.clip_extrema mask with rank comparison instead of scatter

mwcraig:
Part of #971 (the clip_extrema bucket).

`array_api_strict` strict test run on top of `main` (711bb26): **41 failed → 37 failed**, no new failures on any of the four backends (numpy, jax, dask, strict). The failure list after this change is identical to `main`'s minus the four `clip_extrema` tests.

`Combiner.clip_extrema` masked the `nlow` lowest and `nhigh` highest per-pixel values by building flat integer indices by hand and scattering into the mask with `xpx.at(...)[flat_index].set(True)` -- the only integer-array `xpx.at(...).set` call anywhere in the codebase. `array_api_strict` rejects fancy-indexed `__setitem__`, and there is no `put_along_axis` in the array API standard, so this raised `IndexError` on strict.

A second `argsort` along axis 0 inverts the first: `ranks[k, ...]` is the rank of image k value at each pixel, so comparing ranks against `nlow`/`nhigh` produces the identical mask without any scatter. Both `argsort` calls default to `stable=True`, so tie-breaking is unchanged from the old code -- verified against a hand-written reference implementation across 8400 random trials (with ties and NaNs, several shapes) run against the new code path, 0 mismatches. This also let the `numpy.mgrid`-based coordinate construction and the now-unused `from numpy import mgrid as np_mgrid` import be deleted.

`test_clip_extrema_keeps_indices_in_array_namespace`, which existed to spy on the removed scatter indices via a monkeypatched `xpx.at`, no longer has anything to spy on. It is replaced with two tests: one checking the resulting mask namespace/device (reusing the same tied `(3, 2, 3)` fixture data and expected mask), and one checking the mask content directly against a documented 3-way tie.

The four target strict failures (`test_clip_extrema_3d`, `_alone`, `_via_combine`, `_with_other_rejection`) all pass; numpy, jax (`JAX_ENABLE_X64=True`), and dask are unaffected (same pass/skip/xfail counts as `main`, no failures). Ruff clean.

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -18,7 +18,6 @@ from astropy import log
 from astropy.nddata import CCDData, StdDevUncertainty
 from astropy.stats import sigma_clip
 from astropy.utils import deprecated_renamed_argument
-from numpy import mgrid as np_mgrid
 
 from ._nanfuncs import nanmean, nanmedian, nanstd, nansum
 from .core import _native_numpy, sigma_func
@@ -358,6 +357,9 @@ class Combiner:
         will count the masking of that pixel toward the count of nhigh masked
         pixels.
 
+        If ``nlow`` or ``nhigh`` is at least the number of images, every pixel
+        is masked.
+
         Here is a copy of the relevant IRAF help text [0]_:
 
         nlow = 1, nhigh = (minmax)
@@ -379,33 +381,17 @@ class Combiner:
         if nhigh is None:
             nhigh = 0
 
+        n_images = self._data_arr.shape[0]
+        # argsorted[i, ...] is, per pixel, the index of the image whose value
+        # has rank i (0 = lowest, n_images - 1 = highest). Sorting those
+        # indices again inverts the permutation: ranks[k, ...] is the rank of
+        # image k's value at each pixel. Comparing ranks avoids scattering
+        # into the mask with per-pixel indices, which the array API standard
+        # does not support.
         argsorted = xp.argsort(self._data_arr, axis=0)
-        # Not every array package has mgrid, so make it in numpy and convert it to the
-        # array package used for the data.
-        mg = xp.asarray(
-            np_mgrid[
-                [slice(ndim) for i, ndim in enumerate(self._data_arr.shape) if i > 0]
-            ],
-            device=array_api_compat.device(self._data_arr),
-        )
-        for i in range(-1 * nhigh, nlow):
-            coordinates = [xp.reshape(argsorted[(i, ...)], (-1,))]
-            coordinates.extend(
-                xp.reshape(mg[(axis, ...)], (-1,)) for axis in range(mg.shape[0])
-            )
-            # Some array libraries don't support indexing with arrays in multiple
-            # dimensions, so we need to flatten the mask array, set the mask
-            # values for a flattened array, and then reshape it back to the
-            # original shape.
-            flat_index = coordinates[0]
-            for coordinate, dimension in zip(
-                coordinates[1:], self._data_arr.shape[1:], strict=True
-            ):
-                flat_index = flat_index * dimension + coordinate
-            self._data_arr_mask = xp.reshape(
-                xpx.at(xp.reshape(self._data_arr_mask, (-1,)))[flat_index].set(True),
-                self._data_arr.shape,
-            )
+        ranks = xp.argsort(argsorted, axis=0)
+        clip = (ranks < nlow) | (ranks >= n_images - nhigh)
+        self._data_arr_mask = self._data_arr_mask | clip
 
     # set up min/max clipping algorithms
     def minmax_clipping(self, min_clip=None, max_clip=None):
--- a/ccdproc/tests/test_combiner.py
+++ b/ccdproc/tests/test_combiner.py
@@ -1226,21 +1226,6 @@ def test_3d_combiner_with_scaling():
     assert avg_ccd.shape == ccd_data.shape
 
 
-def _clip_extrema_combiner(data):
-    if xp.__name__ == "array_api_strict":
-        # Bypass Combiner.__init__ only for array-api-strict, which cannot stack
-        # a list of arrays. This isolates clip_extrema from that separate limit.
-        combiner = object.__new__(Combiner)
-        combiner._xp = xp
-        combiner._data_arr = xp.asarray(data, device=xp_device)
-        combiner._data_arr_mask = xp.zeros(
-            combiner._data_arr.shape, dtype=xp.bool, device=xp_device
-        )
-    else:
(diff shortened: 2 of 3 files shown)


ccdproc/tests/test_combiner.py
```diff
@@ -1226,13 +1226,7 @@ def test_3d_combiner_with_scaling():
...
 
-def test_clip_extrema_keeps_indices_in_array_namespace(monkeypatch):
-    data = [
-        [[9.0, 2.0, 7.0], [4.0, 8.0, 1.0]],
-        [[3.0, 6.0, 5.0], [9.0, 2.0, 8.0]],
-        [[6.0, 1.0, 4.0], [2.0, 7.0, 3.0]],
-    ]
-
+def _clip_extrema_combiner(data):
     if xp.__name__ == "array_api_strict":
```
mwcraig: Why do we still need this guard? I thought the point of this PR was to fix clip_extrema

ccdproc/combiner.py
```diff
@@ -379,33 +381,17 @@ def clip_extrema(self, nlow=0, nhigh=0):
...
         argsorted = xp.argsort(self._data_arr, axis=0)
-        # Not every array package has mgrid, so make it in numpy and convert it to the
-        # array package used for the data.
-        mg = xp.asarray(
-            np_mgrid[
-                [slice(ndim) for i, ndim in enumerate(self._data_arr.shape) if i > 0]
-            ],
-            device=array_api_compat.device(self._data_arr),
-        )
-        for i in range(-1 * nhigh, nlow):
```
mwcraig, line 391: Why was this bit of nasty code ever needed?

ccdproc/combiner.py
```diff
@@ -379,33 +381,17 @@ def clip_extrema(self, nlow=0, nhigh=0):
         if nhigh is None:
             nhigh = 0
 
+        n_images = self._data_arr.shape[0]
+        # argsorted[i, ...] is, per pixel, the index of the image whose value
+        # has rank i (0 = lowest, n_images - 1 = highest). Sorting those
+        # indices again inverts the permutation: ranks[k, ...] is the rank of
+        # image k's value at each pixel. Comparing ranks avoids scattering
```
mwcraig, line 388: Explain this more in your reply to this (no need to make the comment longer)

ccdproc/tests/test_combiner.py
mwcraig: We don't — it's stale. The guard was carried over from the #928 test (`a4ffcef`, 2026-07-19), when `Combiner.__init__` still built the stack with a nested `xp.asarray([...])` that array-api-strict rejects. That was fixed by `xp.stack` in `ebfe8de` (2026-08-22, the #965 item), so the branch has been dead since; this PR only preserved it because the plan said to reuse the old construction verbatim.

Checked on this branch with the same three images, on both strict's default device and `device1`:

```
Combiner([CCDData(xp.asarray(im, device=...), unit=u.adu) for im in data]).clip_extrema(nlow=1, nhigh=1)
```

`__init__` and `clip_extrema` both succeed and the mask stays in `array_api_strict` on the input device. `test_clip_extrema_3d` and friends already construct their `Combiner` the normal way on strict and pass.

So `_clip_extrema_combiner` can go entirely; each test just does

```python
c = Combiner([CCDData(xp.asarray(image, device=xp_device), unit=u.adu) for image in data])
```

(`device=xp_device` so the device assertion in the first test still exercises the non-default device on strict.) Happy to push that.

ccdproc/combiner.py
mwcraig: It grew in three layers, each one a workaround for the backend of the day:

1. **2016 (`74b71a3`)** — plain numpy. "Mask the *i*-th ranked image at every pixel" is a scatter: for each pixel `p`, set `mask[argsorted[i, p], p] = True`. numpy had no `put_along_axis` yet (added in 1.15, 2018), so the vectorised way was multi-dimensional fancy indexing: build the pixel coordinates with `np.mgrid`, ravel them, and do `mask[(argsorted[i].ravel(), mg[0].ravel(), mg[1].ravel())] = True`. That is the origin of `mgrid` and the `where` tuple.
2. **Array-API port** — same idea with the in-place assignment swapped for `xpx.at(mask)[where].set(True)`, and `mgrid` built in numpy then moved to the namespace because the standard has no `mgrid`.
3. **Dask port (`dfc98cf`, 2025-06)** — dask cannot assign through a *multi-dimensional* integer index, so the tuple was collapsed by hand into a single flat index into a ravelled mask, then reshaped back. #928 (`a4ffcef`) later rewrote that `ravel_multi_index` in namespace arithmetic so the indices did not leak to numpy.

Each step kept the scatter and made the *index* more portable; but the array API standard supports no integer-array `__setitem__` at all, so there was no fourth step in that direction. The rank comparison sidesteps the scatter entirely — no `mgrid`, no flattening, no `xpx.at` — which is why the whole block could go.

ccdproc/combiner.py
mwcraig: Worked through one pixel. Say four images hold `[5, 1, 3, 4]` at that pixel:

- `argsorted = argsort(values)` = `[1, 2, 3, 0]` — position *i* holds *which image* has the *i*-th smallest value. The old code used this directly: "at rank `i`, scatter `True` into image `argsorted[i]`".
- `ranks = argsort(argsorted)` = `[3, 0, 1, 2]` — position *k* now holds *what rank image k has*. That works because `argsorted` is a permutation of `0..n-1`, and `argsort` of a permutation is its inverse: sorting `[1, 2, 3, 0]` puts `0` first (it sits at position 3), then `1` (position 0), … giving `[3, 0, 1, 2]`.

With `nlow=1, nhigh=1`: `ranks < 1` picks image 1 (the 1), `ranks >= 4 - 1` picks image 0 (the 5) — exactly the images the loop `for i in range(-nhigh, nlow)` would have scattered into, but computed as a boolean array of the mask's own shape, so it just gets OR-ed in. Doing it on `axis=0` applies this to every pixel at once.

Equivalence, not just resemblance: the old code marked image `argsorted[i, p]` at pixel `p` for every `i` in that range, and the new code marks image `k` iff `ranks[k, p]` is in that range — those are the same set for *any* output of the first `argsort`, so tie-breaking is inherited unchanged (both `array_api_compat` and strict default to `stable=True`, which is what the tie test pins). I also checked it empirically before opening the PR: 8400 random stacks (3-D and 4-D, with ties and NaNs, all `nlow + nhigh <= n - 1`) give bit-identical masks to `main`.

Cost: two `argsort`s along axis 0 instead of one `argsort` plus a Python loop of `nlow + nhigh` scatters — same `O(n log n)` per pixel, and no per-iteration reshapes.

PR #995, opened 2026-08-25
Fix array-namespace dtype handling in ccd_process, transform_image, combine, and ImageFileCollection

mwcraig:
Part of #971 (the "namespace dtypes" bucket).

Strict-backend count on top of `main` `711bb26`: **41 failed -> 35 failed** (456 -> 463 passed; 43 xfailed, 5 xpassed unchanged). No test regresses on strict, numpy, jax, or dask.

`ccd_process`, `transform_image`, `combine()`, and `ImageFileCollection.ccds()`/`data()` all leaked NumPy-specific dtype conventions across the array-namespace boundary: a builtin `bool` instead of `xp.bool` for `ccd_process`'s bad-pixel mask, running a transform function directly on a boolean mask in `transform_image`, and handing a NumPy dtype *object* (rather than a plain array) to `xp.asarray(..., dtype=...)` when converting FITS data to a requested namespace in `combine()` and `ImageFileCollection`. Each is fixed to speak the array API idiom already used elsewhere in the codebase (`xp.bool`, casting the mask to the data dtype before transforming, converting to native byte order and letting `xp.asarray` do a plain conversion). A new `_native_numpy` helper in `core.py`, next to `_is_array`, is shared by `combine()` and `ImageFileCollection`. `ImageFileCollection`'s mask conversion also had to switch from the public `.mask` setter to the private `_mask` attribute, since `CCDData`'s public setter unconditionally converts its value back to NumPy - the same workaround `ccd_process` already used, with a `TODO` to remove it once `CCDData` supports array namespaces.

Fixing `ccd_process`'s mask dtype uncovered a second, previously-masked bug: `log_meta.py`'s `_replace_array_with_placeholder` (used by the `@log_to_metadata` decorator to build the auto-logging string) crashed for a bare, non-`NDData` array-API array that does not implement `__len__` - which `array_api_strict` deliberately does not, per the array API standard. It fell through to `value.data`, which such an array does not have, raising `AttributeError` instead of the `TypeError` the code was written to catch. That crash was masking `ccd_process`'s tests even after the mask-dtype fix, so it is fixed here too (broadened the `except` to also catch `AttributeError`).

Test-body fixes bundled with the corresponding source fix: `test_transform_image` discarded the return value of `xpx.at(...).set(...)` (a silent no-op on JAX/strict, the same class of bug as #963) and did not build a real boolean mask; both are fixed, and the test now asserts the transformed mask is `True` only at the expected pixel. `test_ccd_process`/`test_ccd_process_gain_corrected` built their bad-pixel mask as a float array and compared it against the (now boolean) result with `xpx.isclose`, which strict namespaces reject for non-numeric dtypes; both now build a boolean mask and compare with `==`. A new `test_generator_ccds_converts_to_array_namespace` in `test_image_collection.py` exercises the same FITS-to-namespace conversion in `ImageFileCollection.ccds()` that `combine()` already had a regression test for (A4 in the working plan; not on the original strict-failure list since no existing test passed `array_package=` to `ImageFileCollection`, but the code path is byte-for-byte the same pattern as `combine()`, so it is included here).

One target test remains failing on strict for a reason outside this PR's scope: `test_combine_ccd_with_uncertainty_and_mask_from_fits[function]` reaches a second, unrelated bug once this PR's fix to `combine()` clears the first one - `test_combiner.py`'s `_make_mean_scaler` calls `ccd_data.data.mean()` as a method, which `array_api_strict` arrays do not support (only the top-level `xp.mean` function form is part of the standard). That is a bug in `test_combiner.py`, a file intentionally left untouched here to avoid conflicting with a parallel PR (fixing `Combiner.clip_extrema`) also touching that file; `[mean]` (the other parametrization) passes.

### DIFF
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -98,6 +98,32 @@ def _is_array(arr):
     return True
 
 
+def _native_numpy(arr):
+    """
+    Return a NumPy array with the same data in native byte order.
+
+    Data read from a FITS file is big-endian (e.g. dtype ``>f8``), a byte
+    order that only NumPy understands. Non-NumPy array namespaces (and
+    ``array_api_strict``) reject such a dtype outright, or, if the dtype
+    happens to be native-endian already, warn when it is compared against
+    their own dtype objects. Converting to native byte order first and
+    handing the resulting plain NumPy array to ``xp.asarray`` avoids both
+    problems.
+
+    Parameters
+    ----------
+    arr : `numpy.ndarray`
+        A NumPy array, possibly in non-native byte order.
+
+    Returns
+    -------
+    `numpy.ndarray`
+        ``arr`` converted to its dtype's native scalar type. No copy is made
+        if one is not required.
+    """
+    return arr.astype(arr.dtype.type, copy=False)
+
+
 def _percentile_fallback(array, percentiles, xp=None):
     """
     Try calculating percentile using namespace, otherwise fall back to
@@ -380,7 +406,7 @@ def ccd_process(
         # TODO: the private _mask attribute is set here to avoid the
         # mask.setter than sets the mask to a numpy array. This can be
         # removed when CCDData supports array namespaces.
-        nccd._mask = xp.asarray(bad_pixel_mask, dtype=bool)
+        nccd._mask = xp.asarray(bad_pixel_mask, dtype=xp.bool)
     else:
         raise TypeError("bad_pixel_mask is not None or an array.")
 
@@ -1147,6 +1173,9 @@ def transform_image(ccd, transform_func, **kwargs):
     # make a copy of the object
     _nccd = _ccd.copy()
 
+    # Set array namespace
+    xp = array_api_compat.array_namespace(_nccd.data)
+
     # transform the image plane
     try:
         _nccd.data = transform_func(_nccd.data, **kwargs)
@@ -1161,7 +1190,11 @@ def transform_image(ccd, transform_func, **kwargs):
 
     # transform the mask plane
     if _nccd.mask is not None:
-        mask = transform_func(_nccd.mask, **kwargs)
+        # Cast the boolean mask to the data's dtype before transforming it;
+        # most transform functions (e.g. scipy.ndimage.shift) do not accept
+        # boolean input, so treat the mask as numeric and re-threshold the
+        # result below.
+        mask = transform_func(xp.astype(_nccd.mask, _nccd.data.dtype), **kwargs)
         _nccd.mask = mask > 0
 
     if _nccd.wcs is not None:
--- a/ccdproc/core.py
+++ b/ccdproc/core.py
@@ -102,15 +102,6 @@ def _native_numpy(arr):
     """
     Return a NumPy array with the same data in native byte order.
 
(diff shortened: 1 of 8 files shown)


ccdproc/core.py
```diff
@@ -98,6 +98,33 @@ def _is_array(arr):
     return True
 
 
+def _native_numpy(arr):
+    """
+    Return a NumPy array with the same data in native byte order.
+
+    Data read from a FITS file is big-endian (e.g. dtype ``>f8``), a byte
```
mwcraig: Move this to notes

ccdproc/core.py
```diff
@@ -98,6 +98,33 @@ def _is_array(arr):
     return True
 
 
+def _native_numpy(arr):
```
mwcraig, line 101: Does doing this conversion have any effects on the computational results we get in numpy or any other backend?

ccdproc/core.py
mwcraig: Agreed — that paragraph is rationale, not a summary, so it belongs under a numpy-style `Notes` section. Proposed shape:

```python
def _native_numpy(arr):
    """
    Return a NumPy array with the same data in native byte order.

    Parameters
    ----------
    arr : `numpy.ndarray`
        A NumPy array, possibly in non-native byte order.

    Returns
    -------
    `numpy.ndarray`
        ``arr`` converted to its dtype's native scalar type. No copy is made
        if one is not required.

    Notes
    -----
    Data read from a FITS file is big-endian (e.g. dtype ``>f8``), a byte
    order that only NumPy understands; other array namespaces reject such
    an array outright. Requesting ``dtype=arr.dtype.type`` from
    ``xp.asarray`` works around that in NumPy and JAX but hands a NumPy
    type object to the namespace, which ``array_api_strict`` warns about.
    Converting to native byte order in NumPy first and handing the
    resulting plain array to ``xp.asarray`` with no ``dtype`` avoids both
    problems.
    """
    return arr.astype(arr.dtype.type, copy=False)
```

Happy to push that once you've settled the other thread.

ccdproc/core.py
mwcraig: No — the values, dtype width, and copy behaviour are all unchanged from the `dtype=arr.dtype.type` path this replaces; only the memory byte order differs, and only for non-native input.

Checked against the old expression (`xp.asarray(arr, dtype=arr.dtype.type)`) for `>f8`, `<f8`, `>f4`, `>i2`, `>u2` and `|b1` inputs, with NaN and inf in the float cases:

- **numpy**: results are bitwise identical (`tobytes()` equal, same dtype). For native-order input `astype(..., copy=False)` returns the *same object*, so `np.asarray` hands back the original array exactly as before; for big-endian input the old path also had to copy (asarray with a differently-ordered dtype can't return a view), so no extra copies either way.
- **jax**: `array_equal(equal_nan=True)` and same dtype for every case.
- **array-api-strict**: same values and dtype as the numpy result (the old path raised `UserWarning` for all six inputs).

Dtype width is preserved because `arr.dtype.type` is the native scalar type of the *same* width (`>f4` → `float32`, `>i2` → `int16`), not a promotion to float64. And the helper only runs on data just read from FITS in `combine()` and `ImageFileCollection.ccds()/data()`, so nothing computed in the namespace is affected.

PR #997, opened 2026-08-25
Normalise combine()'s raw-module array_package and fix correlated add/subtract uncertainty leak

### LINKED ISSUE #982: combine() does not normalise a raw module passed as array_package
In `combine()` (`ccdproc/combiner.py:1055-1059`) the requested array package is handled with

```python
if array_package is not None:
    try:
        xp = array_api_compat.array_namespace(array_package)
    except TypeError:
        xp = array_package
```

`array_api_compat.array_namespace()` takes *arrays*, not modules, so for a module it always raises `TypeError` (`module is not a supported array type` — verified for `numpy` and `dask.array`) and the fallback hands the raw module through unchanged. The intent was clearly to normalise to the array-api-compat namespace; as written the normalisation never happens.

A raw module then reaches code that uses array-API-only features (`xp.bool`, the `device=` keyword in `xp.asarray`/`xp.zeros`, `xp.astype`, ...) and fails, e.g. `combine(files, array_package=dask.array)` dies with `TypeError: from_array() got an unexpected keyword argument 'device'` (on `main` before #976 it died one call later in `clip_extrema` with `module 'dask.array' has no attribute 'argsort'`).

PR #976 fixed the same problem in `Combiner.__init__` by normalising with

```python
xp = array_api_compat.array_namespace(xp.asarray(0))
```

`combine()` should do the same so the two entry points agree on what `array_package` / `xp` may be, and the `array_package` docstring should say a plain module is accepted and converted.

Regression test: `combine([...numpy CCDData...], array_package=np)` and, ideally, `array_package=dask.array` on a FITS-file input so the `CCDData.read` branch that converts data/mask/uncertainty into `xp` is exercised.

Follow-up from the review of #976.

mwcraig:
Part of #971; fixes #982.

Strict count on top of `main` `6724c8e`: **31 failed → 31 failed** (identical failure list; no target from #971 was on the strict failure list). numpy, jax, dask: no regressions; strict/numpy/jax/dask each gain the 8 new regression tests as passes.

**combine()'s raw-module `array_package` (#982).** `combine()` tried to normalise `array_package` with `array_api_compat.array_namespace()`, which only accepts arrays, not modules, so for a raw module (e.g. `numpy` or `dask.array`) it always raised `TypeError` and the `except` branch passed the module through unnormalised instead of converting it. `Combiner.__init__` already normalises its own `xp` argument with `array_api_compat.array_namespace(xp.asarray(0))` (fixed for #976); this PR does the same in `combine()` so the two entry points agree on what a caller may pass, and updates the `array_package` docstring to say so. While adding the regression test for the `from_array() got an unexpected keyword argument 'device'` failure described in #982, I found that `combine()`'s subsequent `xp = array_api_compat.array_namespace(ccd.data)` (which runs right after the first image's data is converted, whether or not `array_package` was itself normalised) already re-derives a correct namespace from the resulting concrete array — I could not reproduce the specific crash from #982 through `combine()` on current `main`, including with a masked, uncertain, scaled, tiled combination that exercises every `device=` call in the function. The crash went inert with #995: the conversion is now `xp.asarray(_native_numpy(...))` with no `device=`/`dtype=`, which a raw `dask.array` handles, and `xp` is then re-derived from the converted array. This PR still closes the underlying gap between intent and code (and between `Combiner`/`combine`), it just is not observed to fix a live crash on current `main`.

**Arrays as `array_package`.** The old code (and docstring) also accepted an array standing in for its namespace; a third commit keeps that working (`is_array_api_obj` check before calling `.asarray(0)`) and covers it in the regression test.

**Correlated add/subtract uncertainty leaks to NumPy in the wrapper.** `_ArrayAPIPropagationMixin._propagate_add`/`_propagate_subtract` delegated correlated-uncertainty math to astropy's `_VariancePropagationMixin._propagate_add_sub`, whose correlation term (`2 * correlation * np.sqrt(this * other)`) is hardcoded to NumPy. That is invisible when `uncertainty_correlation` is `0` (the term is never evaluated), but addition/subtraction with a nonzero correlation on strict fails with `TypeError: Expected Array or Python scalar; got numpy.ndarray` once the NumPy result is combined with an array-API array. #993 already fixed the same class of leak for `_propagate_multiply_divide`; this PR mirrors it by adding an array-namespace `_propagate_add_sub` to the mixin and calling it instead of `super()`'s, extending `test_wrapped_arithmetic_correlated_uncertainty` from `["multiply", "divide"]` to `["add", "subtract", "multiply", "divide"]`.

**Follow-up note.** `Combiner(xp=...)` accepts an explicit namespace for `CCDData` input, but `combine(array_package=...)` is documented as, and remains, ignored when `ccd_list` is already a list of `CCDData` objects — `array_package` only applies when reading from filenames. This asymmetry is unchanged by this PR; flagging it as a possible follow-up.

### DIFF
--- a/ccdproc/_ccddata_wrapper_for_array_api.py
+++ b/ccdproc/_ccddata_wrapper_for_array_api.py
@@ -314,14 +314,6 @@ class _ArrayAPIPropagationMixin:
         """
         Propagate uncertainty for addition or subtraction.
 
-        This is astropy's ``_VariancePropagationMixin._propagate_add_sub``
-        with the NumPy call replaced by its array-namespace equivalent; see
-        the astropy version for the derivation of the formulae. Unlike
-        astropy's version this does not convert the uncertainties between
-        units, because ``_CCDDataWrapperForArrayAPI._arithmetic_wrapper``
-        removes the units from the uncertainties before doing the
-        arithmetic.
-
         Parameters
         ----------
         other_uncert : `~astropy.nddata.NDUncertainty`
@@ -348,6 +340,16 @@ class _ArrayAPIPropagationMixin:
             The propagated uncertainty array, in the same array namespace and
             on the same device as the inputs, in the representation of
             ``self`` (as determined by ``from_variance``).
+
+        Notes
+        -----
+        This is astropy's ``_VariancePropagationMixin._propagate_add_sub``
+        with the NumPy call replaced by its array-namespace equivalent; see
+        the astropy version for the derivation of the formulae. Unlike
+        astropy's version this does not convert the uncertainties between
+        units, because ``_CCDDataWrapperForArrayAPI._arithmetic_wrapper``
+        removes the units from the uncertainties before doing the
+        arithmetic.
         """
         del result_data  # accepted only for compatibility with astropy
         xp = array_api_compat.array_namespace(self.array, other_uncert.array)
@@ -408,14 +410,6 @@ class _ArrayAPIPropagationMixin:
         """
         Propagate uncertainty for multiplication or division.
 
-        This is astropy's
-        ``_VariancePropagationMixin._propagate_multiply_divide`` with the
-        NumPy calls replaced by their array-namespace equivalents; see the
-        astropy version for the derivation of the formulae. Unlike astropy's
-        version this does not convert the uncertainties between units,
-        because ``_CCDDataWrapperForArrayAPI._arithmetic_wrapper`` removes
-        the units from the uncertainties before doing the arithmetic.
-
         Parameters
         ----------
         other_uncert : `~astropy.nddata.NDUncertainty`
@@ -443,6 +437,16 @@ class _ArrayAPIPropagationMixin:
             The propagated uncertainty array, in the same array namespace and
             on the same device as the inputs, in the representation of
             ``self`` (as determined by ``from_variance``).
+
+        Notes
+        -----
+        This is astropy's
+        ``_VariancePropagationMixin._propagate_multiply_divide`` with the
+        NumPy calls replaced by their array-namespace equivalents; see the
+        astropy version for the derivation of the formulae. Unlike astropy's
+        version this does not convert the uncertainties between units,
+        because ``_CCDDataWrapperForArrayAPI._arithmetic_wrapper`` removes
+        the units from the uncertainties before doing the arithmetic.
         """
         del result_data  # accepted only for compatibility with astropy
         xp = array_api_compat.array_namespace(self.array, other_uncert.array)
--- a/ccdproc/_ccddata_wrapper_for_array_api.py
+++ b/ccdproc/_ccddata_wrapper_for_array_api.py
@@ -281,7 +281,7 @@ class _ArrayAPIPropagationMixin:
     def _propagate_add(self, other_uncert, result_data, correlation):
         xp = array_api_compat.array_namespace(self.array, other_uncert.array)
         to_variance, from_variance = self._variance_hooks(xp)
-        return super()._propagate_add_sub(
+        return self._propagate_add_sub(
             other_uncert,
             result_data,
             correlation,
@@ -293,7 +293,7 @@ class _ArrayAPIPropagationMixin:
     def _propagate_subtract(self, other_uncert, result_data, correlation):
         xp = array_api_compat.array_namespace(self.array, other_uncert.array)
         to_variance, from_variance = self._variance_hooks(xp)
-        return super()._propagate_add_sub(
+        return self._propagate_add_sub(
             other_uncert,
             result_data,
             correlation,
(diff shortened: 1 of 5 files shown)


ccdproc/_ccddata_wrapper_for_array_api.py
```diff
@@ -302,6 +302,76 @@ def _propagate_subtract(self, other_uncert, result_data, correlation):
...
+        result_data,
+        correlation,
+        subtract=False,
+        to_variance=lambda x: x,
+        from_variance=lambda x: x,
+    ):
+        """
+        Propagate uncertainty for addition or subtraction.
+
+        This is astropy's ``_VariancePropagationMixin._propagate_add_sub``
```
mwcraig: move to notes

ccdproc/combiner.py
```diff
@@ -1013,10 +1013,12 @@ def combine(
 
         If not specified, the array package used will
         be numpy. The array package can be specified either by passing in
-        an array namespace (e.g. output from ``array_api_compat.array_namespace``),
-        or an imported array package that follows the array API standard
-        (e.g. ``numpy`` or ``jax.numpy``), or an array whose namespace can be
-        determined (e.g. a `numpy.ndarray` or ``jax.numpy.ndarray``).
+        an array namespace (e.g. output from ``array_api_compat.array_namespace``)
```
mwcraig: Can this be shortened?

ccdproc/combiner.py
```diff
@@ -1069,10 +1071,14 @@ def combine(
...
-            try:
-                xp = array_api_compat.array_namespace(array_package)
-            except TypeError:
-                xp = array_package
+            # ``array_package`` may be an array, or a raw module such as
+            # ``numpy`` or ``dask.array``; normalise either to the
+            # array-api-compat namespace the same way ``Combiner.__init__``
+            # does, so the conversions below can rely on array-API features
+            # (e.g. the ``device`` keyword) that a raw module may not provide.
+            if not array_api_compat.is_array_api_obj(array_package):
```
mwcraig: Explain in your reply to this why this is necessary.

ccdproc/_ccddata_wrapper_for_array_api.py
mwcraig: Agreed — that paragraph is background, not the summary line, so it belongs under a numpydoc `Notes` section after `Returns`:

```
        Returns
        -------
        array-like
            ...

        Notes
        -----
        This is astropy's ``_VariancePropagationMixin._propagate_add_sub``
        with the NumPy call replaced by its array-namespace equivalent; see
        the astropy version for the derivation of the formulae. Unlike
        astropy's version this does not convert the uncertainties between
        units, because ``_CCDDataWrapperForArrayAPI._arithmetic_wrapper``
        removes the units from the uncertainties before doing the
        arithmetic.
```

`_propagate_multiply_divide` (from #993, already on `main`) has the identical paragraph in the identical place, so I'd move both in the same commit to keep the two methods parallel. Happy to push that when you say so.

ccdproc/combiner.py
mwcraig: Yes. The three-way enumeration plus the "whichever is given" sentence can collapse to:

```
    array_package : array namespace, module or array, optional
        The array package to use for data read in from files; ignored if
        ``ccd_list`` is already a list of `~astropy.nddata.CCDData` objects.
        Either an array namespace, a plain module that follows the array API
        standard (e.g. ``numpy`` or ``dask.array``), or an array from that
        package; it is normalised to its array-api-compat namespace the same
        way `~ccdproc.Combiner` handles ``xp``. Default is NumPy.
```

That is roughly half the length, keeps the "ignored for `CCDData` input" caveat, and matches the wording of the `Combiner` ``xp`` docstring. Can push it with the `Notes` move above.

ccdproc/combiner.py
mwcraig: Two separate things are happening on these lines.

**Why `.asarray(0)` at all.** `array_api_compat.array_namespace()` only accepts *arrays*; for a module it raises `TypeError: module is not a supported array type` (verified with `numpy` and `array_api_strict`, array-api-compat 1.15.0), and `get_namespace` is just an alias of it. There is no public array-api-compat function that maps a module to its compat wrapper. That is exactly the #982 bug: the old `try: array_namespace(array_package) / except TypeError: xp = array_package` always took the `except` branch for a module, so the raw module went through unnormalised. The only way to get from a module to its compat namespace is to make a throwaway array *with* the module and ask for that array's namespace — `array_namespace(module.asarray(0))` — which is the same idiom `Combiner.__init__` uses (`combiner.py:178`, from #976). So `.asarray(0)` is the normalisation step, not an extra.

**Why the `is_array_api_obj` guard.** The old code (and the docstring) also accepted an *array* as `array_package`, because `array_namespace(array)` works directly. An array has no `.asarray` attribute, so calling `.asarray(0)` unconditionally would have turned that documented input into an `AttributeError`. The guard just routes arrays straight to `array_namespace` and only calls `.asarray(0)` on modules. If you would rather drop the array form (`Combiner(xp=...)` does not accept it either), the guard goes away and this becomes the one-liner `xp = array_api_compat.array_namespace(array_package.asarray(0))` — but that is a behaviour change that would want a changelog line.

For completeness: as the PR body says, on current `main` the raw-module path no longer crashes because `xp` is re-derived from `ccd.data` right after the conversion (line 1091) and the conversion itself no longer passes `device=`/`dtype=` since #995. The normalisation here is what guarantees `xp` is a compat namespace *at* the `xp.asarray(_native_numpy(...))` calls, so a later change that adds a keyword back there cannot regress raw-module callers.

mwcraig: > If you would rather drop the array form (Combiner(xp=...) does not accept it either), the guard goes away...

Let's do that

ccdproc/_ccddata_wrapper_for_array_api.py
mwcraig: Done in 46b8c71 — moved into a `Notes` section after `Returns` in both `_propagate_add_sub` and `_propagate_multiply_divide`.

ccdproc/combiner.py
mwcraig: Done in 46b8c71 — docstring shortened as above, minus the array form since that was dropped in the same commit (see the thread below).

ccdproc/combiner.py
mwcraig: Done in 46b8c71. `combine()` now normalises `array_package` with the same one-liner as `Combiner.__init__`:

```python
xp = array_api_compat.array_namespace(array_package.asarray(0))
```

so it accepts an array namespace or a plain module but no longer an array. The array-form check in `test_combine_array_package_raw_module` is removed, the docstring type is now `array namespace or module`, and there is a bullet under *Other Changes and Additions* in `CHANGES.rst` recording the change.

Verified: numpy, jax and dask pass the `array_package` tests; the strict failure set for `test_combiner.py` + `test_ccddata_wrapper_for_array_api.py` is identical before and after (25, all in the #998 bucket).

Issue #933, opened 2026-07-07
Array API: subtract_overscan model fitting goes through numpy-only astropy.modeling


mwcraig:
### Problem

`subtract_overscan` with `model=...` fits via `astropy.modeling`:

https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/core.py#L616

astropy's fitters operate on numpy internally (`LinearLSQFitter` ends up in `np.linalg.lstsq`), so CuPy input hits the implicit-conversion error. This matches the `test_subtract_overscan_model` failures noted in #910.

### Options

- Treat model fitting as a CPU-only operation under the policy in #935 (explicit host transfer + convert the evaluated model back to `xp`).
- For the common low-order polynomial case, an xp-native least-squares fit (`xp.linalg.lstsq` is in the array API linalg extension) could keep everything on-device.
- Longer term: array-API support in `astropy.modeling` (tracked in #940).

---
Found during a review of the array API implementation from #885; follow-up to #909 / #910.

PR #998, opened 2026-08-25
Fix strict array-API test bodies in test_combiner/test_ccdproc/test_image_collection

mwcraig:
Part of #971 (section 2: strict test-body fixes).

Strict count on top of `main` `6724c8e`: **31 failed → 15 failed** (469 passed/36 skipped/43 xfailed/5 xpassed → 483 passed/36 skipped/45 xfailed/5 xpassed). numpy, jax, and dask show no regressions; jax additionally improves (541 passed/7 xfailed → 548 passed/0 xfailed/0 xpassed — see the third commit).

**Add numpy_copy/numpy_ccddata test helpers for strict device1 arrays.** The strict suite runs on array-api-strict's non-default `Device("device1")`, where `np.asarray()` deliberately raises, but several tests need to hand data to `astropy.io.fits`, which requires NumPy. Adds `numpy_copy()` (move to the default device, then convert) and `numpy_ccddata()` (a CCDData with data/mask/uncertainty all converted) to `pytest_fixtures.py` for the write-path fixes below.

**Fix strict-incompatible test bodies in test_combiner.py.** `_make_mean_scaler` called `ndarray.mean()`/`np.ma.average()` directly; rewritten to use each argument's own array namespace, since the reference image can be plain NumPy (read from FITS) while the scaled array is a namespace array. Several tests called `CCDData.multiply()`, iterated over an array, or called `xp.asarray()` on a list of arrays — all of which either route through astropy's arithmetic wrapper (which calls `np.result_type` and breaks on strict) or are flatly disallowed by array-api-strict; replaced with in-namespace construction, `xp.stack()`, and explicit `arr[i, ...]` indexing. `combine()` only honors `array_package` for filename/string input and the `Combiner` reference built from `CCDData.read()` defaults to NumPy, so several tests now pass `array_package=xp`/`Combiner(..., xp=xp)` (or convert the `CCDData` list's `.data` by hand) so both sides of each comparison are namespace arrays. Writing a device1 `CCDData` via `CCDData.write()` — which `ImageFileCollection` round-trips and `combine(output_file=...)` both do — needs `numpy_ccddata()`; comparisons against results that land on the namespace's default device use `numpy_copy()` + `assert_allclose()` instead of `xp.all(...)`.

Two of this bucket's targets still fail after these fixes, on causes outside test-code's reach: `test_combiner_result_dtype`'s `dtype=int` case, because `combine()` passes the bare Python `int` straight to `xp.astype`/`xp.asarray`, which array-api-strict rejects (`AttributeError: type object 'int' has no attribute '_np_dtype'`) — a new gap, not previously reachable because an earlier `.multiply()` failure masked it; and `test_combine_overwrite_output`, because `combine(output_file=...)` writes the still-in-namespace result via `ccd.write()` at `combiner.py:1273` (`AttributeError: 'Array' object has no attribute 'astype'`), the same class of bug as #935. Both are left failing, unmarked, for the coordinator to track in #971. Three more tests that call default `median_combine()` (`test_combiner_with_scaling`, `test_combiner_with_scaling_uncertainty[median_combine]`, `test_writeable_after_combine[median_combine]`) remain failing on #929's `sigma_func`/`astropy.stats` densification, unrelated to and untouched by this PR — `test_writeable_after_combine[median_combine]` was already known to the plan; the other two are additional #929 instances this PR's fixes exposed by clearing their first-listed cause.

Because these fixes remove every `CCDData.multiply()` call from the affected tests, the jax-only `backend_xfail` markers on `test_combiner_with_scaling`, `test_combiner_result_dtype`, `test_combine_overwrite_output`, and `test_combiner_with_scaling_uncertainty` (added for a jax `DeprecationWarning` triggered by astropy arithmetic receiving a jax array as `dtype=`) no longer apply and are removed; jax now XPASSes all of them cleanly.

**Fix test_subtract_overscan_model for strict; xfail the astropy.modeling path.** The test built its overscan gradient with integer/float division array-api-strict rejects, then unpacked the resulting 3D array by iteration, which array-api-strict also disallows for arrays with more than one dimension; fixed with an explicit float64 dtype/device and `[0, ...]`/`[1, ...]` indexing. Also replaced two `ndarray.copy()` calls with `xp.asarray(..., copy=True)`. With those fixed, both parametrizations reach `subtract_overscan(model=...)`, which goes through astropy.modeling's `LinearLSQFitter` — NumPy-only and unrelated to this test's array handling (#933); marked `backend_xfail("array-api-strict", ...)` citing #933. Exact stop: `TypeError: object of type 'Array' has no len()` from `xp.arange(len(oscan))` in `ccdproc/core.py`'s `subtract_overscan`.

**Reset leaked `_config_ccd_requires_unit` before `test_generator_ccds_without_unit`.** astropy's `_arithmetic` decorator sets the module global `_config_ccd_requires_unit = False` around an arithmetic call and restores it only on success, with no try/finally; when an earlier test's arithmetic raises (`test_unit_mismatch_behaves_as_expected` does this deliberately, on every backend), nothing resets the flag before later tests run (astropy/astropy#20268). This previously only reproduced on jax (and, after this PR's strict fixes, on array-api-strict too) and was marked `backend_xfail` for both — but removing several successful `.multiply()` calls in the `test_combiner.py` commit above also removed incidental resets of that global, which surfaced the same leak as a genuine regression on plain NumPy. Rather than add NumPy to the xfail list, the test now resets `astropy.nddata.ccddata._config_ccd_requires_unit = True` itself and drops the marker entirely — it passes deterministically on all four backends regardless of what ran before it.

Leftover failures after this PR (for #971 bookkeeping):
- `test_combiner_result_dtype` — `combine(dtype=int)` vs. array-api-strict's dtype objects (new gap, undocumented pre-PR).
- `test_combine_overwrite_output` — `combine(output_file=...)` writes a namespace array via `CCDData.write()` (same class as #935).
- `test_combiner_with_scaling`, `test_combiner_with_scaling_uncertainty[median_combine]`, `test_writeable_after_combine[median_combine]` — #929 (`median_combine`'s default `sigma_func` densifies via `astropy.stats`).
- The remaining ~10 failures are the pre-existing, untouched upstream bucket (`test_unit_mismatch_behaves_as_expected`, `test_sigma_func_for_ccddata`, `test_combiner_dtype`, `test_combiner_sigmaclip_*`, `test_combiner_median`, `test_combine_result_uncertainty_and_mask[median_combine-*]`, `test_rebin_ccddata[True-True]`).

### DIFF
--- a/ccdproc/tests/pytest_fixtures.py
+++ b/ccdproc/tests/pytest_fixtures.py
@@ -88,11 +88,16 @@ def numpy_copy(array):
     ``np.asarray`` only works for arrays on the namespace's default
     device. The strict tests run on ``Device("device1")``, on which
     array-api-strict deliberately refuses to export to NumPy, so the array
-    is moved to the default device first.
+    is moved to the default device first. That move is a no-op on every
+    other backend the suite runs. The default device is asked for through
+    the standard ``__array_namespace_info__().default_device()``, which
+    the 2025.12 standard allows to be `None` (JAX does this); in that case
+    there is nothing to move to and ``np.asarray`` is used directly.
     """
     xp = array_api_compat.array_namespace(array)
-    if xp.__name__ == "array_api_strict":
-        array = xp.asarray(array, device=xp.Device("CPU_DEVICE"))
+    default_device = xp.__array_namespace_info__().default_device()
+    if default_device is not None:
+        array = array_api_compat.to_device(array, default_device)
     return np.asarray(array)
 
 
--- a/ccdproc/tests/pytest_fixtures.py
+++ b/ccdproc/tests/pytest_fixtures.py
@@ -2,6 +2,7 @@
 
 from shutil import rmtree
 
+import array_api_compat
 import numpy as np
 import pytest
 from astropy import units as u
@@ -66,6 +67,39 @@ def ccd_data(
     ccd = CCDData(xp.asarray(data, device=xp_device), unit=u.adu)
     ccd.header = fake_meta
     return ccd
+
+
+def numpy_copy(array):
+    """
+    Return a NumPy copy of ``array``, moving it off a non-default device first.
+
+    ``np.asarray`` raises on an array-api-strict array that lives on one of
+    the library's fake non-default devices (the suite runs strict tests on
+    ``Device("device1")`` for exactly that reason), so move the array to the
+    default device before converting.
+    """
+    xp = array_api_compat.array_namespace(array)
+    if xp.__name__ == "array_api_strict":
+        array = xp.asarray(array, device=xp.Device("CPU_DEVICE"))
+    return np.asarray(array)
+
+
+def numpy_ccddata(ccd):
+    """
+    Return a copy of ``ccd`` with ``data``, ``mask`` and ``uncertainty.array``
+    (when present) converted to NumPy via `numpy_copy`.
+
+    Unit and meta are copied over unchanged. Useful for handing a namespace
+    ``CCDData`` to code that requires NumPy, such as ``CCDData.write``.
+    """
+    new_ccd = CCDData(numpy_copy(ccd.data), unit=ccd.unit, meta=ccd.meta)
+    if ccd.mask is not None:
+        new_ccd.mask = numpy_copy(ccd.mask)
+    if ccd.uncertainty is not None:
+        new_ccd.uncertainty = ccd.uncertainty.__class__(
+            numpy_copy(ccd.uncertainty.array)
+        )
+    return new_ccd
 
 
 @pytest.fixture
--- a/ccdproc/tests/pytest_fixtures.py
+++ b/ccdproc/tests/pytest_fixtures.py
@@ -71,12 +71,24 @@ def ccd_data(
 
 def numpy_copy(array):
     """
-    Return a NumPy copy of ``array``, moving it off a non-default device first.
+    Return a NumPy copy of an array from any array namespace.
 
-    ``np.asarray`` raises on an array-api-strict array that lives on one of
-    the library's fake non-default devices (the suite runs strict tests on
-    ``Device("device1")`` for exactly that reason), so move the array to the
-    default device before converting.
+    Parameters
+    ----------
+    array : array-like
+        An array from any array-API namespace, on any device.
+
+    Returns
+    -------
+    `numpy.ndarray`
+        A NumPy array with the same shape, dtype and values as ``array``.
+
+    Notes
+    -----
+    ``np.asarray`` only works for arrays on the namespace's default
+    device. The strict tests run on ``Device("device1")``, on which
+    array-api-strict deliberately refuses to export to NumPy, so the array
+    is moved to the default device first.
     """
     xp = array_api_compat.array_namespace(array)
     if xp.__name__ == "array_api_strict":
@@ -86,11 +98,25 @@ def numpy_copy(array):
 
 def numpy_ccddata(ccd):
     """
-    Return a copy of ``ccd`` with ``data``, ``mask`` and ``uncertainty.array``
-    (when present) converted to NumPy via `numpy_copy`.
+    Return a copy of a ``CCDData`` whose arrays are all NumPy arrays.
 
-    Unit and meta are copied over unchanged. Useful for handing a namespace
-    ``CCDData`` to code that requires NumPy, such as ``CCDData.write``.
+    Parameters
+    ----------
+    ccd : `~astropy.nddata.CCDData`
+        Image whose ``data``, and ``mask`` and ``uncertainty`` if present,
+        may be in any array namespace.
+
+    Returns
+    -------
+    `~astropy.nddata.CCDData`
+        A new ``CCDData`` with ``data``, ``mask`` and ``uncertainty.array``
+        converted with `numpy_copy`; ``unit`` and ``meta`` are carried over
+        unchanged and the uncertainty keeps its class.
+
+    Notes
+    -----
+    Use this to hand a namespace ``CCDData`` to code that requires NumPy,
+    such as ``CCDData.write``.
     """
     new_ccd = CCDData(numpy_copy(ccd.data), unit=ccd.unit, meta=ccd.meta)
     if ccd.mask is not None:
--- a/ccdproc/tests/test_combiner.py
+++ b/ccdproc/tests/test_combiner.py
@@ -30,6 +30,7 @@ from ccdproc.combiner import (
 # Set up the array library to be used in tests
 from ccdproc.conftest import testing_array_device as xp_device
 from ccdproc.conftest import testing_array_library as xp
+from ccdproc.core import _native_numpy
 from ccdproc.image_collection import ImageFileCollection
 from ccdproc.tests.pytest_fixtures import ccd_data as ccd_data_func
 from ccdproc.tests.pytest_fixtures import numpy_ccddata, numpy_copy
@@ -874,9 +875,8 @@ def test_combine_average_ccddata():
     fitsfile = get_pkg_data_filename("data/a8280271.fits")
     ccd = CCDData.read(fitsfile, unit=u.adu)
     # ``combine`` ignores ``array_package`` for CCDData input, so convert
-    # the data to the namespace ourselves (``.astype(float)`` makes the
-    # big-endian FITS data native; strict/jax reject ``>f8``).
-    ccd.data = xp.asarray(ccd.data.astype(float))
+    # the data to the namespace ourselves.
+    ccd.data = xp.asarray(_native_numpy(ccd.data))
     ccd_list = [ccd] * 3
     c = Combiner(ccd_list)
     ccd_by_combiner = c.average_combine()
--- a/ccdproc/tests/test_combiner.py
+++ b/ccdproc/tests/test_combiner.py
@@ -7,7 +7,6 @@ import array_api_compat
 import array_api_extra as xpx
 import astropy.units as u
 import numpy as np
-import numpy.ma as np_ma
 import pytest
 from astropy.nddata import CCDData
 from astropy.stats import median_absolute_deviation as mad
@@ -33,6 +32,7 @@ from ccdproc.conftest import testing_array_device as xp_device
 from ccdproc.conftest import testing_array_library as xp
 from ccdproc.image_collection import ImageFileCollection
 from ccdproc.tests.pytest_fixtures import ccd_data as ccd_data_func
+from ccdproc.tests.pytest_fixtures import numpy_ccddata, numpy_copy
 
 # Several tests have many more NaNs in them than real data. numpy generates
 # lots of warnings in those cases and it makes more sense to suppress them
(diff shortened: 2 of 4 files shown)


ccdproc/tests/pytest_fixtures.py
```diff
@@ -68,6 +69,39 @@ def ccd_data(
     return ccd
 
 
+def numpy_copy(array):
+    """
+    Return a NumPy copy of ``array``, moving it off a non-default device first.
```
mwcraig: Make this a proper numpy-style docstring

ccdproc/tests/pytest_fixtures.py
```diff
@@ -68,6 +69,39 @@ def ccd_data(
...
+    default device before converting.
+    """
+    xp = array_api_compat.array_namespace(array)
+    if xp.__name__ == "array_api_strict":
+        array = xp.asarray(array, device=xp.Device("CPU_DEVICE"))
+    return np.asarray(array)
+
+
+def numpy_ccddata(ccd):
+    """
```
mwcraig, line 105: Same docstring comment here

ccdproc/tests/pytest_fixtures.py
```diff
@@ -68,6 +69,39 @@ def ccd_data(
...
+    Return a NumPy copy of ``array``, moving it off a non-default device first.
+
+    ``np.asarray`` raises on an array-api-strict array that lives on one of
+    the library's fake non-default devices (the suite runs strict tests on
+    ``Device("device1")`` for exactly that reason), so move the array to the
+    default device before converting.
+    """
+    xp = array_api_compat.array_namespace(array)
+    if xp.__name__ == "array_api_strict":
```
mwcraig: Should the tests do this for all devices? For example, if I run this on cupy will I need to add that name to this conditional? If so it seems better to always do it.

ccdproc/tests/test_combiner.py
```diff
@@ -857,6 +873,10 @@ def test_combine_image_file_collection_input(tmp_path):
 def test_combine_average_ccddata():
     fitsfile = get_pkg_data_filename("data/a8280271.fits")
     ccd = CCDData.read(fitsfile, unit=u.adu)
+    # ``combine`` ignores ``array_package`` for CCDData input, so convert
+    # the data to the namespace ourselves (``.astype(float)`` makes the
+    # big-endian FITS data native; strict/jax reject ``>f8``).
```
mwcraig: Don't we already have a way of converting to native somewhere else in the test suite?

ccdproc/tests/pytest_fixtures.py
mwcraig: Agreed. Proposed numpy-style version (assuming the device-generic body from the thread on line 82, so the explanation moves out of the summary and into `Notes`):

```python
def numpy_copy(array):
    """
    Return a NumPy copy of an array from any array namespace.

    Parameters
    ----------
    array : array-like
        An array from any array-API namespace, on any device.

    Returns
    -------
    `numpy.ndarray`
        A NumPy array with the same shape, dtype and values as ``array``.

    Notes
    -----
    ``np.asarray`` only works for arrays on the namespace's default
    device. The strict tests run on ``Device("device1")``, on which
    array-api-strict deliberately refuses to export to NumPy, so the array
    is moved to the default device first. That move is a no-op on every
    other backend the suite runs.
    """
```

Can push together with the other docstring and the device change.

ccdproc/tests/pytest_fixtures.py
mwcraig: Agreed — same treatment:

```python
def numpy_ccddata(ccd):
    """
    Return a copy of a ``CCDData`` whose arrays are all NumPy arrays.

    Parameters
    ----------
    ccd : `~astropy.nddata.CCDData`
        Image whose ``data``, and ``mask`` and ``uncertainty`` if present,
        may be in any array namespace.

    Returns
    -------
    `~astropy.nddata.CCDData`
        A new ``CCDData`` with ``data``, ``mask`` and ``uncertainty.array``
        converted with `numpy_copy`; ``unit`` and ``meta`` are carried over
        unchanged and the uncertainty keeps its class.

    Notes
    -----
    Use this to hand a namespace ``CCDData`` to code that requires NumPy,
    such as ``CCDData.write``.
    """
```

ccdproc/tests/pytest_fixtures.py
mwcraig: You are right, and it does not need a name list at all. I checked what works on every backend the suite runs (strict on `device1`, jax, numpy, dask): moving the array to the namespace's *own* default device, asked for through the standard `__array_namespace_info__().default_device()`, then `np.asarray`:

```python
def numpy_copy(array):
    xp = array_api_compat.array_namespace(array)
    default_device = xp.__array_namespace_info__().default_device()
    array = array_api_compat.to_device(array, default_device)
    return np.asarray(array)
```

On strict `default_device()` is `Device("CPU_DEVICE")`, so the `device1` array is moved and `np.asarray` works; on jax it is `None` (`to_device` is then a no-op), on numpy/dask it is `"cpu"`. No backend is named, so a new one is picked up automatically as long as it follows the 2023.12 standard.

One honest caveat for a real accelerator: on CuPy the *default* device is the GPU, so this moves nothing and `np.asarray` still fails (CuPy needs `.get()`/`cp.asnumpy`). array-api-compat special-cases `to_device(x, "cpu")` for exactly that (it returns a NumPy array for CuPy and a CPU tensor for torch), but the `"cpu"` string is rejected by strict (`Unsupported device 'cpu'`) and by jax (`device_put` wants a `Device`), so one unconditional call cannot cover both worlds. If you want CuPy covered too the shape would be "try `to_device(array, "cpu")`, fall back to the default-device move", which is what the CPU-only policy in #935 would want anyway. I would go with the default-device version above now and add the `"cpu"` branch when CuPy is actually run, since it cannot be exercised here.

ccdproc/tests/test_combiner.py
mwcraig: Yes — in the library rather than the test suite: `_native_numpy` in `ccdproc/core.py` (line 101, added in #995), which is what `combine()` and `ImageFileCollection.ccds()/data()` use for exactly this FITS `>f8` case. It does `arr.astype(arr.dtype.type, copy=False)`, so for this file it produces the same native `float64` as `.astype(float)` does, without hard-coding `float`.

The test suite already imports private helpers from `core` (`_percentile_fallback` in `test_ccdmask.py`, `_short_names` in `test_ccdproc_logging.py`, `_cosmicray_median_array` in `test_cosmicray.py`), so this would be consistent:

```python
from ccdproc.core import _native_numpy
...
    ccd.data = xp.asarray(_native_numpy(ccd.data))
```

with the comment trimmed to "`combine` ignores `array_package` for `CCDData` input, so convert the data to the namespace ourselves". This is the only place in the PR that does the conversion by hand — the other four `combine()` tests go through `array_package=xp`/`Combiner(xp=xp)`, which call `_native_numpy` internally. Happy to switch it.

ccdproc/tests/pytest_fixtures.py
mwcraig: Has any of this changed in the latest version of the API, 2025.12?

ccdproc/tests/pytest_fixtures.py
mwcraig: Done in d4c09bc. One wording difference from the proposal above: the `Notes` describe what the body does *today* (moves a strict array to the default device by name), since the device-generic body is still open in the thread on line 82. If that lands, the last sentence of `Notes` grows the "no-op on every other backend" clause.

ccdproc/tests/pytest_fixtures.py
mwcraig: Done in d4c09bc, exactly as above.

ccdproc/tests/test_combiner.py
mwcraig: Done in d4c09bc — `test_combine_average_ccddata` now uses `ccdproc.core._native_numpy` and the comment is trimmed as proposed.

ccdproc/tests/pytest_fixtures.py
mwcraig: Short answer: one small thing, and it argues *for* the device-generic version with a one-line guard; nothing else moved.

I read the [2025.12 changelog](https://data-apis.org/array-api/latest/changelog.html) and re-ran the experiments above under 2025.12 — the installed `array-api-strict` 2.6.1 already defaults to `api_version="2025.12"` (since 2.5), so every result in my earlier comment was in fact measured against 2025.12.

**What 2025.12 changed that touches this**

1. `__array_namespace_info__().default_device()` is now explicitly allowed to return `None` "when the default device is not predictable". JAX does exactly that (`default_device() -> None`, `devices() -> (None, CpuDevice(id=0))`). `array_api_compat.to_device(x, None)` happens to work on jax today, but `None` is not a valid `to_device` target in the spec, so the generic body should skip the move in that case:

   ```python
   xp = array_api_compat.array_namespace(array)
   default_device = xp.__array_namespace_info__().default_device()
   if default_device is not None:
       array = array_api_compat.to_device(array, default_device)
   return np.asarray(array)
   ```
2. `devices()` now returns a tuple instead of a list, and there is new normative text on which device *outputs* land on. Neither affects a conversion helper.

**What did not change**

- There is still no standard `"cpu"` device identifier; `to_device(x, "cpu")` remains an array-api-compat extension, and strict (`Unsupported device 'cpu'`) and jax (`device_put` wants a `Device`) still reject it. So the CuPy caveat stands as written.
- The DLPack story is unchanged from 2023.12 (2024.12 only clarified `copy=True`; 2025.12 touched nothing). The spec's blessed way to get a host copy from any device is the *consumer's* `from_dlpack`, i.e. `np.from_dlpack(x, device="cpu", copy=True)`. That works here on strict-`device1`, jax and numpy, and it is what would cover CuPy — but `dask.array` has no `__dlpack__` at all, so it cannot be the only path for this suite.
- array-api-strict itself: since 2.4 it exports to NumPy through the buffer protocol rather than `__array__`, and 2.6.1 still raises `RuntimeError: Can't convert array on the 'device1' device to a NumPy array` for a non-default device — that refusal is deliberate and is the whole reason the helper exists. (2.6.1 also added `Device("no_x64")` and `Device("no_float64")`; `default_device()` is still `CPU_DEVICE`, so the generic move handles them too.)

**Recommendation** — unchanged: default-device move + `np.asarray`, now with the `None` guard above, and no backend names. When a GPU backend is actually run, add `np.from_dlpack(array, device="cpu", copy=True)` as the first attempt with the default-device move as the fallback for libraries without DLPack (dask). I have not changed the body in this PR pending your call.

mwcraig: Done in 2d34b6e — `numpy_copy` now moves the array to `__array_namespace_info__().default_device()` when that is not `None`, then `np.asarray`; no backend is named. The `Notes` section explains the `None` case. Full suite unchanged on numpy (549 passed) and strict (15 failed / 5 xpassed, same set as before); `test_combiner.py` passes on jax and dask.

Issue #983, opened 2026-08-23
Consider marray as a uniform masked-array representation across array backends


mwcraig:
Follow-up to a question raised on #979 (https://github.com/astropy/ccdproc/pull/979/changes#r3837117088): should ccdproc use [`marray`](https://github.com/mdhaber/marray), the array-API-compatible masked-array package, as a uniform "data + mask" representation across backends?

### Current state

Masked data is not part of the array API standard, so ccdproc carries masks around by hand, in several different shapes:

- `numpy.ma.MaskedArray` inputs, detected with `hasattr(x, "mask")` / `isinstance` checks and split into `.data` / `.mask` (`cosmicray_median`, `core.py`; `block_reduce`/`block_average` comment at `core.py:~2283`).
- `CCDData.mask`, which astropy keeps as a numpy array even when `.data` lives in another namespace, so every consumer coerces it with `xp.asarray(ccd.mask, dtype=xp.bool, device=...)` (`Combiner.__init__`, `combine()` in three places, `flat_correct`, `transform_image`, `wcs_project`, `rebin`, `cosmicray_lacosmic`, `cosmicray_median`).
- Internal `(data, mask)` pairs in `Combiner` (`_data_arr`, `_data_arr_mask`) with the mask applied via `xp.where` / NaN substitution before reductions.

This works but the attribute juggling is repeated, easy to get subtly wrong (#932 was exactly a dropped mask; #965 was the mask stacking), and mask device/namespace placement has to be re-established at every site.

### What `marray` would and would not give

- **Would**: one object carrying data + mask in any array-API namespace, with mask-aware reductions (`mean`, `sum`, `std`, ...) and elementwise ops, so `Combiner` could drop much of its NaN/`xp.where` bookkeeping, and functions could accept/return one masked type instead of branching on `MaskedArray` vs `CCDData` vs bare array.
- **Would not**: help with anything that leaves the array API. `scipy.ndimage.median_filter` (`cosmicray_median`), `astroscrappy`, `reproject` and `scipy.ndimage` in `transform_image` all see only the underlying data, so the fill-then-filter pattern around them stays regardless (see the #979 discussion).

### Open questions

1. Is a new required dependency acceptable, or would it be optional with a fallback (which would mean keeping the manual paths anyway)?
2. Maturity / maintenance status of `marray`, and whether its masked reductions match the semantics ccdproc currently documents (e.g. `Combiner` weighted averages exclude masked *and* clipped pixels, #952).
[...]

Issue #936, opened 2026-07-07
Array API: units/Quantity handling with non-numpy arrays


mwcraig:
### Problem

`astropy.units.Quantity` is numpy-backed, so any path that attaches units to image data pulls the data out of the user's array library.

Evidence:

- The workaround in `flat_correct`, which strips and re-applies units manually to avoid Quantity:

  https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/core.py#L991

- #911 failures of the form `TypeError: unsupported operand type(s) for <<: 'ndarray' and 'IrreducibleUnit'` in scaling-related combiner tests — a CuPy array being shifted into a Quantity.

- The related TODO about `CCDData` arithmetic being numpy-based:

  https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/core.py#L1016

### Task

Define one strategy for units with non-numpy data — the current wrapper approach (`_CCDDataWrapperForArrayAPI` keeps units *beside* the data rather than *wrapping* it) seems right, but it needs to cover the combiner scaling paths too. Audit every place a `Quantity` is constructed from image data and route it through the same mechanism. Track astropy's own array-API plans for units in #940.

---
Found during a review of the array API implementation from #885; follow-up to #909 / #911.

mwcraig:
Audited this: there is no ccdproc-side bug to fix. Every `Quantity` that ccdproc constructs on the gain/readnoise path wraps a Python scalar, not an array, so none of them can drag the data out of its array namespace or off its device:

- `core.py` ~L753 — `create_deviation`'s default gain
- `core.py` ~L2171 — `cosmicray_lacosmic`'s `gain * u.electron / u.adu`
- `core.py` ~L2176 — `cosmicray_lacosmic`'s `readnoise * u.electron`
- `core.py` ~L2246 — `cosmicray_lacosmic`'s `gain.value * u.one` special case

[...]

PR #999, opened 2026-08-25
Write combine(output_file=) from a NumPy copy; map dtype to the array namespace

mwcraig:
Closes out the last three ccdproc-side items in section 1 of #971.

**`combine(output_file=...)`** handed the namespace result straight to `CCDData.write` (`combiner.py:1273`), which `astropy.io.fits` cannot take (`AttributeError: 'Array' object has no attribute 'astype'` on array-api-strict). The writer now gets a `CCDData` built from NumPy copies of `data`, `mask` and `uncertainty.array` (same uncertainty class, `unit`/`meta`/`wcs` carried over) and the returned result stays in its namespace; `test_combine_overwrite_output` now asserts both.

The copy is a new `ccdproc.core._to_numpy(arr)`: the deliberate host-side conversion for NumPy-only consumers. It moves the array to the namespace's default device first (asked for through the standard `__array_namespace_info__().default_device()`, skipped when that is `None` as on JAX), because array-api-strict refuses to export from its non-default devices. This is the same body as the `numpy_copy` test helper from #998, which now just delegates to it so the device logic lives in one place. Two consequences worth knowing:
- it is the first *library-side* host-copy helper, so it is the natural building block for whatever #935 decides for the CPU-only operations;
- in the escape log, test-suite calls through `numpy_copy` are now attributed to `core.py _to_numpy` (the innermost non-test ccdproc frame), so that site's count is inflated by the test helpers. It is one baseline entry either way.

**`Combiner(dtype=)` / `combine(dtype=)`** passed the user's dtype straight to `xp.asarray`/`xp.astype` (`combiner.py:191`, `:1104`); a builtin `int` or a string such as `"float32"` is a valid NumPy dtype but array-api-strict rejects it. `core._namespace_dtype(dtype, xp)` resolves the name with `numpy.dtype(...)` and looks it up on the namespace; anything NumPy cannot interpret (the namespace's own dtype objects, e.g. `array_api_strict.float32`) passes through untouched. Verified for builtins, strings, NumPy scalar types/dtypes and namespace dtypes on numpy, jax, dask and strict. New `test_combiner_dtype_mapped_to_namespace` covers `int`, `float`, `"float32"`, `np.float32`; the `dtype` docstrings are updated. `test_combiner_result_dtype` then failed in its own body — comparing the integer result with a float reference, which the standard does not promote — and now compares with a Python `int`.

**`subtract_overscan`** used `len(oscan)` for the model fit (`core.py:698`); now `oscan.shape[0]`. The test stays xfailed on #933 (`astropy.modeling`), as expected.

**Escape baseline** (`ccdproc/tests/array_escape_baseline.txt`): regenerated on dask with `CCDPROC_WRITE_ESCAPE_BASELINE=1`. `_to_numpy` is the one new site, tagged `BOUNDARY`. The rewrite also dropped the `average_combine`/`median_combine`/`sum_combine` entries, which have not fired since #992 replaced the `mask=` constructor calls with `_mask`, and updated the header comment (the writer's current text). The dask enforce run passes.

**Results**
- strict: **13 failed / 502 passed / 36 skipped / 45 xfailed / 0 xpassed** (was 15). `test_combine_overwrite_output` and `test_combiner_result_dtype` clear. All 13 that remain are upstream-blocked: 11 × #929, #936, #983 — section 1 of #971 is done.
- numpy: 561 passed / 35 skipped. jax and dask: `test_combiner.py` + `test_ccdproc.py` 177 passed each. dask + escape ratchet: 554 passed, no new escapes.

### DIFF
--- a/ccdproc/combiner.py
+++ b/ccdproc/combiner.py
@@ -20,7 +20,7 @@ from astropy.stats import sigma_clip
 from astropy.utils import deprecated_renamed_argument
 
 from ._nanfuncs import nanmean, nanmedian, nanstd, nansum
-from .core import _native_numpy, sigma_func
+from .core import _namespace_dtype, _native_numpy, _to_numpy, sigma_func
 
 __all__ = ["Combiner", "combine"]
 
@@ -98,9 +98,12 @@ class Combiner:
     ccd_iter : list or generator
         A list or generator of CCDData objects that will be combined together.
 
-    dtype : str or `numpy.dtype` or None, optional
-        Allows user to set dtype. See `numpy.array` ``dtype`` parameter
-        description. If ``None`` it uses ``np.float64``.
+    dtype : dtype-like or None, optional
+        The dtype for the stacked data and the results: a dtype object of
+        the array namespace, or anything NumPy accepts as a dtype (e.g.
+        ``int``, ``"float32"``, `numpy.float32`), which is mapped to the
+        namespace's dtype of the same name. If ``None`` the namespace's
+        ``float64`` is used.
         Default is ``None``.
 
     xp : array namespace, optional
@@ -178,6 +181,8 @@ class Combiner:
         self._xp = xp
         if dtype is None:
             dtype = xp.float64
+        else:
+            dtype = _namespace_dtype(dtype, xp)
 
         self.unit = default_unit
         self.weights = None
@@ -990,9 +995,10 @@ def combine(
         - ``sigma_clip_func`` : function, optional
         - ``sigma_clip_dev_func`` : function, optional
 
-    dtype : str or `numpy.dtype` or None, optional
-        The intermediate and resulting ``dtype`` for the combined CCDs. See
-        `ccdproc.Combiner`. If ``None`` this is set to ``float64``.
+    dtype : dtype-like or None, optional
+        The intermediate and resulting ``dtype`` for the combined CCDs; see
+        `ccdproc.Combiner` for the accepted forms. If ``None`` this is set to
+        the namespace's ``float64``.
         Default is ``None``.
 
     combine_uncertainty_function : callable, None, optional
@@ -1089,6 +1095,8 @@ def combine(
     xp = array_api_compat.array_namespace(ccd.data)
     if dtype is None:
         dtype = xp.float64
+    else:
+        dtype = _namespace_dtype(dtype, xp)
 
     if sigma_clip_func is None:
         sigma_clip_func = xp.mean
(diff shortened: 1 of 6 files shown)


ccdproc/combiner.py
```diff
@@ -98,9 +98,12 @@ class Combiner:
     ccd_iter : list or generator
         A list or generator of CCDData objects that will be combined together.
 
-    dtype : str or `numpy.dtype` or None, optional
-        Allows user to set dtype. See `numpy.array` ``dtype`` parameter
-        description. If ``None`` it uses ``np.float64``.
+    dtype : dtype-like or None, optional
+        The dtype for the stacked data and the results: a dtype object of
```
mwcraig: Shorten this

ccdproc/core.py
```diff
@@ -127,6 +128,76 @@ def _native_numpy(arr):
     return arr.astype(arr.dtype.type, copy=False)
 
 
+def _to_numpy(arr):
```
mwcraig, line 131: Wasn't something like this added earlier?

ccdproc/combiner.py
mwcraig: Yes. Proposed, 4 lines instead of 7 and no longer spelling out every accepted form (the `_namespace_dtype` docstring does that):

```
    dtype : dtype-like or None, optional
        The dtype of the stacked data and the results; NumPy-style dtypes
        (e.g. ``int``, ``"float32"``) are mapped to the namespace's dtype
        of the same name. Default is ``None``, i.e. ``float64``.
```

The `combine()` one already just points here, so it can stay as is. Can push when you say so.

ccdproc/core.py
mwcraig: Yes — `numpy_copy` in `ccdproc/tests/pytest_fixtures.py`, added in #998 (`54f321d`, made device-generic in `2d34b6e`). Its body is exactly this function.

This PR does not duplicate it, it *moves* it: the body now lives here as `core._to_numpy` because the library needs it for the `output_file` writer, and `numpy_copy` is reduced to a one-line delegate (`return _to_numpy(array)`) so the device logic exists in one place. That is the `pytest_fixtures.py` hunk in the diff. There is no other library-side host copy to reuse — `_native_numpy` (#995, `core.py:102`) is the different job of converting a NumPy array to native byte order, and a grep of the package for `to_device`/`default_device`/`np.asarray` finds nothing else outside the test suite and the escape logger.

If you would rather not keep the test-suite alias at all, the two test callers (`numpy_copy` itself and `numpy_ccddata`) can import `_to_numpy` from `ccdproc.core` directly, the way `test_combiner.py` already imports `_native_numpy`; that removes `numpy_copy` and its docstring entirely. Happy to do either.

ccdproc/combiner.py
mwcraig: Done in 57fd985, as proposed above.

ccdproc/core.py
mwcraig: Done in 57fd985 — `numpy_copy` is gone; `numpy_ccddata` and `test_combiner.py` import `_to_numpy` from `ccdproc.core` directly. Full suite unchanged on numpy (562 passed) and strict (the same 13, all upstream); jax and the dask escape ratchet pass.

PR #1005, opened 2026-08-30
Make the array-api-strict suite green and promote the strict CI job

mwcraig:
Follow-up to #1000/#1001, finishing the strict-backend items on the array-API migration list.

**Commit 1 — make the strict suite green.** The one remaining strict failure in ccdproc's own code was `flat_correct` checking the flat's mask with the `mask.any()` method, which spec-only namespaces like array-api-strict don't provide; it now uses `xp.any`. The stale `backend_xfail` on `test_flat_correct_masked_flat_with_immutable_array` (its reason described an older failure mode) is removed and the test passes everywhere. The two other strict failures are astropy-upstream and get accurate `backend_xfail` marks in line with the rest of the suite:

- `test_unit_mismatch_behaves_as_expected`: astropy's arithmetic mixin builds the Quantity with `data << unit`, and array-api-strict rejects the astropy unit operand (see #936).
- `test_rebin_ccddata[True-True]`: astropy's `CCDData.mask` setter coerces the mask with `np.asarray`, which fails for an array on the non-default device.

**Commit 2 — promote the strict CI job.** With the suite green (839 passed, 46 xfailed, 0 failed locally), the expected-failures carve-out is retired: `py313-strict` moves into the regular `ci-tests` matrix where a failure fails CI, and the `continue-on-error` step, outcome-artifact plumbing, and the `strict_status.yml` reporting workflow are removed.

Verified locally on the full suite for every backend: numpy (890 passed), dask (879 passed, escape-baseline ratchet enforced), jax (885 passed, ratchet enforced), array-api-strict (839 passed, 46 xfailed).

https://claude.ai/code/session_01LVxnTLrWKStBdPwcmDNhxA


CHANGES.rst
```diff
@@ -20,6 +20,11 @@ New Features
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+- The ``py313-strict`` CI job (the ``array-api-strict`` suite on a
```
mwcraig: Shorten

CHANGES.rst
```diff
@@ -37,6 +42,10 @@ Other Changes and Additions
 Bug Fixes
 ^^^^^^^^^
 
+- ``flat_correct`` now checks whether the flat has any masked pixels with
```
mwcraig: Shorten

ccdproc/tests/test_rebin.py
```diff
@@ -64,6 +64,12 @@ def test_rebin_smaller():
 
 
 # test rebinning with ccddata object
+@pytest.mark.backend_xfail(
+    "array-api-strict",
+    reason="astropy's CCDData.mask setter coerces the mask with "
+    "np.asarray, which fails for an array-api-strict array on the "
+    "non-default device",
+)
 @pytest.mark.parametrize("mask_data, uncertainty", [(False, False), (True, True)])
```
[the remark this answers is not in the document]
mwcraig: Confirmed — `backend_xfail` is applied non-strict (`conftest.py` adds `pytest.mark.xfail(reason=..., strict=False)`), so with the function-level marker a genuine regression in the currently passing `[False-False]` case would be reported as an expected failure rather than failing the job. Will scope the marker to the failing case via the suite's existing per-parameter pattern (as with `_GAIN_APPLY_STRICT_XFAIL` in `test_cosmicray.py`):

```python
@pytest.mark.parametrize(
    "mask_data, uncertainty",
    [
        (False, False),
        pytest.param(
            True,
            True,
            marks=pytest.mark.backend_xfail(
                "array-api-strict",
                reason="astropy's CCDData.mask setter coerces the mask with "
                "np.asarray, which fails for an array-api-strict array on "
                "the non-default device",
            ),
        ),
    ],
)
def test_rebin_ccddata(mask_data, uncertainty):
```

CHANGES.rst
mwcraig: Proposed shorter entry:

```rst
- The ``py313-strict`` CI job is now part of the regular CI matrix and fails
  CI on a test failure; the expected-failures carve-out and the
  ``strict_status.yml`` reporting workflow are retired. [#1005]
```

CHANGES.rst
mwcraig: Proposed shorter entry:

```rst
- ``flat_correct`` now checks the flat's mask with ``xp.any`` instead of the
  ``mask.any()`` method, which spec-only array-API namespaces do not
  provide. [#1005]
```

PR #1006, opened 2026-08-31
Hoist None/tuple axis handling into _nanfuncs._setup

### LINKED ISSUE #1004: Hoist None/tuple axis handling into _nanfuncs._setup
Follow-up agreed in the #1000 review (https://github.com/astropy/ccdproc/pull/1000#discussion_r3890062495), deferred until that PR was in: with `normalize_axis_tuple` adopted there, moving the flatten-for-`None` and permute-last-and-merge steps from `core._mad_fallback` into `_nanfuncs._setup` is a small change that gives `_median_fallback` and the combiner fallbacks tuple-axis (and `axis=None`) support for free, and shrinks `_mad_fallback` to a few lines. It would also let `Combiner.sigma_clipping`'s array-API path lift its single-integer-axis restriction (#1001 documents that restriction).

mwcraig:
Follow-up agreed in the #1000 review; closes #1004.

Moves the flatten-for-`None` and permute-and-merge steps from `core._mad_fallback` into the shared `_nanfuncs._setup`, which now accepts a single integer, `None`, or a tuple/list of axes and returns a `restore` callable mapping a full-shape array back to the caller's layout (identity for a single integer axis).

What this buys:

- every reduction fallback in `_nanfuncs` (`nansum`, `nanmean`, `nanstd`, `nanmedian`, `median`, `nanmad`) gains tuple-axis and `axis=None` support for free;
- `_mad_fallback` shrinks to the median tier plus a delegation to `nanmad`;
- `Combiner.sigma_clipping`'s array-API path lifts its single-integer-axis restriction and now accepts the same axis forms `astropy.stats.sigma_clip` does, with the clip mask handed back in the data's own shape via `restore`.

Behavior note: a bare-bool or non-integer `axis` now raises `TypeError` instead of `NotImplementedError`, matching the `TypeError` a bool entry in a tuple already raised.

Tests: the differential grids in `test_nanfuncs.py` / `test_combiner.py` gained `None`, tuple (including negative entries), and single-entry-tuple cases checked against numpy/astropy, a list-equals-tuple equivalence test, expanded bad-axis cases, and Combiner-level axis-form tests. Full suites pass locally: numpy 943, dask 676/676, jax 676/676, array-api-strict 664 passed + 12 pre-existing xfails on the three affected test files.

https://claude.ai/code/session_01LVxnTLrWKStBdPwcmDNhxA

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,10 +17,12 @@ New Features
   ``'median'``/``'mean'``/``'std'``/``'mad_std'`` use the namespace's
   NaN-aware reductions or ccdproc's fallbacks. [#1001]
 - ``Combiner.sigma_clipping`` outside NumPy now accepts ``axis=None`` and a
-  tuple of axes, as ``astropy.stats.sigma_clip`` does on the NumPy path, and
-  the reduction fallbacks in ``ccdproc._nanfuncs`` gained the same axis
-  forms; the None/tuple axis handling formerly in ``_mad_fallback`` moved
-  into the shared ``_nanfuncs._setup``. [#1006]
+  tuple or list of axes, as ``astropy.stats.sigma_clip`` does on the NumPy
+  path, and the reduction fallbacks in ``ccdproc._nanfuncs`` gained the same
+  axis forms; the None/tuple axis handling formerly in ``_mad_fallback``
+  moved into the shared ``_nanfuncs._setup``, and a bool or otherwise
+  non-integer ``axis`` now raises ``TypeError`` rather than
+  ``NotImplementedError``. [#1006]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,11 @@ New Features
   classified differently from astropy (NumPy data still use astropy);
   ``'median'``/``'mean'``/``'std'``/``'mad_std'`` use the namespace's
   NaN-aware reductions or ccdproc's fallbacks. [#1001]
+- ``Combiner.sigma_clipping`` outside NumPy now accepts ``axis=None`` and a
+  tuple of axes, as ``astropy.stats.sigma_clip`` does on the NumPy path, and
+  the reduction fallbacks in ``ccdproc._nanfuncs`` gained the same axis
+  forms; the None/tuple axis handling formerly in ``_mad_fallback`` moved
+  into the shared ``_nanfuncs._setup``. [#1006]
 
 Other Changes and Additions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/ccdproc/_nanfuncs.py
+++ b/ccdproc/_nanfuncs.py
@@ -24,6 +24,10 @@ from functools import partial
 
 import array_api_compat
 
+# Host-side axis normalisation for tuple axes: operates on python ints
+# only, never on array data, so it does not tie the fallbacks to numpy.
+from numpy.lib.array_utils import normalize_axis_tuple
+
 __all__ = ["median", "nanmad", "nanmean", "nanmedian", "nanstd", "nansum"]
 
 
@@ -58,16 +62,24 @@ def _promote_to_real(x, xp, device):
 
 def _setup(x, axis, xp):
     """
-    Validate ``axis``, resolve the namespace and device, promote to float.
+    Normalise ``axis``, resolve the namespace and device, promote to float.
+
+    ``axis`` may be a single integer, ``None`` or a tuple/list of integers.
+    ``None`` flattens ``x`` so the caller reduces over everything; a tuple
+    or list moves the listed axes to the end and merges them into one, so
+    the caller's single-axis reduction reduces over all of them at once.
+    Either way the caller only ever sees a single non-negative integer
+    axis.
 
     Parameters
     ----------
     x : array
         Input array.
-    axis : int
-        Axis along which the caller will reduce. Booleans, ``None`` and
-        tuples of axes are rejected; anything else goes through
-        `operator.index`, so numpy integer scalars are accepted.
+    axis : int, tuple of int, list of int or None
+        Axis or axes along which the caller will reduce. Booleans are
+        rejected -- bool subclasses int, so ``axis=True`` would silently
+        mean axis 1 -- while numpy integer scalars are accepted. Negative
+        values count from the last axis.
     xp : array namespace or None
         Namespace to use. ``None`` resolves it from ``x``.
 
@@ -75,47 +87,82 @@ def _setup(x, axis, xp):
     -------
     x : array
         The input, promoted if necessary to the namespace's default real
-        floating dtype.
+        floating dtype, flattened when ``axis`` is ``None``, and with the
+        listed axes moved to the end and merged into one when ``axis`` is
+        a tuple or list.
     axis : int
-        The axis, normalised to a non-negative integer.
+        The single axis of the returned ``x`` to reduce, normalised to a
+        non-negative integer.
     xp : array namespace
         The resolved namespace.
     device : device
         The device ``x`` lives on.
+    restore : callable
+        Maps an array shaped like the returned ``x`` back to the layout of
+        the input ``x``; the identity for a single integer ``axis``.
+        Reductions remove the reduced axis and never need it;
+        ``ccdproc.combiner._sigma_clip_mask`` keeps the full shape and
+        uses it to hand its mask back in the caller's layout.
 
     Raises
     ------
-    NotImplementedError
-        If ``axis`` is not a single integer.
+    TypeError
+        If ``axis``, or an entry of a tuple/list ``axis``, is a bool or
+        not an integer.
     ValueError
-        If ``axis`` is out of bounds for ``x``.
+        If ``axis``, or an entry of a tuple/list ``axis``, is out of
+        bounds for ``x``, or a tuple/list names an axis more than once
+        (including via a negative alias).
     """
+    if xp is None:
+        xp = array_api_compat.array_namespace(x)
+    device = array_api_compat.device(x)
+    x = _promote_to_real(x, xp, device)
+    ndim = x.ndim
+
+    if axis is None:
+        shape = x.shape
+
+        def restore(a):
+            return xp.reshape(a, shape)
+
+        return xp.reshape(x, (-1,)), 0, xp, device, restore
+
+    if isinstance(axis, tuple | list):
+        # normalize_axis_tuple would silently treat True as 1.
+        if any(isinstance(ax, bool) for ax in axis):
+            raise TypeError("axis entries must be integers, not bool")
+        axes = normalize_axis_tuple(axis, ndim)
+        # Move the reduced axes to the end and merge them into one, so that
+        # a single-axis reduction reduces over all of them at once.
+        kept = tuple(ax for ax in range(ndim) if ax not in axes)
+        order = kept + axes
+        permuted_shape = tuple(x.shape[ax] for ax in order)
+        x = xp.reshape(xp.permute_dims(x, order), permuted_shape[: len(kept)] + (-1,))
+        inverse = tuple(order.index(ax) for ax in range(ndim))
+
+        def restore(a):
+            return xp.permute_dims(xp.reshape(a, permuted_shape), inverse)
+
+        return x, len(kept), xp, device, restore
+
     # bool subclasses int -- axis=True would silently mean axis 1 -- so it is
     # rejected explicitly, while operator.index accepts the numpy integer
     # scalars that isinstance(axis, int) would refuse.
-    if axis is None or isinstance(axis, bool):
-        raise NotImplementedError(
-            "NaN-aware reduction fallbacks support only a single integer axis."
-        )
+    if isinstance(axis, bool):
+        raise TypeError("axis must be an integer, not bool")
     try:
         axis = operator.index(axis)
     except TypeError:
-        raise NotImplementedError(
-            "NaN-aware reduction fallbacks support only a single integer axis."
+        raise TypeError(
+            f"axis must be an integer, a tuple or list of integers, or None, "
+            f"got {axis!r}"
         ) from None
 
-    if xp is None:
-        xp = array_api_compat.array_namespace(x)
-
-    ndim = x.ndim
     if not -ndim <= axis < ndim:
         raise ValueError(f"axis {axis} is out of bounds for array of dimension {ndim}")
-    axis = axis % ndim
 
-    device = array_api_compat.device(x)
-    x = _promote_to_real(x, xp, device)
-
-    return x, axis, xp, device
+    return x, axis % ndim, xp, device, lambda a: a
 
 
 def _sum_and_count(x, axis, xp, device, *, keepdims):
@@ -197,19 +244,20 @@ def nansum(x, /, *, axis=0, xp=None):
     x : array
         Input array. Integer and boolean inputs are promoted to the
         namespace's default real floating dtype.
-    axis : int, optional
-        Axis along which to sum. Default is 0. ``None`` and tuples of axes
-        are not supported.
+    axis : int, tuple of int, list of int or None, optional
+        Axis or axes along which to sum. Default is 0. ``None`` sums over
+        every axis; a tuple or list sums over all the listed axes at once.
     xp : array namespace, optional
         Namespace to use. Defaults to ``array_api_compat.array_namespace(x)``.
 
     Returns
     -------
     array
-        Sum of ``x`` along ``axis``, with that axis removed. Slices that are
-        entirely NaN sum to zero, matching `numpy.nansum`.
+        Sum of ``x`` along ``axis``, with the reduced axes removed (0-d
+        when ``axis`` is ``None``). Slices that are entirely NaN sum to
+        zero, matching `numpy.nansum`.
     """
-    x, axis, xp, device = _setup(x, axis, xp)
+    x, axis, xp, device, _ = _setup(x, axis, xp)
     total, _ = _sum_and_count(x, axis, xp, device, keepdims=False)
     return total
 
@@ -223,21 +271,22 @@ def nanmean(x, /, *, axis=0, xp=None):
     x : array
         Input array. Integer and boolean inputs are promoted to the
         namespace's default real floating dtype.
-    axis : int, optional
-        Axis along which to average. Default is 0. ``None`` and tuples of
-        axes are not supported.
+    axis : int, tuple of int, list of int or None, optional
+        Axis or axes along which to average. Default is 0. ``None``
+        averages over every axis; a tuple or list over all the listed axes.
     xp : array namespace, optional
         Namespace to use. Defaults to ``array_api_compat.array_namespace(x)``.
 
     Returns
     -------
     array
-        Mean of ``x`` along ``axis``, with that axis removed. Slices that
-        are entirely NaN yield NaN silently, matching ``bottleneck.nanmean``
+        Mean of ``x`` along ``axis``, with the reduced axes removed (0-d
+        when ``axis`` is ``None``). Slices that are entirely NaN yield
+        NaN silently, matching ``bottleneck.nanmean``
         (the numpy-backend default); `numpy.nanmean` warns here, but a fully
         masked pixel is a routine input for the combiner, not an anomaly.
     """
-    x, axis, xp, device = _setup(x, axis, xp)
+    x, axis, xp, device, _ = _setup(x, axis, xp)
     total, count = _sum_and_count(x, axis, xp, device, keepdims=False)
     return _safe_divide(total, count, xp, device)
 
@@ -255,16 +304,18 @@ def nanstd(x, /, *, axis=0, xp=None):
     x : array
         Input array. Integer and boolean inputs are promoted to the
         namespace's default real floating dtype.
-    axis : int, optional
-        Axis along which to compute the deviation. Default is 0. ``None``
-        and tuples of axes are not supported.
+    axis : int, tuple of int, list of int or None, optional
+        Axis or axes along which to compute the deviation. Default is 0.
+        ``None`` reduces over every axis; a tuple or list over all the
+        listed axes.
     xp : array namespace, optional
         Namespace to use. Defaults to ``array_api_compat.array_namespace(x)``.
 
     Returns
     -------
     array
-        Standard deviation of ``x`` along ``axis``, with that axis removed.
+        Standard deviation of ``x`` along ``axis``, with the reduced axes
+        removed (0-d when ``axis`` is ``None``).
         Slices that are entirely NaN yield NaN silently, matching
         ``bottleneck.nanstd`` (the numpy-backend default); `numpy.nanstd`
         warns here, but a fully masked pixel is a routine input for the
@@ -279,7 +330,7 @@ def nanstd(x, /, *, axis=0, xp=None):
     single-pass form suffers when the values are large relative to their
     spread, which is not unusual for CCD counts.
     """
-    x, axis, xp, device = _setup(x, axis, xp)
+    x, axis, xp, device, _ = _setup(x, axis, xp)
 
     isnan = xp.isnan(x)
     zero = xp.asarray(0, dtype=x.dtype, device=device)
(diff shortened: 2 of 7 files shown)


ccdproc/combiner.py
```diff
@@ -699,11 +704,11 @@ def sigma_clipping(
 
         kwd
             ``axis`` (default ``0``) and ``maxiters`` (default ``1``) are
-            honoured for every array namespace. Outside numpy ``axis`` must
-            be a single integer: ``None`` and a tuple of axes, both of
-            which :func:`~astropy.stats.sigma_clip` accepts, raise
-            `NotImplementedError` there. ``masked`` and ``return_bounds``
-            are never accepted, on any array namespace -- this method
+            honoured for every array namespace; like
```
mwcraig: American spelling, please

ccdproc/tests/test_nanfuncs.py
```diff
@@ -118,15 +123,25 @@ def test_nanmad_matches_astropy(axis):
     assert bool(xp.all(xpx.isclose(result, expected, equal_nan=True)))
 
 
+@pytest.mark.parametrize("func", [nansum, nanmean, nanstd, nanmedian, median])
+def test_list_axis_matches_tuple(func):
```
mwcraig, line 142: Add an explanation of the test, and maybe remember that you should always do that

ccdproc/_nanfuncs.py
```diff
@@ -58,64 +62,107 @@ def _promote_to_real(x, xp, device):
 
 def _setup(x, axis, xp):
     """
-    Validate ``axis``, resolve the namespace and device, promote to float.
+    Normalise ``axis``, resolve the namespace and device, promote to float.
+
+    ``axis`` may be a single integer, ``None`` or a tuple/list of integers.
```
mwcraig: This goes in notes

ccdproc/_nanfuncs.py
```diff
@@ -197,19 +244,20 @@ def nansum(x, /, *, axis=0, xp=None):
     x : array
         Input array. Integer and boolean inputs are promoted to the
         namespace's default real floating dtype.
-    axis : int, optional
-        Axis along which to sum. Default is 0. ``None`` and tuples of axes
-        are not supported.
+    axis : int, tuple of int, list of int or None, optional
+        Axis or axes along which to sum. Default is 0. ``None`` sums over
```
mwcraig: Can this be pulled out some how since it is repeated so many times? Or maybe factor out most of the docstring with function-specific substitutions?

ccdproc/_nanfuncs.py
```diff
@@ -58,64 +62,107 @@ def _promote_to_real(x, xp, device):
...
+    if axis is None:
+        shape = x.shape
+
+        def restore(a):
+            return xp.reshape(a, shape)
+
+        return xp.reshape(x, (-1,)), 0, xp, device, restore
+
+    if isinstance(axis, tuple | list):
+        # normalize_axis_tuple would silently treat True as 1.
```
mwcraig: Explain in more detail what is going on in this block -- no comments in the code, just explain in the reply

ccdproc/_nanfuncs.py
```diff
@@ -58,64 +62,107 @@ def _promote_to_real(x, xp, device):
...
+        if any(isinstance(ax, bool) for ax in axis):
```
mwcraig: **Bug (found independently by two reviewers): `np.bool_` slips through this guard.** `np.True_` is not a subclass of Python `bool`, so `isinstance(ax, bool)` is `False` for it and `normalize_axis_tuple` then silently treats it as axis 1 — `nansum(x, axis=(0, np.True_))` reduces axes `(0, 1)`, which is exactly the hazard the comment above says this rejects. numpy itself raises `TypeError` here. Same hole in the scalar guard at line 152 (`operator.index(np.True_)` returns 1, currently with a `DeprecationWarning` that numpy will turn into an error — and the test suite's `filterwarnings = ["error", ...]` means a `test_bad_axis` row for this fails on the warning, not the intended `TypeError`). The scalar hole is pre-existing; the tuple-entry one is new here. Fix in both places: `isinstance(ax, bool | np.bool_)` — numpy is already imported host-side. This also falsifies `_setup`'s docstring claim that booleans are rejected while numpy scalars are accepted.

ccdproc/_nanfuncs.py
```diff
@@ -58,64 +62,107 @@ def _promote_to_real(x, xp, device):
...
+            raise TypeError("axis entries must be integers, not bool")
+        axes = normalize_axis_tuple(axis, ndim)
+        # Move the reduced axes to the end and merge them into one, so that
+        # a single-axis reduction reduces over all of them at once.
+        kept = tuple(ax for ax in range(ndim) if ax not in axes)
+        order = kept + axes
+        permuted_shape = tuple(x.shape[ax] for ax in order)
+        x = xp.reshape(xp.permute_dims(x, order), permuted_shape[: len(kept)] + (-1,))
```
mwcraig: **Bug: the `-1` merge reshape crashes when a *kept* dimension has size 0, where numpy returns an empty result.** `nansum(np.zeros((0, 3, 4)), axis=(1, 2))` → `ValueError: cannot reshape array of size 0 into shape (0,newaxis)` on every backend, while `np.nansum` gives shape `(0,)`. Inherited from the old `_mad_fallback` merge, but the hoist spreads it to all six reductions plus `sigma_clipping`/`mad_std`. One-line fix: reshape with the explicit product instead of `-1`, e.g. `permuted_shape[: len(kept)] + (math.prod(permuted_shape[len(kept):]),)`. Rare input in practice, but it's a genuine numpy divergence in the newly advertised feature.

ccdproc/_nanfuncs.py
```diff
@@ -58,64 +62,107 @@ def _promote_to_real(x, xp, device):
...
+    x = _promote_to_real(x, xp, device)
+    ndim = x.ndim
+
+    if axis is None:
+        shape = x.shape
+
+        def restore(a):
+            return xp.reshape(a, shape)
+
+        return xp.reshape(x, (-1,)), 0, xp, device, restore
```
mwcraig: **Simplification: this whole branch folds into the tuple branch.** Replacing it with
```python
if axis is None:
    axis = tuple(range(ndim))
```
produces identical results: `kept=()` makes the permutation the identity, the merge reshape becomes `reshape(x, (-1,))`, the returned axis is `len(kept) == 0`, and `restore` is `permute_dims(reshape(a, shape), identity)`. Verified with a differential harness (all six public functions + `_sigma_clip_mask`, numpy and array-api-strict, 11 axis forms including 0-d input): every value, shape, mask, and error identical. Only cost is a no-op identity `permute_dims` on this path (a view on numpy). Removes one of `_setup`'s three `restore` definitions and an early return.

ccdproc/_nanfuncs.py
```diff
@@ -58,64 +62,107 @@ def _promote_to_real(x, xp, device):
...
-    ndim = x.ndim
     if not -ndim <= axis < ndim:
         raise ValueError(f"axis {axis} is out of bounds for array of dimension {ndim}")
-    axis = axis % ndim
-
-    device = array_api_compat.device(x)
-    x = _promote_to_real(x, xp, device)
 
-    return x, axis, xp, device
+    return x, axis % ndim, xp, device, lambda a: a
```
mwcraig: **Simplification: this hand-rolled bounds check is `normalize_axis_tuple(axis, ndim)[0]`.** The `AxisError` it raises carries a character-for-character identical message and subclasses `ValueError`, so `test_bad_axis` still passes unchanged. It also removes an existing asymmetry: the tuple branch already raises `AxisError` for out-of-bounds entries while this branch raises plain `ValueError`. (Keep the bool guard and the `operator.index` try/except above — their message is better.)
```python
return x, normalize_axis_tuple(axis, ndim)[0], xp, device, lambda a: a
```
Together with the `axis is None` fold, `_setup`'s body goes from ~43 to ~34 lines with proven-identical behavior.

ccdproc/tests/test_nanfuncs.py
```diff
@@ -38,6 +38,11 @@
     (_some_nan, 1),  # a non-zero axis
     (_some_nan, -1),  # a negative axis
     (_some_nan, np.int64(1)),  # a numpy integer axis
+    (_some_nan, None),  # reduce over everything
+    (_some_nan, (0, 1)),  # a tuple covering every axis
+    (_rng.normal(size=(5, 4, 3)), (0, 2)),  # a tuple of axes
+    (_rng.normal(size=(5, 4, 3)), (-1, 0)),  # a negative entry in a tuple
+    (_rng.normal(size=(5, 4, 3)), (1,)),  # a single-entry tuple
```
mwcraig, line 45: **Coverage gaps in the new grid, both verified to work but unpinned:** (1) `axis=()` matches numpy's no-op reduction (elementwise, NaN→0 for nansum) — but only by accident of the size-1 merged axis, so worth one `(_some_nan, ())` row to make the acceptance deliberate; (2) no tuple containing numpy ints, e.g. `(np.int64(0), np.int64(1))` — the scalar `np.int64(1)` row above doesn't cover the tuple path. One row each.

mwcraig (commented):
Adversarial self-review (multi-agent, via Claude Code): every finding below was verified by running code against this branch — differential harnesses vs numpy/astropy on numpy + array-api-strict + dask, plus full runs of the changed test files (all pass). Two confirmed edge-case bugs and a handful of simplifications, all as inline comments; two findings fall outside the diff:

**`test_ccdproc.py:450-466` — `test_mad_fallback_rejects_bad_axis_tuple` now duplicates `test_bad_axis`.** Post-hoist, both grids exercise the identical `_setup` lines, which `test_nanfuncs.py::test_bad_axis` already covers 5-fold. It could shrink to a single case proving `_mad_fallback` propagates `_setup`'s error, with its `match=` strings moved into `test_bad_axis` (which currently asserts only exception types).

**Unrelated hunks:** `core.py` carries five string-concatenation cleanups (~lines 1014, 1728, 2062, 2750, 2757) with no connection to the axis work.

**Probed and clean:** `restore` round-trips exactly for arbitrary tuple orders on 4-D; all six reductions match numpy/astropy for None/tuple/negative/all-NaN cases; `_sigma_clip_mask` masks are element-identical to astropy's across axis forms; no in-place mutation, boolean indexing, or data-dependent shapes (backend-safe); duplicate/out-of-range axes raise correctly; no `NotImplementedError` catchers existed anywhere, and the TypeError change is asserted in `test_bad_axis`. Simplifications considered and rejected: inlining `_mad_fallback` (still owns the tiered `med` closure and is a test seam), routing single-int axes through the tuple branch (permute/copy on the hot combiner path), and passing tuple axes straight to `xp.sum` (forks `_setup`'s contract while median/mad still need the merge).

ccdproc/tests/test_combiner.py
```diff
@@ -1891,6 +1898,38 @@ def test_sigma_clip_mask_argument_handling():
...
+    result = _sigma_clip_mask(
+        data, sigma_lower=2, sigma_upper=2, axis=axis, maxiters=2, xp=xp
+    )
+
+    # The reference gets a tuple axis with its negative entries normalised:
+    # astropy's bottleneck dispatch transposes with the tuple as given and
+    # raises on a negative entry. The helper receives the tuple as written.
+    ref_axis = axis
+    if isinstance(axis, tuple):
+        ref_axis = tuple(ax % np_data.ndim for ax in axis)
```
mwcraig: **Test simplification: this normalization is duplicated.** `_sigma_clip_reference` already computes the same `ax % ndim` normalization for its bounds shape (line ~1761). Normalizing the tuple once inside `_sigma_clip_reference` before calling `sigma_clip` (astropy's bottleneck path can't take negative tuple entries anyway) lets this `ref_axis` block and its comment disappear, and the helper under test still receives the tuple exactly as written.

ccdproc/combiner.py
```diff
@@ -699,11 +704,11 @@ def sigma_clipping(
...
+            :func:`~astropy.stats.sigma_clip`, ``axis`` may be a single
+            integer, ``None`` or a tuple of axes on every one. ``masked``
```
mwcraig: Nit: "a tuple of axes" — `_setup` and the `_nanfuncs` docstrings say "tuple or list", and a list is accepted here too (there's a test for it). One word for consistency.

CHANGES.rst
```diff
@@ -16,6 +16,11 @@ New Features
   classified differently from astropy (NumPy data still use astropy);
   ``'median'``/``'mean'``/``'std'``/``'mad_std'`` use the namespace's
   NaN-aware reductions or ccdproc's fallbacks. [#1001]
+- ``Combiner.sigma_clipping`` outside NumPy now accepts ``axis=None`` and a
+  tuple of axes, as ``astropy.stats.sigma_clip`` does on the NumPy path, and
+  the reduction fallbacks in ``ccdproc._nanfuncs`` gained the same axis
```
mwcraig: The NotImplementedError→TypeError change for bad `axis` isn't mentioned. Verified non-breaking (`_nanfuncs` landed in unreleased #986, latest tag 2.5.1), so strictly optional — but one clause here, e.g. "an invalid ``axis`` now raises ``TypeError`` rather than ``NotImplementedError``", makes the entry airtight.

ccdproc/combiner.py
mwcraig: Agreed — and this diff has more of them than this line: `honoured` here, `Normalise`/`normalised`/`axis normalisation` in `_nanfuncs.py` (module comment and `_setup` docstring), and `normalised`/`honour` in `test_combiner.py` comments. Will switch all of them to American spelling on the branch (honored, normalize/normalized/normalization).

ccdproc/tests/test_nanfuncs.py
mwcraig: What the test pins, in full: a list of axes must mean exactly the same thing as the equivalent tuple. That is worth a test because the two dispatch paths disagree upstream — numpy's reductions reject a list outright (`np.nansum(x, axis=[0, 2])` → `TypeError: 'list' object cannot be interpreted as an integer`, verified on numpy 2.5.2), while `astropy.stats.sigma_clip` accepts one. `Combiner.sigma_clipping` forwards `axis` verbatim to whichever path the namespace selects, so if the fallbacks copied numpy's rejection, `axis=[0, 2]` would work for numpy data and blow up for every other backend. `_setup` therefore treats a list like a tuple, and this test locks that in for all five reductions.

Will expand the docstring to say that — and noted for good: every test gets an explanation of what it pins and why (saved to memory).

ccdproc/_nanfuncs.py
mwcraig: Agreed — the paragraph explaining *how* the axis forms are handled (flatten for `None`, permute-and-merge for a tuple/list) is extended discussion, not signature summary, and numpydoc puts that in Notes. Will move it into a `Notes` section of `_setup`'s docstring, keeping the one-line summary up top.

ccdproc/_nanfuncs.py
mwcraig: Yes — the `x`/`axis`/`xp` Parameters blocks of the six public functions are identical up to the verb ("to sum" / "over which to average" / ...) and the default, so they can be generated. Concrete shape (the scipy `doccer` pattern, host-side only):

```python
_COMMON_PARAMS = """x : array
    Input array. Integer and boolean inputs are promoted to the
    namespace's default real floating dtype.
axis : int, tuple of int, list of int or None, optional
    Axis or axes along which {verb}. Default is 0. ``None`` {verb_none}
    over every axis; a tuple or list {verb_plural} over all the listed
    axes at once.
xp : array namespace, optional
    Namespace to use. Defaults to ``array_api_compat.array_namespace(x)``."""

def _fill_doc(**subs):
    def deco(func):
        if func.__doc__:  # python -OO strips docstrings
            func.__doc__ = func.__doc__.format(params=_COMMON_PARAMS.format(**subs))
        return func
    return deco
```

with each function keeping its own summary line and Returns block inline — those differ meaningfully (all-NaN slices sum to zero for `nansum` but give NaN for `nanmean`/`nanstd`, `median` propagates NaN, ...), and templating them would obscure exactly the part a reader needs.

Trade-off to sign off on: the raw source (and editors that read source rather than `__doc__`) shows `{params}` at the definition; Sphinx/numpydoc and `help()` render the filled version. If that's acceptable, happy to implement; if not, my honest alternative is leaving the six short blocks as-is — they're repetitive but each is self-contained.

ccdproc/_nanfuncs.py
mwcraig: Step by step:

1. **The bool guard.** `normalize_axis_tuple` converts entries with `operator.index`, and a Python `True` passes that (`operator.index(True) == 1`), so `axis=(0, True)` would silently reduce axes `(0, 1)`. The guard turns that into a `TypeError` before the conversion runs.
2. **`normalize_axis_tuple(axis, ndim)`** does the host-side validation and normalisation in one call: each entry through `operator.index`, negatives wrapped mod `ndim` (so `-1` on a 3-D array becomes `2`), `AxisError` on out-of-bounds, `ValueError` on duplicates — including duplicates via a negative alias, e.g. `(0, -3)` on 3-D. Out comes a tuple of unique non-negative ints.
3. **`kept`** is the axes *not* being reduced, in their original order; **`order = kept + axes`** is then a permutation of `range(ndim)` that moves every reduced axis to the end. `permute_dims(x, order)` rearranges the array into that layout, and the `reshape` collapses the trailing reduced axes into one axis whose length is the product of their sizes. Every reduction in this module (sum, mean, std, median, mad) is insensitive to the order of elements *within* the reduced set, so how the merge interleaves them is irrelevant to the result.
4. The caller then runs its ordinary single-axis reduction over axis **`len(kept)`** — the merged trailing axis — which reduces over all the requested axes at once.
5. **`inverse`** is the permutation that undoes `order` (`inverse[ax]` says where original axis `ax` ended up), and **`restore`** maps a *full-shape* array in the permuted-merged layout back to the caller's layout: un-merge (`reshape` to `permuted_shape`), then un-permute (`permute_dims` with `inverse`). The reductions drop the axis and never use it; `combiner._sigma_clip_mask` computes its mask at full shape in the working layout and uses `restore` to hand it back in the shape and axis order of the input.

ccdproc/_nanfuncs.py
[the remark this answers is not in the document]
mwcraig: Confirmed — this is the same defect as https://github.com/astropy/ccdproc/pull/1006#discussion_r3891274082, and your framing of the mechanism is right: with total size 0 the `-1` is genuinely ambiguous to `reshape` (any merged length satisfies `0 × n = 0`), even though the intended value — the product of the reduced axes' sizes — is well-defined. Reproduced on this branch: `nansum(np.zeros((0, 3, 4)), axis=(1, 2))` → `ValueError: cannot reshape array of size 0 into shape (0,newaxis)` where `np.nansum` returns shape `(0,)`. Will fix by reshaping with the explicit product instead of `-1`, and add an empty-array regression row.

ccdproc/_nanfuncs.py
mwcraig: Confirmed, with one version nuance worth recording. On the oldest supported numpy (2.0.x — `pyproject.toml` floor is `numpy>=2.0`): `normalize_axis_tuple((0, np.True_), 2)` returns `(0, 1)` with only a `DeprecationWarning` — exactly the silent wrong-axes reduction described (and under the suite's warnings-as-errors config, the wrong failure mode for a test row). On current numpy (verified 2.5.2) that deprecation has expired: `operator.index(np.True_)` now raises `TypeError`, so both the tuple entry and the scalar already fail there, just with numpy's message instead of ours.

So the fix is still exactly right for the supported range: `isinstance(ax, bool | np.bool_)` in both guards gives uniform behaviour and our clearer message on every numpy from 2.0 up, and makes the docstring's "booleans are rejected" claim true again. Will apply, plus `test_bad_axis` rows for `np.True_` as a scalar and as a tuple entry (safe on all supported numpys once the guard is explicit).

ccdproc/_nanfuncs.py
mwcraig: Confirmed — reproduced verbatim on this branch: `nansum(np.zeros((0, 3, 4)), axis=(1, 2))` → `ValueError: cannot reshape array of size 0 into shape (0,newaxis)` on numpy, where `np.nansum` returns shape `(0,)`. (Copilot found the same bug independently in its comment above.) Will fix with the explicit product as suggested — `math.prod` needs a new `import math`, everything else is one line — and add the empty-array regression row. Bonus: the explicit product also makes `axis=()` on an empty array work (`prod(()) == 1`), where `-1` would hit the same ambiguity.

ccdproc/_nanfuncs.py
mwcraig: Agreed — the mechanism checks out from the code: `axis=None → tuple(range(ndim))` gives `kept=()`, so the permutation is the identity, the merge reshape degenerates to `reshape(x, (-1,))` (or the full-size product once the explicit-product fix from the other thread lands — the two compose cleanly), the returned axis is `len(kept) == 0`, and `restore` is the same reshape the dedicated branch builds, behind a no-op `permute_dims`. Your differential harness covering the 0-d case settles the rest. Will apply, dropping the dedicated `None` branch and its `restore`.

ccdproc/_nanfuncs.py
mwcraig: Verified both claims: `normalize_axis_tuple(2, 2)` raises `AxisError('axis 2 is out of bounds for array of dimension 2')` — character-identical to the hand-rolled message — and `AxisError` subclasses `ValueError`, so `test_bad_axis` passes unchanged. The asymmetry point is real too: the tuple branch already surfaces `AxisError` for out-of-bounds entries while this branch raises bare `ValueError`, and this change removes that. Will apply, keeping the bool guard and the `operator.index` try/except above for their better `TypeError` message, together with the `axis is None` fold.

ccdproc/tests/test_nanfuncs.py
mwcraig: Both verified on this branch: `nansum(x, axis=())` matches `np.nansum(x, axis=())` (elementwise, NaN→0) and `(np.int64(0), np.int64(1))` reduces identically to `(0, 1)`. Will add the two rows. Note the `axis=()` acceptance stops being an accident once the explicit-product merge fix from the reshape thread lands — the merged axis size becomes `prod(()) == 1` by construction — but a pinned row keeps it deliberate either way.

ccdproc/tests/test_combiner.py
mwcraig: Agreed — `_sigma_clip_reference` already computes the same `ax % ndim` normalisation for its bounds shape, so normalising the tuple once at the top of the helper (before its `sigma_clip` call, whose bottleneck dispatch can't take negative tuple entries) serves both uses, and this `ref_axis` block and its comment disappear while `_sigma_clip_mask` still receives the axis exactly as the parametrize row writes it. Will apply.

ccdproc/combiner.py
mwcraig: Agreed. Double-checked that "or list" is true on *both* dispatch paths before changing the doc: `astropy.stats.sigma_clip` accepts `axis=[1, 2]` (verified), so the numpy path is fine, and `_setup` handles a list off-numpy (`test_nanfuncs.py::test_list_axis_matches_tuple`). Will make it "a tuple or list of axes".

CHANGES.rst
mwcraig: Agreed — verified the non-breaking claim independently (`_nanfuncs` first appears in unreleased #986; latest release tag is 2.5.1, which has no `_nanfuncs` module), so it's a doc-completeness fix, not a compat note. Will append the clause to the #1006 entry.

mwcraig:
Replies to the two outside-diff findings from the self-review (each inline finding has an answer in its own thread — every one verified and accepted; the two edge-case bugs reproduce exactly as described):

**`test_mad_fallback_rejects_bad_axis_tuple` duplication** — agreed. Post-hoist both grids land on the same `_setup` lines, and `test_nanfuncs.py::test_bad_axis` covers them 5-fold. Will shrink it to a single case pinning that `_mad_fallback` propagates `_setup`'s error, and move the `match=` strings (`"repeated axis"`, `"out of bounds"`, `"not bool"`) into `test_bad_axis`, which currently asserts only exception types — strictly more coverage, less duplication.

**Unrelated `core.py` hunks** — confirmed: the five hunks (~1014, 1728, 2062, 2750, 2757) are implicit-string-concatenation collapses with no connection to the axis work, evidently formatter fallout. Will revert them from this branch; they can land separately as a trivial cleanup if wanted.

Planned change set from all threads, for one push: the two bug fixes (`np.bool_` guards + explicit-product merge reshape) with regression rows, the three `_setup` simplifications (`None`→tuple fold, `normalize_axis_tuple` scalar path, ~9 lines net), the two test-grid rows (`axis=()`, numpy-int tuple), the `ref_axis` dedupe, `Notes` section + expanded test docstring, American spellings throughout the diff, "tuple or list" wording, the CHANGES clause, and the `test_ccdproc.py` shrink + `core.py` revert above. The docstring-templating question (thread on `_nanfuncs.py:248`) awaits a call on the `{params}`-in-source trade-off.

ccdproc/_nanfuncs.py
mwcraig: Implement it

ccdproc/_nanfuncs.py
mwcraig: Add brief comments along these lines to the code

mwcraig:
The change set promised across the review threads has landed in two commits:

- c801431 — both edge-case bug fixes (`np.bool_` guards, explicit-product merge reshape) with regression rows, the three `_setup` simplifications, the new grid rows, the `ref_axis` dedupe, the Notes section + merge-block comments, the test-explanation docstrings, American spellings, "tuple or list" wording, the CHANGES clause, the `test_ccdproc.py` shrink, and the `core.py` cleanup revert.
- 9d18599 — the docstring templating (`_COMMON_PARAMS` + `_fill_doc`, per the thread).

One find from adding the empty-array rows: `numpy.median` itself crashes on a tuple axis with a size-0 kept dimension — the very same `-1` merge in `numpy/lib/_function_base_impl.py` — so the differential test substitutes `np.nanmedian` as the reference for that row (commented in the test). Might be worth an upstream report.

Local runs after the changes: numpy 964 (full suite), dask 701, jax 968 (full suite), array-api-strict 689 + 12 pre-existing xfails; ruff, black, and pre-commit clean.

PR #1010, opened 2026-09-07
Array-API-native k×k window filters for cosmicray_median, median_filter, background_deviation_filter and ccdmask

mwcraig:
Depends on #1009 (rebased on top of it; merge that first).

Part of #971, section 3. The masked-median fix for #984 is a planned follow-up (PR-2) that builds on the NaN handling added here.

## What changed and why

`scipy.ndimage`'s window filters are numpy-only, so the four call sites that used them silently copied a jax or dask array to the host, and failed outright for an array on a non-default device. This adds `ccdproc/_windowfilters.py`, an implementation of the four window reductions `ccdproc.core` needs, written purely in terms of the array API standard:

| helper | replaces |
|---|---|
| `window_rank` | `ndimage.percentile_filter` |
| `window_median` | `ndimage.median_filter` |
| `window_any` | `ndimage.maximum_filter` (on a boolean mask) |
| `window_reduce` | `ndimage.generic_filter` |

The approach is deliberately literal: the array is padded once — the standard has no `pad`, so `reflect` and `nearest` are built from `flip`/`concat` of edge slices — and each window offset is taken as a shifted slice, with the offsets stacked into a trailing axis, which turns a window reduction into an ordinary reduction over that axis. The rank filters reduce with a new `_nanfuncs._nanrank`, factored out of `nanmedian`'s sentinel-sort machinery; that is why they reproduce ndimage's rank exactly, including the upper-middle element of an even window.

The stack holds `prod(size)` copies of the input — 3.8 GiB for an 11x11 window over a 2048x2048 float64 image — so `_windowed` processes the output in bands of rows sized to a 256 MiB module-constant budget. Bands are cut from the *padded* array with the window overhang included, so each output pixel sees exactly the window it would have seen unbanded; `band_rows` stays on the private helpers (a test pins banded == unbanded) and is not exposed on any public function.

## Dispatch policy: the numpy path is unchanged

Four wrappers in `core.py` — `_dispatch_median_filter`, `_dispatch_percentile_filter`, `_dispatch_maximum_filter`, `_dispatch_generic_filter` — send numpy input to `scipy.ndimage` exactly as before and everything else to `_windowfilters`. Concentrating the choice in one place per filter also leaves the seam #984 needs, where numpy input *carrying a mask* will start taking the native path too.

[...]

PR #1009, opened 2026-09-07
Array-API-native block_reduce, block_average and block_replicate off numpy

mwcraig:
`ccdproc.core.block_reduce`, `block_average` and `block_replicate` are thin
wrappers around `astropy.nddata`, which starts by calling `numpy.asanyarray`
on its input. On a non-numpy array library that silently hands back a numpy
array (dask, jax) or fails outright when the data live on a device numpy
cannot reach (array-api-strict, and by extension cupy on a GPU). These were
three of the remaining array-API escapes in the escape baseline.

This adds `ccdproc/_blocks.py`, which does the same work using only array API
operations (`reshape`, `permute_dims`, `repeat` and slicing), and dispatches
to it for every namespace that is not numpy.

This is deliberately a stopgap: it exists so ccdproc's array-API work does not
have to wait for an astropy release. `_blocks.block_reduce` and
`_blocks.block_replicate`, minus the ccdproc dispatch, are the intended
starting point for astropy/astropy#15073; once astropy's own `blocks.py` is
array-API aware and that version is ccdproc's floor, this module and the
dispatch both go away. (`block_average` has no astropy counterpart -- it is
ccdproc's own thin wrapper -- so only the other two are candidates for the
lift.)

## Dispatch policy

The numpy path is unchanged: `if array_api_compat.is_numpy_namespace(xp)` the
functions call `astropy.nddata` exactly as before, with the same arguments.
Everything else goes to `_blocks`. The CCDData rebuild and the
"following attributes were set ... will be ignored" warning behaviour are
untouched -- the native helpers are decorated with
`astropy.nddata.support_nddata`, as astropy's own are, so a `CCDData`
argument is unpacked and warned about identically on every backend.

`block_size` validation is done in pure Python and reproduces astropy's three
checks in astropy's order with astropy's messages verbatim; the tests assert
against astropy's live output rather than hard-coded strings so the two cannot
drift apart.

## Behaviour differences

Results are identical to astropy's except for two dtype promotions, both of
integer/boolean input to the namespace's default real floating dtype:

- `block_replicate(..., conserve_sum=True)` promotes before dividing, because
  array-api-strict rejects integer true division rather than promoting.
- `block_average` promotes before averaging, because array-api-strict rejects
  a non-floating `mean`. numpy promotes on its own here, and jax and dask
  follow it, so without this an integer image averaged fine on three backends
  and raised on the fourth.

numpy returns `float64` in both cases anyway, so these differ only for a
[...]

PR #1011, opened 2026-09-07
Make the host round trip explicit in wcs_project, subtract_overscan and cosmicray_lacosmic

### LINKED ISSUE #930: Array API: wcs_project mixes numpy output from reproject with xp operations
### Problem

In `wcs_project`, `reproject_interp` returns numpy arrays regardless of the input library, but the subsequent mask handling uses the *input's* namespace:

https://github.com/astropy/ccdproc/blob/9d25eeefda8a38fc442c1a18e79ca2fa8ca40559/ccdproc/core.py#L1163-L1180

With `xp = cupy`, `xp.isnan(projected_image_raw)` receives a numpy array — CuPy functions reject numpy input (`TypeError: 'a' must be a cupy.ndarray`). Even where it happens to work, the returned `CCDData` holds numpy data while the caller supplied another library's array, silently switching namespaces mid-pipeline. Likely contributes to the WCS-related failures in #910.

### Suggested direction

`reproject` is CPU-only, so this falls under the general policy decision for CPU-only operations (#935): either convert explicitly to host, reproject, and convert the result (data *and* mask) back to the input namespace — or document `wcs_project` as numpy-only and raise an informative error.

---
Found during a review of the array API implementation from #885; follow-up to #909 / #910.

### LINKED ISSUE #933: Array API: subtract_overscan model fitting goes through numpy-only astropy.modeling

### LINKED ISSUE #935: Array API: decide a policy for inherently CPU-only operations

mwcraig:
Part of #971. Closes #930, closes #933, closes #935. Independent of #1009 and #1010 except for one expected conflict in `ccdproc/tests/test_cosmicray.py` (see below).

## What changed and why

Three ccdproc operations depend on a library that is NumPy-only: `wcs_project` (reproject), `subtract_overscan` with a `model` (`astropy.modeling` fitting) and `cosmicray_lacosmic` (astroscrappy). Until now they each let their input leak to NumPy implicitly and handed a NumPy array back to a caller who had passed jax, dask or a device array. This PR makes that host round trip explicit and symmetric:

- A new `ccdproc.HostCopyWarning` (subclass of `AstropyUserWarning`, exported from `ccdproc`) is emitted once per call for non-NumPy input, attributed to the caller's own line (the `stacklevel` accounts for the `log_to_metadata` / `deprecated_renamed_argument` decorators on all three functions). NumPy input is neither copied nor warned about. The warning is filtered out in the test configuration.
- Two helpers next to `_to_numpy` in `core.py`: `_from_numpy(array, like=..., xp=...)` converts a result back to the namespace and device of the input, and `_warn_host_copy(name, xp, stacklevel=...)` issues the warning.
- `wcs_project`: data and mask are copied to the host explicitly, the reprojected data and mask are converted back, the mask is combined in the namespace via `_mask`, and the pixel-area ratio is a Python `float` (astropy returns a NumPy scalar, which array-api-strict refuses as an operand). `target_shape` is now coerced to a tuple of Python ints so any two-integer sequence or array is accepted, including an array from `ccd`'s namespace; reproject needs `len()` of it and array-api-strict arrays have none.
- `subtract_overscan` with a `model`: the fit and its evaluation run in NumPy on the host and the result is converted back.
- `cosmicray_lacosmic`, both the `CCDData` and the bare-array branch: data, mask and any array-valued `inbkg`/`invar` go to the host; `cleanarr` and `crmask` come back in the caller's namespace and device. The bare-array branch now returns caller-namespace arrays too.
- `_astroscrappy_gain_apply_helper` multiplies by `float(gain.value)` instead of the NumPy `float64` `gain.value`. array-api-strict rejects a NumPy scalar as an operand (`TypeError: Expected Array or Python scalar`); `gain_correct` already did this coercion at its own call site. This is what the two "gain and mask" strict markers were really recording.

## Tests and markers

- New `ccdproc/tests/test_hostcopy.py` (16 tests per backend) pins the contract: the warning is emitted exactly once, attributed to the caller, only for non-NumPy input; every returned array is in the input's namespace and on the input's device; NumPy input is silent and uncopied; and a scalar `Quantity` gain keeps namespace and device on array-api-strict's non-default device (the #936 audit, see below).
- The array-api-strict `backend_xfail` markers these three functions had on `test_ccdproc.py` and `test_cosmicray.py` are removed; on strict, 31 previously-xfailed test instances now pass. The `subtract_overscan` line is deleted from `array_escape_baseline.txt`; the escape ratchet (`CCDPROC_LOG_ARRAY_ESCAPES=1 CCDPROC_ENFORCE_ESCAPE_BASELINE=1`, full dask run) reports no escapes outside the baseline and no stale entries. All three sites' copies are attributed to `core.py _to_numpy`, so no new caller-attributed lines were needed.
- Tests that ran on strict for the first time needed NumPy-isms removed: `.sum()` calls, `assert_allclose` on device arrays (now via `_to_numpy`, the pattern `test_combiner.py` uses), Python-int slice bounds, and `add_cosmicrays` copying through a host round trip because a host copy of a jax array is read-only. The `no_cosmics` stand-in for `detect_cosmics` now returns NumPy arrays like the real one, so it exercises the conversion back.
- `test_wcs_project_onto_scale_wcs`'s input mask is a bool array rather than a float one. reproject casts to float internally either way; for the nearest-neighbour order the test uses the outputs are bit-identical.

## Results

Full `pytest ccdproc` on each backend (`JAX_ENABLE_X64=1` for jax), compared with `main` at cfdbdd4:

| backend | main | this branch |
|---|---|---|
| numpy | 964 passed, 35 skipped | 977 passed, 39 skipped |
| jax | 963 passed, 36 skipped | 977 passed, 39 skipped |
| dask | 957 passed, 42 skipped | 971 passed, 45 skipped |
| array-api-strict | 917 passed, 36 skipped, 46 xfailed | 962 passed, 39 skipped, 15 xfailed |

No test regressed on any backend. Every backend gains exactly 17 tests: the 16 in `test_hostcopy.py` (their pass/skip split differs per backend because the warning and no-warning tests are complementary) and the new `target_shape` test. On strict the 31 previously-xfailed instances all pass. Skip counts are higher than in CI because this machine lacks bottleneck and is not Linux.

## #936

Audited as part of this: every `Quantity` ccdproc constructs on the gain/readnoise path wraps a Python scalar, so none of them can pull data out of its namespace or off its device. What did bite was the `gain.value` NumPy scalar above, now fixed and pinned by `test_scalar_quantity_gain_keeps_namespace_and_device`. An *array-valued* gain or readnoise would need upstream astropy work (`data << unit` in the arithmetic mixin is not array-API aware) and is the remaining `test_unit_mismatch_behaves_as_expected` xfail, tracked separately. A closing comment for #936 will follow separately rather than via this PR.

## Expected conflict with #1010

`add_cosmicrays` in `ccdproc/tests/test_cosmicray.py` is shared with the `cosmicray_median` tests #1010 touches, and both PRs remove markers in that file. Whichever lands second rebases; the conflicts are adjacent-line.

## Not verified

- CuPy: not installed, so untested, as usual for this repo.
- The docs build was not run locally (`sphinx-astropy` is not installed here); CI will check the new `docs/array_api.rst` section and the three docstring Notes.

https://claude.ai/code/session_01F6f9L1GyMHrKnvLrrgWbWN


mwcraig:
Rebased onto `main` now that #1009 has merged; the branch was force-pushed with `--force-with-lease`.

Only one file conflicted, `CHANGES.rst`: #1009 and this PR both added entries at the end of the *New Features* list, so both were kept, #1009's two entries first and this PR's host-round-trip entry after them. Everything else merged without conflict — `ccdproc/core.py` and `docs/array_api.rst` merged cleanly because #1009 touched the block functions and this PR touches `wcs_project`, `subtract_overscan` and `cosmicray_lacosmic`. No adjacent-line conflict with #1010, which has not merged yet.

Nothing here needed to adopt the helpers #1009 introduced: this branch never normalises an array *module* to a namespace (the three functions take `xp` or read it off the data with `array_api_compat.array_namespace`), so there is no hand-written copy of `_namespace_from_module` to replace, and no `_block_dispatch`-style `except TypeError` guard either. The rebase is therefore the 13 original commits unchanged, plus one new commit for the baseline.

The escape baseline was regenerated the documented way (a full dask run with `CCDPROC_WRITE_ESCAPE_BASELINE=1`) rather than hand-merged. Its *content* is exactly what the rebase produced: #1009's three `block_*` entries are gone, this PR's `subtract_overscan` entry is gone, and the six entries that stay — `combine`, `_to_numpy`, `_cosmicray_median_array`, `cosmicray_median`, `background_deviation_filter`, `create_deviation` — are all still there, including `_to_numpy`, which this PR deliberately keeps as the one named place where a host copy happens. The regeneration did change the file's layout: `subtract_overscan` was the last `numpy.asanyarray` entry, so the coercion column is now one word narrower. That is the whole of the extra commit.

Local tox, all green:

| env | result |
| --- | --- |
| `codestyle` | all checks passed |
| `strict` | 1096 passed, 42 skipped, 12 xfailed |
| `py312-alldeps-jax` | 1103 passed, 47 skipped |
| `py312-alldeps-dask-enforce` | 1105 passed, 45 skipped, ratchet: "no library escapes outside the baseline" |

